PR #6, opened 2020-05-04
Feature/planning data

aaronchongth:
Added some trajectories data for a single planning state instance, it is in similar form of [trajectories.json](https://github.com/osrf/romi-dashboard/blob/master/src/mock/data/trajectories.json), so ideally we can use the same utilities used in parsing and visualizing regular trajectories for these planning state trajectories.

Requires https://github.com/osrf/romi-dashboard/pull/5 to be merged first, as it was a reversion of a mistake.

Works with https://github.com/osrf/rmf_schedule_visualizer/pull/45


koonpeng:
Closing this since it doesn't appears to be relevant anymore.

Issue #7, opened 2020-05-04
Draw doors on the map


baviera08:
We need to draw doors on the map. 

### Checks
* Door is in the currently selected floor.
* Mode of the door. (0:Open, 1:Moving, 2:Open).

I'm currently looking into this.

PR #19, opened 2020-05-07
Feature/ Draw doors on canvas

baviera08:
Related to #7  
Requires #20 to be merged first. 

### What's new
* Splited doors-overlay into doors-overlay and doors.
* Added two doors components `SingleHingeDoor`, `DoubleHingeDoor`.
* Doors changing styles on state change (Open, close, process).
* Added a new door "exit".
* Set new doors coordinates.

dash

### Discussion
*  Maybe it'll be a good idea to create a "door" folder and add all doors components there in different files. We need to cover these cases https://github.com/osrf/traffic_editor/blob/master/building_map_msgs/msg/Door.msg
* For now, the doors will not have movement (we don't have the data to do that yet), as we discussed with @Yadunund the states of doors will be differentiated with colors and shapes. I'll create an issue to draw the door movement as future work.


baviera08:
So, I'll close this, I used this PR as a base to make #21 
And I already made some changes on the doors on #21

Issue #8, opened 2020-05-04
Draw lifts on the map


baviera08:
We need to draw lifts on the map.

### Checks
* Lift is in the currently selected floor.
* Mode of the lift:
     - DOOR_CLOSED = 0;
     - DOOR_MOVING = 1;
     - DOOR_OPEN = 2;
     - MOTION_STOPPED = 0;
     - MOTION_UP = 1;
     - MOTION_DOWN = 2;
     - MOTION_UNKNOWN = 3;
     - MODE_UNKNOWN = 0;
     - MODE_HUMAN = 1;
     - MODE_AGV = 2;
     - MODE_FIRE = 3;
     - MODE_OFFLINE = 4;
     - MODE_EMERGENCY = 5;

PR #21, opened 2020-05-07
Feature/draw lift on canvas

### LINKED ISSUE #7: Draw doors on the map

baviera08:
Requires #20 to be merged first.
Fixes #7 and #8 

### Doors
* Splited doors-overlay into doors-overlay and doors.
* Added two doors components SingleHingeDoor, DoubleHingeDoor.
* Doors changing styles on state change (Open, close, process).
* Added a new door "exit".
* Set new doors coordinates.
* Added tests.
* Created a spinner component to show it when the door has PROCESS mode.

### Some  examples
open door
open door

door in process mode
opening door

Doors look

image

### Lift
* Created Lift and LiftOverlay components
* Draw lifts on the office map.
* Lift changing styles on mode (FIRE, EMERGENCY, OFFLINE, etc) and motion (UP, DOWN, STOP, UNKNOW) change.
* Added new lift on the building map.
* Set new lift coordinates. Transformed the middle coordinate got it from the lift msg, to the top left to draw it as an SVG on the map.
* Created UpArrow and DownArrow components and calculated their position based on the center of the lift. Added tests for this. 
* Added tests. 

### Some  examples
Lift not on the current floor and door closed
image

Lift on fire mode and the door closed with motion Up.

image

Lift on the current floor, doors open, motion Stopped, mode AGV
image

## Discussion
* Maybe it'll be a good idea to create a "door" folder and add all doors components there in different files. We need to cover these cases https://github.com/osrf/traffic_editor/blob/master/building_map_msgs/msg/Door.msg
* For now, the doors will not have movement (we don't have the data to do that yet), as we discussed with @Yadunund the states of doors will be differentiated with colors and shapes. I'll create an issue to draw the door movement as future work.


src/components/schedule-visualizer/lift-overlay.tsx
```diff
@@ -0,0 +1,46 @@
...
+  const viewBox = viewBoxFromLeafletBounds(props.bounds);
+  const useStateValue = () => useContext(LiftStateContext);
+  const liftsState = useStateValue();
+  return (
+    <>
+      <SVGOverlay {...otherProps}>
+        <svg viewBox={viewBox}>
+          {lifts.map(lift => (
+            <Lift
+              key={`lift-${lift.name}`}
```
AlejoAsd, line 26: Are lift names unique? If that is not the case, this could generate some invalid values for `key`.

src/components/schedule-visualizer/lift.tsx
```diff
@@ -0,0 +1,184 @@
...
+  const isModeHuman = currentMode === LiftModeStates.HUMAN;
+  if (isDangerous) return classes.danger;
+  if (isOffLine) return classes.offLine;
+  if (!isInCurrentFloor) return classes.liftOnAnotherFloor;
+  if (isInCurrentFloor && isModeAGV) return classes.liftOnCurrentFloor;
+  if (isInCurrentFloor && isModeHuman) return classes.humanMode;
+};
+
+/**
+ * Gets the text to insert to the lift, the text depend on the current mode, motion state and the  * current and destination floor of the lift.
```
AlejoAsd, line 54: Two comment lines were joined here. A linebreak is needed.

koonpeng:
I haven't looked at the code, but there are a couple of small details I noticed when going for a quick spin run.

1. FIRE, EMERGENCY and OFFLINE mode has the lift opaque, while other states have transparency.
2. The lift layer is on top of the robot layer, so it covers the robot when it's in modes that are opaque.
4. FIRE, EMERGENCY and OFFLINE mode size is inconsistent with the other modes, its bigger and has more rounded corners.
5. It looks like lift state is not subscribed until the lift panel is opened. Since we are now drawing the lifts we can subscribe to it as soon as the transport is ready.
6. Adding a "unknown" label when we don't know the state of the lift would be nice.

src/components/schedule-visualizer/doors-overlay.tsx
```diff
@@ -1,62 +1,47 @@
...
-          );
-        })}
-      </svg>
-    </SVGOverlay>
+    <>
+      <SVGOverlay {...otherProps}>
+        <svg viewBox={viewBox}>
+          {doors.map(door => (
+            <Door
+              key={`building-door-${door.name}`}
```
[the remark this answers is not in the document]
baviera08: Yeah, I agree with you, It looks like there are unique, let's confirm this with @koonpeng ?

src/components/schedule-visualizer/lift-overlay.tsx
baviera08: Let's confirm this with @koonpeng too

src/components/schedule-visualizer/arrow.test.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const wrapper = mount(
+    <svg>
+      <DownArrow x={2} y={2} />
+    </svg>,
+  );
+  const { x1, y1, x2, y2, x3, y3 } = wrapper
+    .find(Arrow)
+    .at(0)
+    .props();
+  expect([2, 2, 1, 0, 3, 0]).toEqual([x1, y1, x2, y2, x3, y3]);
```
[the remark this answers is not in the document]
baviera08: Good catch! makes total sense.

src/components/schedule-visualizer/arrow.tsx
```diff
@@ -0,0 +1,82 @@
...
+  );
+};
+
+export const Arrow = (props: BaseArrowProps) => {
+  const { x1, y1, x2, y2, x3, y3, strokeWidth, padding = [0, 0] } = props;
+  const classes = useStyles();
+  return (
+    <polygon
+      points={`${x1 + padding[0]},${y1 + padding[1]} ${x2 + padding[0]},${y2 + padding[1]} ${x3 +
+        padding[0]},${y3 + padding[1]} `}
```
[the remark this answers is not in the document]
baviera08: Is valid! Thanks!

src/components/schedule-visualizer/door.tsx
```diff
@@ -0,0 +1,139 @@
...
+// Door state
+export enum DoorStates {
+  CLOSE,
+  PROCESS,
+  OPEN,
+}
+
+// Door Type
+const DOOR_TYPE_SINGLE_SLIDING = 1;
+const DOOR_TYPE_DOUBLE_SLIDING = 2;
```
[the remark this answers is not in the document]
baviera08: Yeah, I missed that one. Thank you!

src/components/schedule-visualizer/door.tsx
```diff
@@ -0,0 +1,139 @@
...
+export interface DoorProps {
+  door: RomiCore.Door;
+  onClick?(e: React.MouseEvent<SVGGElement>, door: RomiCore.Door): void;
+  currentMode?: number;
```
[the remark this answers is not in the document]
baviera08: No, in this case `currentMode` it's ok. `DoorState` is a bad name.
Now the enum it's called 
``` typescript
export enum DoorMode {
  CLOSE,
  PROCESS,
  OPEN,
}
```
Because the `State` contains the `currentMode` of the door

src/components/schedule-visualizer/door.tsx
```diff
@@ -0,0 +1,139 @@
...
+  doorState?: RomiCore.DoorState;
```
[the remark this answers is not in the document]
baviera08: Done!

src/components/schedule-visualizer/door.tsx
```diff
@@ -0,0 +1,139 @@
...
+}
+
+const getDoorStyle = (classes: any, currentMode: number | undefined) => {
+  return currentMode === DoorStates.OPEN
+    ? classes.doorOpen
+    : currentMode === DoorStates.PROCESS
+    ? classes.doorProcess
+    : classes.doorClose;
```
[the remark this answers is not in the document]
baviera08: Done! Thanks!

src/components/schedule-visualizer/door.tsx
```diff
@@ -0,0 +1,139 @@
...
+};
+
+export const SingleHingeDoor = React.forwardRef(function(
```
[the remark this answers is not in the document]
baviera08: Done! Thanks!

src/components/schedule-visualizer/doors-overlay.tsx
```diff
@@ -1,62 +1,47 @@
...
 }
 
 export default function DoorsOverlay(props: DoorsOverlayProps): React.ReactElement {
-  const classes = useStyles();
-  const bounds =
-    props.bounds instanceof L.LatLngBounds ? props.bounds : new L.LatLngBounds(props.bounds);
+  const { doors, onDoorClick, ...otherProps } = props;
+  const viewBox = viewBoxFromLeafletBounds(props.bounds);
+  const useStateValue = () => useContext(DoorStateContext);
+  const doorsState = useStateValue();
```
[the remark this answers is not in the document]
baviera08: Yeah! It's equal, it was a leftover, thanks!

baviera08:
> I haven't looked at the code, but there are a couple of small details I noticed when going for a quick spin run.
> 
> 1. FIRE, EMERGENCY and OFFLINE mode has the lift opaque, while other states have transparency.
> 2. The lift layer is on top of the robot layer, so it covers the robot when it's in modes that are opaque.
> 3. FIRE, EMERGENCY and OFFLINE mode size is inconsistent with the other modes, its bigger and has more rounded corners.
> 4. It looks like lift state is not subscribed until the lift panel is opened. Since we are now drawing the lifts we can subscribe to it as soon as the transport is ready.
> 5. Adding a "unknown" label when we don't know the state of the lift would be nice.
> 6. When the door is opened, it drawn as a dashed line, clicking on the door only works when clicking on the dashed part, some tricks with transparency to make the whole door clickable would be nice.

Thanks for the your comments, I addressed 1,2,3,5,6 on this commit 7626e05f42029e98b3f97cfaf7d4039ec0eab6cb . I'm currently working on 4.

src/components/schedule-visualizer/doors-overlay.tsx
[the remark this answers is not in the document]
koonpeng: yes, door names are unique, although there isn't anything enforcing this rule, we can assume they are unique by design.

src/components/schedule-visualizer/lift-overlay.tsx
koonpeng: yes they are unique

src/components/schedule-visualizer/doors-overlay.tsx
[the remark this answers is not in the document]
baviera08: Great! thank you!

src/components/schedule-visualizer/lift-overlay.tsx
baviera08: Great! Thank you!

src/components/app.tsx
```diff
@@ -101,23 +103,18 @@ export default function App(props: AppProps): JSX.Element {
...
   );
 
   const liftStateManager = React.useMemo(() => new LiftStateManager(), []);
-  const [liftStates, setLiftStates] = React.useState<Readonly<Record<string, RomiCore.LiftState>>>(
-    {},
-  );
-  const [lifts, setLifts] = React.useState<readonly RomiCore.Lift[]>([]);
-  // FIXME: not used for now as there is not enough information to render lifts.
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const [liftStates, setLiftStates] = React.useState(liftStateManager.liftStates());
```
koonpeng, line 114:
```suggestion
  const [liftStates, setLiftStates] = React.useState(() => liftStateManager.liftStates());
```
`liftStateManager.liftStates()` creates a shallow copy of all the current states, which may be somewhat expensive so its better to use lazy initial state.

src/components/app.tsx
```diff
@@ -101,23 +103,18 @@ export default function App(props: AppProps): JSX.Element {
   const trajManager = React.useRef<RobotTrajectoryManager | undefined>(undefined);
 
   const doorStateManager = React.useMemo(() => new DoorStateManager(), []);
-  const [doorStates, setDoorStates] = React.useState<Readonly<Record<string, RomiCore.DoorState>>>(
-    {},
-  );
+  const [doorStates, setDoorStates] = React.useState(doorStateManager.doorStates());
```
koonpeng, line 106:
```suggestion
  const [doorStates, setDoorStates] = React.useState(() => doorStateManager.doorStates());
```
`doorStateManager.doorStates()` creates a shallow copy of all the current states, which may be somewhat expensive so its better to use lazy initial state.

src/components/schedule-visualizer/doors-overlay.tsx
```diff
@@ -1,62 +1,46 @@
...
+  onDoorClick?(door: RomiCore.Door): void;
 }
 
 export default function DoorsOverlay(props: DoorsOverlayProps): React.ReactElement {
-  const classes = useStyles();
-  const bounds =
-    props.bounds instanceof L.LatLngBounds ? props.bounds : new L.LatLngBounds(props.bounds);
+  const { doors, onDoorClick, ...otherProps } = props;
+  const viewBox = viewBoxFromLeafletBounds(props.bounds);
+  const doorsState = useContext(DoorStateContext);
```
koonpeng, line 20: I usually avoid using contexts as I feel it creates a dependency and makes the component less isolated. That being said, I'm still quite new to React so it would help if you could explain the advantage of using context here over passing them as props.

src/components/schedule-visualizer/door/door.tsx
```diff
@@ -0,0 +1,117 @@
+import * as RomiCore from '@osrf/romi-js-core-interfaces';
+import React from 'react';
+import { Spinner } from '../spinner';
+import SingleHingeDoor from './door-single-hinge';
+import DoubleHingeDoor from './door-double-hinge';
+import SingleSlideDoor from './door-single-slide';
+import DoubleSlideDoor from './door-double-slide';
+import DefaultDoor from './door-default';
+
+export enum DoorMode {
```
koonpeng, line 10: Is it possible to reuse the constants in `romi-js-core-interfaces`?

src/components/schedule-visualizer/door/door.tsx
```diff
@@ -0,0 +1,117 @@
...
+  CLOSE,
+  PROCESS,
+  OPEN,
+}
+
+export enum DoorType {
```
koonpeng, line 16: Is it possible to reuse the constants in `romi-js-core-interfaces`?

src/components/schedule-visualizer/door/door.tsx
```diff
@@ -0,0 +1,117 @@
...
+          v2={v2}
+          door={door}
+          onClick={handleClick}
+          currentMode={currentMode}
+        />
+      )}
+      {doorType === DoorType.UNDEFINED && <DefaultDoor v1={v1} v2={v2} onClick={handleClick} />}
+      {currentMode === DoorMode.PROCESS && (
+        <g>
+          <Spinner cx={door.v1_x} cy={-door.v1_y} r={0.4} strokeWidth={0.1}></Spinner>
```
koonpeng, line 110: imo the spinner looks very out of place, until there is a proper animation we might be better off just changing the color.

src/components/schedule-visualizer/lift.tsx
```diff
@@ -0,0 +1,211 @@
+import { makeStyles } from '@material-ui/core';
+import * as RomiCore from '@osrf/romi-js-core-interfaces';
+import React from 'react';
+import { RomiCoreLift } from './lift-overlay';
+import Door from './door/door';
+import { UpArrow, DownArrow } from './arrow';
+
+export enum LiftMotionStates {
```
koonpeng, line 8: Can we reuse the constants from `romi-js-core-interfaces`?

src/components/schedule-visualizer/lift.tsx
```diff
@@ -0,0 +1,211 @@
...
+  STOPPED,
+  UP,
+  DOWN,
+  UNKNOWN,
+}
+
+export enum LiftModeStates {
```
koonpeng, line 15: Can we reuse the constants from `romi-js-core-interfaces`?

src/components/app.tsx
baviera08: Yeah! you're totally right! Thank you!

src/components/app.tsx
baviera08: Yeah! you're totally right! Thank you!

src/components/schedule-visualizer/door/door.tsx
baviera08: It makes sense I'll open an issue to discuss doors movements animation. Thanks for the feedback.

src/components/schedule-visualizer/door/door.tsx
baviera08: Done! Thanks!

src/components/schedule-visualizer/lift.tsx
baviera08: Done! Thanks!

src/components/schedule-visualizer/lift.tsx
baviera08: Done! Thanks!

src/components/schedule-visualizer/door/door.tsx
baviera08: Done! Thanks!

src/components/schedule-visualizer/doors-overlay.tsx
baviera08: Using context it's a way to avoid doing something called prop-drilling. 

_Prop Drilling means to pass props multiple levels of components deep in order to reach a particular component that needs those props. And since a parent can’t pass a prop directly to one of its grandchildren it needs to go through other components on its way down.
The problem is that all those other components won’t actually use the prop, they will just pass it down to lower level. This leads to the unnecessary exposure of information and tighter coupling in the code base. Changing the data’s structure or the name of the prop would require modifications in multiple places.
It’s a common occurrence to rename a prop in a component due to name collisions or for the sake of clarity. Doing that a couple of times in the prop drilling chain will only leave us confused and frustrated._

src: https://hackernoon.com/lessons-learned-common-react-code-smells-and-how-to-avoid-them-f253eb9696a4 or https://medium.com/@jtonti/context-vs-prop-drilling-in-react-aa88fff7cd44

E.g:
_With propDrilling: App -> Schedule Visualizer -> DoorOverlay -> Door_
_With context: App -> Door_

Let me know what you think about it.

Hope this helps.

koonpeng: Thanks for the great explanation! I have pretty much the same gripes about contexts regarding how it affects the maintainability of the code.

In the example above, the downside I see from context is that even though CH1 and CH2 does not use the context, you cannot render them without a provider (CP) somewhere in the ancestors. This can make the code less readable as there is no where in the definition of CH1 and CH2 that says that they require a context provider, it creates a somewhat hidden dependency that might not be obvious at first glance.

Having a guideline for the use of context sounds like a good idea, but I don't think this project is large enough to need it yet and I don't think I'm experienced enough to write it. But please do share if you know of a good react guideline that we can use.

koonpeng:
I just updated `romi-js-core-interfaces` with the latest message definitions, I'm ready to merge this in once the TODOs with lifts messages are taken care of.

baviera08:
Hey @koonpeng, I removed the TODOs of the lift, but there is a new issue that we need to solve before you merge. https://github.com/osrf/romi-js-core-interfaces/issues/5 .

src/components/schedule-visualizer/doors-overlay.tsx
baviera08: I'm glad that it helps. I'll look for one. :)

koonpeng:
@baviera08 i published a new version with the door msg updated. I'm just gonna merge this now as I don't find any problems with `0.0.2-alpha.5`, you can update later PRs to `0.0.2-alpha.6`.

PR #25, opened 2020-05-19
Feature/added Storybook to the project

baviera08:
Related to #7
Requires #21  to be merged first.

### What's new
* Added storybook
* Updated Typescript version
* Added storybook script
* Splitted lift component into container component and presentational component.
* Added the lift component in the Storybook
 
storybook

We added something call Storybook to the project. The storybook is a user interface, software development environment, and playground for UI components. The main purpose of it is to allow you to create components independently in an isolated development environment outside the main application. We can also use this to show our components in their different states.

Storybook runs on its own web server and provides a catalog view for the components for our application. We can run the storybook project with `yarn storybook` command.

Using the Storybook will also force us to split the component into two: (1) presentational components (the component with no logic behind); and (2) container components (where the component logic is. E.g: coordinates calculations, etc.). Which is a widely used design.


koonpeng:
this looks like based on an older version of #21, please merge master and fix the conflicts.

baviera08:
Yeah, you were right, it was not updated. Done, merged master, and solved conflicts. Thanks @koonpeng

src/components/schedule-visualizer/lift/lift.tsx
```diff
@@ -0,0 +1,114 @@
+import { makeStyles } from '@material-ui/core';
```
koonpeng, line 1: There is another `lift.tsx` at `src/components/schedule-visualizer/lift.tsx`, maybe something went wrong during the merge.

src/components/schedule-visualizer/spinner.tsx
```diff
@@ -0,0 +1,56 @@
+import { makeStyles } from '@material-ui/core';
```
koonpeng, line 1: I don't think this is used anymore, if we ever need it next time it is probably better to use a third party implementation like the one from material ui.

src/stories/1-Button.stories.js
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
```
koonpeng, line 1: Let's remove this since it's not useful for the app.

src/stories/elements/lift.stories.js
```diff
@@ -0,0 +1,115 @@
...
+            x: 50,
+            y: 50,
+            text: 'L1->L2',
+            fontSize: '0.8rem',
+          }}
+          liftModeText={{
+            x: 50,
+            y: 25,
+            text: 'EMERGENCY!',
+            fontSize: '0.8rem',
```
koonpeng, line 42: These are using fixed values for `liftMotionText` and `liftModeText` props, they may diverge from the actually style used in the app. Is it possible to refactor this to render the lift with a `Lift` and `LiftState` object, letting the `LiftContainer` component handle the props?

src/components/schedule-visualizer/spinner.tsx
baviera08: Yeah, I'll remove this. 

FYI It is a little tricky to use the spinner inside the layer. Material-ui indeed uses a circular vector to draw the spinner but it is contained in a <div> element, which complicates using them inside the layer.

src/components/schedule-visualizer/lift/lift.tsx
baviera08: Yeah, it looks like it. I guess I'll stick to rebase and cherry pick. =/

src/stories/1-Button.stories.js
baviera08: Done!

src/components/schedule-visualizer/spinner.tsx
baviera08: Done!

src/stories/elements/lift.stories.js
baviera08: Yeah, I understand what are you saying. I guess I can refactor the lift component and the storybook so we can get close to the current style without much effort (e.g: using the mock data).

The liftContainer has calculations used to scale the components to the canvas that we have on Romi-dashboard, so we can't use the liftContainer just to handle the styles. But I guess we can separate the style and text handler from the Component and call it here.

We will have as many states as lifts.

baviera08: What do you think of this?:

* We can use the Storybook to visualize (in this case) the lift components and its states and have a playground to modify the visual component.

* And the calculations and states of the container components should be tested with jest/enzyme.

koonpeng: I think an advantage of storybook is If we can get the storybook rendered components as close as possible to the actual render code used in the actual app, it can help visualize how different calculations affect the look.

Of course its still would be good to have tests in jest. Storybook would help with the visual design aspects, and jest would help in the technical aspects.

baviera08:
This has become too old and too big, I'll split it up and close this one.

PR #27, opened 2020-05-25
Feature/addLoopRequestForm

baviera08:
## What's new
* Added @material-ui/lab, sweetalter2, and a `uuid` generator libraries
* Split robot-item into robot-item-information and robot-item-form
* Added a `requestLoop` publisher and form.
* Fixed location of the mock lift.
* Removed some outdated TODOs.
* Added a file where are hardcoded the places where a fleet should go. https://github.com/osrf/romi-dashboard/blob/11e99903359222dd0224bba06fef577ade005ff2/src/mock/data/places.ts
* Added alert components

loop working with form

## Discussion

* Do you want me to remove or find another way to display the success alert?. I like [this](https://codepen.io/verknxt/pen/BaNyGNy?__cf_chl_jschl_tk__=9c00533c5619ddedf4d5b7ecdd6617f1a96827e0-1590699824-0-AUcCtrcm0BxZnu0Zck5Tka0tZtNklVDk_pgE7dZE4ocDC43fY_WCYXN3_-VHVWJSBhHLH7AglZMNdHgTOXE84DG7wd5ZKRkrvC4kLI-vdH_1gQuSCkckF4V6Rif8Bdm6lcXeyW6EoF38kneg-Nr9OcueQ-Z9OOZ4TxAZut0Yuc0NiWOn7jBfuAT_DpXD40LLdMkJDoW_Uf952Eb_nMwY7Z81eFF-uTh0DwJigF5PZrgQe60rO5ioVJSzgeRfGNh1hFRCyDc8Mtnqgn1ynQ0tquxo71HbjtJ3Yza7hzg7nfWsxcK22pVAjqV2W7xGMw5SEL57QzuusgPQ78bzkzZXLdBUwtIrcp8ccQVlcRzj90el) one too This is nice too


baviera08:
Hi guys, @koonpeng and @Yadunund, Do you have some feedback or comments about UI, UX? =)

Yadunund:
Hi @baviera08 , thanks for prototyping the UX so quickly.
It looks good! I have these minor comments.
- Could we have static labels for each field, perhaps above text input box? The "helper" in the text box can be a more detailed description such as "Fleet name for request", "Name of start waypoint", etc.
- it would be nice if the form cached previously inputted values. Right now when I switch to the `Info` tab and back, the form is reset. But not a priority and sending the actual Loop msg would be more imp.

baviera08:
Thanks for the feedback @Yadunund :)

koonpeng:
Given that there is a deadline for this, I am ok if anything as long as it works. Below are just my thoughts on the UI/UX, which should be only regarded once the required functionality is working.

* Regarding the label on top of the field, see if we can use material ui https://material-ui.com/components/text-fields/. This should also fix the font.
* The cancel button looks redundant.
* paddings, margins, alignment etc: The text field could expand to fill more of the empty space, some margin above the buttons would be nice. The buttons could also be better aligned, maybe in the center or filling the width.
* The scheduler visualizer seems to be unloaded or hidden when the popup appears.
* Theming for the robot info tab looks to be broken.

baviera08:
> Given that there is a deadline for this, I am ok if anything as long as it works. Below are just my thoughts on the UI/UX, which should be only regarded after the required functionality is working.
> 
> * Regarding the label on top of the field, see if we can use material ui https://material-ui.com/components/text-fields/. This should also fix the font.
> * The cancel button looks redundant.
> * paddings, margins, alignment etc: The text field could expand to fill more of the empty space, some margin above the buttons would be nice. The buttons could also be better aligned, maybe in the center or filling the width.
> * The scheduler visualizer seems to be unloaded or hidden when the popup appears.
> * Theming for the robot info tab looks to be broken.


Thanks for the comments @koonpeng . As you said, once this is actually requesting the loop, I'll address those suggestions.

baviera08:
I need a new release of `romi-js-core-interfaces` to have `RomiCore.loopRequest`

koonpeng:
The latest version (0.0.2-alpha.7) should have this.

baviera08:
> The latest version (0.0.2-alpha.7) should have this.

Thanks!

baviera08:
I tested this with 
* rmf_core: master
* rmf_demos: master
* rmf_schedule_visualizer:master
* traffic_edito:master

src/components/robot-item-form.tsx
```diff
@@ -0,0 +1,157 @@
+import { makeStyles, TextField, Button } from '@material-ui/core';
+import Autocomplete from '@material-ui/lab/Autocomplete';
+import React, { useState, useEffect } from 'react';
+import { successMsg } from '../util/alerts';
+import fakePlaces from '../mock/data/places';
```
koonpeng, line 5: There used to be a places field in `BuildingMap` but I think it's deprecated and no longer contains anything. @Yadunund  Do you know where should the places be acquired from now?

Once we can query the list of places, we should use that instead of a mock.

src/components/robot-item-form.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const [numLoopsError, setNumLoopsError] = useState('');
+  const [startLocationError, setStartLocationError] = useState('');
+  const [finishLocationError, setFinishLocationError] = useState('');
+
+  useEffect(() => {
+    const listOfPlaces = fakePlaces()[fleetName];
+    setListOfPlacesToGo(listOfPlaces);
+    !listOfPlacesToGo && console.error('List of places to go it`s empty');
+    setStartLocation(listOfPlaces[0]);
+    setFinishLocation(listOfPlaces[1]);
```
koonpeng, line 36: We should check if `listOfPlaces` has length >= 2

src/components/robot-item-form.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+  const [numLoops, setNumLoops] = useState(0);
+  const [startLocation, setStartLocation] = useState('');
+  const [finishLocation, setFinishLocation] = useState('');
+  // Error states
+  const [numLoopsError, setNumLoopsError] = useState('');
+  const [startLocationError, setStartLocationError] = useState('');
+  const [finishLocationError, setFinishLocationError] = useState('');
+
+  useEffect(() => {
```
koonpeng, line 31: Instead of using an effect, it might be better to use lazy state initialization or receiving the list of places as a prop to avoid double rendering on the initial mount.

src/components/robot-item-form.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const handleRequestLoop = (event: any) => {
+    event.preventDefault();
+    if (isformValid()) {
+      requestLoop(fleetName, numLoops, startLocation, finishLocation);
+      successMsg('Success');
+      cleanUpForm();
+    }
+  };
+
+  const isformValid = () => {
```
koonpeng, line 62:
```suggestion
  const isFormValid = () => {
```
probably typo

src/components/robot-item-form.tsx
```diff
@@ -0,0 +1,157 @@
...
+    padding: '0.5rem',
+    display: 'flex',
+    flexDirection: 'column',
+  },
+  divForm: {
+    padding: '0.5rem',
+    width: '100%',
+  },
+  error: {
+    color: 'red',
```
koonpeng, line 144:
```suggestion
    color: theme.palette.error.main,
```
We can use the theme color here.

src/components/robot-item-form.tsx
baviera08: From our previous discussion with Yadu

The start_name and finish_name should contain valid waypoint names as listed in BuildingMap->levels->nav_graphs[i]->vertices[j]->name. However as there is no public mapping between the nav_graph index (i) and the fleet name, it is not possible to know which fleet of robots uses which nav graph from the BuildingMap msg.

A node on Places:
BuildingMap->levels->places is not currently populated. The use of places is kind of deprecated. The information of locations that a robot can travel to is stored as a GraphNode in BuildingMap->levels->nav_graphs[i]->vertices. This is because there are named waypoints that Robot A can travel to but not Robot B.

This is a workaround, until we figure it out a good way to get that information (Through a message, through a service, etc). But please correct me if I'm wrong.

src/components/robot-item-form.tsx
baviera08: Yeah! Thanks! Fixed!

src/components/robot-item-form.tsx
baviera08: Makes sense! Thanks!

src/components/robot-item-form.tsx
baviera08: Makes sense, I'll pass it as a prop. We will deal with that list in the future again when we can get the data from the backend. Thanks!

src/components/robot-item-form.tsx
baviera08: You're totally right! Thanks!

src/components/robot-item-form.tsx
Yadunund: @koonpeng I asked Matias to use a file (`places.ts`) to get a list of named waypoints for each fleet as this is not explicitly conveyed in the BuildingMap msg. We can iterate on other methods in the future. Perhaps the same service serving robot icons can fulfill this requirement.

baviera08:
@Yadunund Just a friendly reminder, I can't actually merge the PR, you guys have to do it. =)

PR #29, opened 2020-05-28
Fix for dashboard does not start after PR#21

### LINKED ISSUE #28: Dashboard does not start after PR#21
Following the merge of #21, starting the dashboard with a live backend leaves the browser in this state until shutdown.
 
image

baviera08:
Fixes #28 
Updated `romi-js-core-interfaces` to **"0.0.2-alpha.7"** and added `dispenser_states` on soss config

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1363,9 +1363,9 @@
       "integrity": "sha512-shAmDyaQC4H92APFoIaVDHCx5bStIocgvbwQyxPRrbUY20V1EYTbSDchWbuwlMG3V17cprZhA6+78JfB+3DTPw=="
     },
     "@osrf/romi-js-core-interfaces": {
-      "version": "0.0.2-alpha.5",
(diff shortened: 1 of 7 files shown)


koonpeng:
@Yadunund can you help test and approve this if it works, I am on a feature branch of `rmf_core` and would rather avoid recompiling stuffs.

Yadunund:
Hi @baviera08 ,

I am able to open the dashboard and visualize the doors/lifts on canvas now. Two issues to highlight:

1. The `main_door` which is a "DoubleSwing" type is not drawn. 
 
image


image

2. The `yaw` of the lift cabin is not considered when drawing on canvas. It should be rotated by 90deg as defined in the BuildingMap
image

baviera08:
Hi @Yadunund , Thanks for the feedback.
image

Fixed DobleHingeDoor draw and the lift rotation.

baviera08:
@Yadunund I don't have permission to merge it, you have to do it.

PR #32, opened 2020-05-28
Improved lift UX #30

### LINKED ISSUE #30: Lift Visualization Enhancements
Hi @baviera08 ,

I would like to share some feedback on the visualization of lifts after running some trials.

Some general comments:
- I find the yellow of the cabin slightly alarming. Could we use something more sombre like grey instead? (personally I like the shade of grey used when the lift is not initialized)
- I think it adds more value if the `current_level` of the lift can be printed in place of `current_mode` ("Unknown" in images below). 

1.Screenshot below of lift in moving state:
image
-  Would be be possible to increase the transparency of the cabin so it looks like a lighter shade of grey when moving? This would help convey the `current_mode`. 
- The printout of the destination is great but it tends to get covered when the robot is inside the lift. Could we have it printed where the "Unknown" currently is? So when the lift is stationary at a level, the printout is the current_level name but if the lift is moving, it prints the destination text (L1->L3 text) at the same location.

baviera08:
Fixes #30 

## What's new
* We now use dark grey when the lift is not initialized.
* If the mode of the lift is `UNKNOWN` and if the lift has defined states, the current_level will be shown instead of the `UNKNOWN` label.
* When the lift is stationary at a level, the printout is the current_level name but if the lift is moving, it prints the destination text (L1->L3) at the same location.
* Switched places of the LiftMotion text and LiftMode text

### DIFF
--- a/src/components/schedule-visualizer/lift.tsx
+++ b/src/components/schedule-visualizer/lift.tsx
@@ -27,21 +27,21 @@ const getLiftStyle = (classes: any, currentMode: number | undefined, isInCurrent
     MODE_OFFLINE,
     MODE_UNKNOWN,
   } = RomiCore.LiftState;
-  const isOffLine = currentMode === MODE_OFFLINE || currentMode === MODE_UNKNOWN;
   const isModeAGV = currentMode === MODE_AGV;
   const isModeHuman = currentMode === MODE_HUMAN;
 
   if (currentMode === MODE_EMERGENCY) return classes.emergency;
   if (currentMode === MODE_FIRE) return classes.fire;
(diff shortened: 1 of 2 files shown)


baviera08:
I still want to try this with the proper backend and changing the lift states

baviera08:
From Yadu
We can keep that there when the lift is moving. But remove the "L2"
 text when it is stationary. As this is now going to be on top of the lift
I think for those special modes, we can print the text in the middle. And show change in color

image

baviera08:
@Yadunund file conflicts solved, ready for review.

baviera08:
Adding some tests

PR #35, opened 2020-06-03
Fix/weird behavior on door opening

### LINKED ISSUE #31: Request doors (open/close) are failing
> Request doors and lift were failing. When I requested a door to open I got this error on the soss console Segmentation fault (core dumped)

I found the cause of this, see https://github.com/osrf/soss/issues/36. Particularly, it fails while converting `builtin_interfaces/Time` as this is the definition

```
int32 sec
uint32 nanosec
```

conversion of `int32` fails because of above bug.

_Originally posted by @koonpeng in https://github.com/osrf/romi-dashboard/issues/28#issuecomment-635199287_

baviera08:
Fixes #31 

For this to work, we need to use this branch https://github.com/osrf/soss/tree/fix/int-segfault-%2336 on `SOSS` project.

Obs: Make sure to use the `soss.yaml` file of this branch. 

openingdoor

### DIFF
--- a/src/components/doors-panel.tsx
+++ b/src/components/doors-panel.tsx
@@ -16,7 +16,7 @@ export default function DoorsPanel(props: DoorsPanelProps): JSX.Element {
   const doorRefs = React.useRef<Record<string, HTMLElement | null>>({});
   const [expanded, setExpanded] = React.useState<Record<string, boolean>>({});
   const doorRequestPub = React.useMemo(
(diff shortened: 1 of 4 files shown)


Yadunund:
Hi @baviera08 ,

I was able to successfully open/close doors with this fix.

However, it doesn't look like there is  a form for a sending [LiftRequest](https://github.com/osrf/rmf_core/blob/master/rmf_lift_msgs/msg/LiftRequest.msg) msg?
image

PR #36, opened 2020-06-03
Feature/add lift request form

baviera08:
### What's new

New form to request lifts

image

Lift's doors now showing on panel (but still without state)

image


baviera08:
CC: @Yadunund

src/components/lift-item/lift-item.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <TabPanel value={tabValue} index={0}>
+            <LiftInformation lift={lift} liftState={liftState} />
+          </TabPanel>
+          <TabPanel value={tabValue} index={1}>
+            {liftState?.available_floors && (
+              <LiftRequestForm
+                liftRequest={handleRequest}
+                doorStateList={LiftRequestManager.getListOfDoorModes()}
+                requestTypeList={LiftRequestManager.getListOfLiftRequestModes()}
+                destinationList={liftState?.available_floors}
```
koonpeng, line 89:
```suggestion
                destinationList={lift.levels}
```
Looks like the list of levels is available on the building map, we can use that so that we don't need to depend on the lift state being available.

src/components/lift-item/lift-item.tsx
```diff
@@ -0,0 +1,152 @@
...
+        </ExpansionPanelDetails>
+      </ExpansionPanel>
+    </React.Fragment>
+  );
+});
+
+interface LiftRequestMenuState {
+  lift: RomiCore.Lift;
+  anchor: PopoverPosition;
+}
```
koonpeng, line 102: Don't think we are using this anymore

src/components/lift-item/lift-item.tsx
```diff
@@ -0,0 +1,152 @@
...
+ * TODO: Show indicator why lift controls are disabled.
+ */
+
+import {
+  ExpansionPanel,
+  ExpansionPanelDetails,
+  ExpansionPanelProps,
+  ExpansionPanelSummary,
+  makeStyles,
+  PopoverPosition,
```
koonpeng, line 11: Don't think we are using this anymore

src/lift-state-manager.ts
```diff
@@ -18,5 +18,146 @@ export default class LiftStateManager extends EventEmitter<Events> {
...
+  static getListOfLiftRequestModes() {
+    let listOfModes: Array<any> = [];
+    this.liftModes.forEach(element => {
+      listOfModes.push(this.liftModeToString(element))
+    });
+    return listOfModes
+  }
+
+  static getAllDoorModes() {
+
```
koonpeng, line 99: formatting (empty line)

src/lift-state-manager.ts
```diff
@@ -18,5 +18,146 @@ export default class LiftStateManager extends EventEmitter<Events> {
...
+    RomiCore.LiftRequest.REQUEST_AGV_MODE,
+    RomiCore.LiftRequest.REQUEST_HUMAN_MODE,
+  ];
+
+  static doorModes = [
+    RomiCore.LiftRequest.DOOR_CLOSED,
+    RomiCore.LiftRequest.DOOR_OPEN,
+  ]
+
+
```
koonpeng, line 79: formatting (empty line)

src/lift-state-manager.ts
```diff
@@ -18,5 +18,146 @@ export default class LiftStateManager extends EventEmitter<Events> {
...
+  static getLiftRequestModes() {
```
koonpeng, line 80: avoid returning `any[]`

src/lift-state-manager.ts
```diff
@@ -18,5 +18,146 @@ export default class LiftStateManager extends EventEmitter<Events> {
...
+    let listOfModes: Array<any> = [];
+    this.liftModes.forEach(element => {
+      const key = this.liftModeToString(element);
+      const value = element;
+      listOfModes.push({ key, value })
+    });
+    return listOfModes
+  }
+
+  static getListOfLiftRequestModes() {
```
koonpeng, line 90: avoid returing `any[]`

src/lift-state-manager.ts
```diff
@@ -18,5 +18,146 @@ export default class LiftStateManager extends EventEmitter<Events> {
...
+      default:
+        return `Unknown (${motionState})`;
+    }
+  }
+
   private _liftStates: Record<string, RomiCore.LiftState> = {};
 }
+
+export class LiftRequestManager {
+  static liftModes = [
```
koonpeng, line 68:
```suggestion
  static readonly liftModes = [
```
Doesn't look like it should be modified

src/lift-state-manager.ts
```diff
@@ -18,5 +18,146 @@ export default class LiftStateManager extends EventEmitter<Events> {
...
+    RomiCore.LiftRequest.REQUEST_END_SESSION,
+    RomiCore.LiftRequest.REQUEST_AGV_MODE,
+    RomiCore.LiftRequest.REQUEST_HUMAN_MODE,
+  ];
+
+  static doorModes = [
```
koonpeng, line 74:
```suggestion
  static readonly liftModes = [
```
Doesn't look like it should be modified

src/lift-state-manager.ts
```diff
@@ -18,5 +18,146 @@ export default class LiftStateManager extends EventEmitter<Events> {
...
+
+  static getListOfLiftRequestModes() {
+    let listOfModes: Array<any> = [];
+    this.liftModes.forEach(element => {
+      listOfModes.push(this.liftModeToString(element))
+    });
+    return listOfModes
+  }
+
+  static getAllDoorModes() {
```
koonpeng, line 98: avoid returning `any[]`

src/lift-state-manager.ts
```diff
@@ -18,5 +18,146 @@ export default class LiftStateManager extends EventEmitter<Events> {
...
+    let listOfModes: Array<any> = [];
+    this.doorModes.forEach(element => {
+      const key = this.doorStateToString(element);
+      const value = element;
+      listOfModes.push({ key, value })
+    });
+    return listOfModes
+  }
+
+  static getListOfDoorModes() {
```
koonpeng, line 109: avoid returning `any[]`

src/lift-state-manager.ts
```diff
@@ -18,5 +18,146 @@ export default class LiftStateManager extends EventEmitter<Events> {
...
+    let listOfModes: Array<any> = [];
+    this.doorModes.forEach(element => {
+      listOfModes.push(this.doorStateToString(element))
+    });
+    return listOfModes
+  }
+
+  static liftModeToString(liftMode: number) {
```
koonpeng, line 117:
```suggestion
  static requestModeToString(requestMode: number) {
```
this is confusing with `LiftStateManager.liftModeToString()`

src/lift-state-manager.ts
```diff
@@ -18,5 +18,146 @@ export default class LiftStateManager extends EventEmitter<Events> {
...
+    }
+  }
+
+  static StringToDoorState(doorState: string) {
+    switch (doorState) {
+      case 'Open':
+        return RomiCore.LiftState.DOOR_OPEN;
+      case 'Closed':
+        return RomiCore.LiftState.DOOR_CLOSED;
+    }
```
koonpeng, line 160: I would avoid these functions, converting string back to number looks flaky, there could be many edge cases like caps, whitespaces, multi-language support that makes this unreliable. It seems its only used in the lift request form, it can receive the number and string in the props so it does not need to resort to this.

src/lift-state-manager.ts
```diff
@@ -18,5 +18,146 @@ export default class LiftStateManager extends EventEmitter<Events> {
...
+    });
+    return listOfModes
+  }
+
+  static liftModeToString(liftMode: number) {
+    switch (liftMode) {
+      case RomiCore.LiftRequest.REQUEST_END_SESSION:
+        return 'End Session';
+      case RomiCore.LiftRequest.REQUEST_AGV_MODE:
+        return 'AVG';
```
koonpeng, line 122:
```suggestion
        return 'AGV';
```
typo

src/components/app.tsx
```diff
@@ -303,6 +303,10 @@ export default function App(props: AppProps): JSX.Element {
     setCurrentView(OmniPanelViewIndex.Dispensers);
   }
 
+  const doorsFromLifts = React.useMemo(() => {
+    return lifts.flatMap(x => x.doors);
+  }, [lifts]);
+
```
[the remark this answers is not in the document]
Yadunund: If by that you mean seeing the lift doors under the `Doors` panel and being able to request a lift door to open, then no. The visualizations can be the same however.

src/components/lift-item/lift-item.tsx
Yadunund: @koonpeng there may be scenarios where a lift only serves a select number of floors in the building. This information is captured in the LiftState msg.

koonpeng: https://github.com/osrf/traffic_editor/blob/df9e58434e03010ace631234fff2fc0190bda5b2/building_map_msgs/msg/Lift.msg#L2 I see a `levels` field here, doesn't it include only the levels the lift can go to?

src/components/app.tsx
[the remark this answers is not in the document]
koonpeng: Yes, that is what this code does, if this is the case can you remove this @baviera08?

src/components/lift-item/lift-item.tsx
Yadunund: Ah my bad for overlooking that. Yes that field will contain the list of levels the lift can access.

baviera08:
> The tab bar can fill the width


It looks like this with padding 
image

It looks like this with full width:
image

koonpeng:
> > The tab bar can fill the width
> 
> It looks like this with padding
> image
> 
> It looks like this with full width:
> image

sry I was referring to the "Info" and "Request" headers, there are unused white spaces to the right. For the content, I agree that having some padding looks better.

src/components/lift-item/lift-item-form.tsx
```diff
@@ -0,0 +1,152 @@
...
+          onChange={(e, value) => setDestination(value || '')}
+          options={destinationList}
+          renderInput={params => (
+            <TextField {...params} label="Pick a Destination" variant="outlined" />
+          )}
+          value={!!destination ? destination : null}
+        />
+        {destinationError && (
+          <p id="destinationError" className={classes.error}>
+            {destinationError}
```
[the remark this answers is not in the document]
koonpeng: Can you use the `renderInput` prop to set the error state on the `TextField`?

src/components/lift-item/lift-item-form.tsx
```diff
@@ -0,0 +1,152 @@
...
+  };
+
+  const handleLiftRequest = (event: any) => {
+    event.preventDefault();
+    if (isFormValid()) {
+      const liftDoor = LiftRequestManager.StringToDoorState(doorState);
+      const liftRequestType = LiftRequestManager.StringToLiftMode(requestType);
+      if (liftDoor !== undefined && liftRequestType !== undefined) {
+        liftRequest(liftDoor, liftRequestType, destination);
+        successMsg('Success');
```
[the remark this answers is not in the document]
koonpeng: In the future versions the UI may be able to get some feedback, for example, the UI can listen to a `SupervisorHeartBeat` message to know if its message has been received. We could leave this alone for now until such a feature is available.

baviera08:
How about this?
image

koonpeng:
> How about this look?
> image

yep looks good!

src/components/lift-item/lift-item-form.tsx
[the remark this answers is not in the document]
baviera08: Wow, That's weird it's working now, I already tried to do it that way. 

image

baviera08:
Thanks for the review @koonpeng  @Yadunund Let me know if you have more comments.

Yadunund:
Hi @baviera08 ,

I tried testing with live backend but the form does not have a list of available floors in the Destination Floor drop down

image


This is what the LiftState looks like
image

Yadunund:
When I build + run your `beta` branch I can see the list of floors populated in the form.

image

baviera08:
> Hi @baviera08 ,
> 
> I tried testing with live backend but the form does not have a list of available floors in the Destination Floor drop down
> 
> image
> 
> This is what the LiftState looks like
> image

Hi @Yadunund , I guess that happened because you have no lift levels on the building-map (I don't have a way to reproduce it this but I think that's what's happening). We had the same problem for the demo, so I made this change to fix it. 
```
destinationList={lift.levels}
```
for this
```
destinationList={liftState?.available_floors}
```
But in this comment https://github.com/osrf/romi-dashboard/pull/36#discussion_r437833650 we agreed with @koonpeng  that we will use the levels of the building map and not the one from the LiftState (correct me if I'm wrong), so I changed back.

I think we are not on the same page here or maybe I'm missing something here, which direction should I go?

Yadunund:
> But in this comment [#36 (comment)](https://github.com/osrf/romi-dashboard/pull/36#discussion_r437833650) we agreed with @koonpeng that we will use the levels of the building map and not the one from the LiftState (correct me if I'm wrong), so I changed back.
> 
> I think we are not on the same page here or maybe I'm missing something here, which direction should I go?

Hi @baviera08 , you're right I was using a map with only one level defined where the lift's behavior is mocked. I apologize for the confusion with my previous comment. We should stick to using the information from `BuildingMap.lifts` as agreed upon earlier. 

I'm wondering if there is way to make the dashboard also compatible with maps that are configured as such. Is there a way to make the Destination floor field accept values from the dropdown list as well as manual entry by users?

baviera08:
> > But in this comment [#36 (comment)](https://github.com/osrf/romi-dashboard/pull/36#discussion_r437833650) we agreed with @koonpeng that we will use the levels of the building map and not the one from the LiftState (correct me if I'm wrong), so I changed back.
> > I think we are not on the same page here or maybe I'm missing something here, which direction should I go?
> 
> Hi @baviera08 , you're right I was using a map with only one level defined where the lift's behavior is mocked. I apologize for the confusion with my previous comment. We should stick to using the information from `BuildingMap.lifts` as agreed upon earlier.
> 
> I'm wondering if there is way to make the dashboard also compatible with maps that are configured as such. Is there a way to make the Destination floor field accept values from the dropdown list as well as manual entry by users?

We can do that or I can check whether the `BuildingMap.lifts` it's empty, if it's empty we can use `liftState?.available_floors` instead.

Yadunund:
> We can do that or I can check whether the `BuildingMap.lifts` it's empty, if it's empty we can use `liftState?.available_floors` instead.

I'm in favor of the former. @koonpeng what do you think?

koonpeng:
> Hi @baviera08 , you're right I was using a map with only one level defined where the lift's behavior is mocked. I apologize for the confusion with my previous comment. We should stick to using the information from BuildingMap.lifts as agreed upon earlier.

would it be possible to make the building map report the lift levels you want in the mock? Using lift states creates a weird situation where we know there is a lift but we don't know where it can go. I don't like allowing manual entry, I feel it complicates the UX and there shouldn't be any scenario in an actual deployment that we want to let the user input an arbitrary value.

Yadunund:
> would it be possible to make the building map report the lift levels you want in the mock? Using lift states creates a weird situation where we know there is a lift but we don't know where it can go. I don't like allowing manual entry, I feel it complicates the UX and there shouldn't be any scenario in an actual deployment that we want to let the user input an arbitrary value.

That is a fair point about complicating the UX. I will update the mock I'm using instead. We can leave the behavior as it is here.

baviera08:
Hi @Yadunund @koonpeng It seems that all comments are resolved and addressed. Do we want to add/modify another feature or it's ready to get it into master?

Issue #9, opened 2020-05-04
Warn about conflicting trajectories between robots


matiasbavera:
Create an alert when two robots have a conflicting trajectory (this alert could disappear automatically after some time)

Mark the as red the path and robot that has conflicting trajectories. 

Nice to have:
We can also make the robot bigger too when it has conflicting trajectories.

baviera08:
If possible we should test this on the new “participant system”:
* rmf_core: develop 
* rmf_scheduler_visualizer: develop 

We should test that way because the old system does not produce the time of the conflict. Ideally we want to show the exact point of conflict instead of making the world path red. This goes in line with a separate improvement that customizes the trajectory based on the robot (different colors/patterns) to allow the trajectories to be easily differentiated when there are large amounts of robots in the system.

baviera08:
* I did not find a way to set the color to the robot when it has a conflicting path. Because te robot and trajectory have independent messages. The idea here was to make the robot bigger and change its color when it has path conflicts.

cnboonhan94:
https://cgh.webex.com/meet/CT_CHART

PR #40, opened 2020-06-19
Feature/warn about conflicting trajectories#9

### LINKED ISSUE #60: Lack of information of which robot collided with which
Hi @Yadunund, in order to merge this PR https://github.com/osrf/romi-dashboard/pull/40 . We need to discuss and solve an issue related to the robot trajectory conflict.

The backend is sending us the robot trajectory information in the following format:

```
{
  "response":"trajectory",
  "conflicts": [310,311,312,313]
  "values":[
{...},
{...},
{...},
{...},
],
}
```

The problem here is that for example if robot A conflicts with robot B (310,311) and robot C conflicts with robot D (312,313), it would print A, B, C, D conflicting when there are actually 2 separate conflicts going on.

image

As a proposal, can the backend send us a Trajectory message with the Conflict props being an array of conflicts ids.
Current = `[1,2,3,4]`
My proposal =` [[1,2],[2,3],[1,4]]`

What do you think about it? Do you have another solution in mind? 

CC: @koonpeng

matiasbavera:
Pre-requisite we can merge this PR after https://github.com/osrf/rmf_schedule_visualizer/pull/65 is merged 

Fixes #60 and #9 

## What's new
* Added a notification bar for different uses on the main page.
* Updated trajectory message definitions with https://github.com/osrf/rmf_schedule_visualizer/pull/43
* Name of robots showed in a notification bar on a trajectory conflict.

Notification bar with conflicts message.
image

## Discussion
*  Do I need to add the Fleet + RobotName or just the RobotName to the collision message?


baviera08:
@koonpeng Can you give me some feedback on the UI for this first version?

koonpeng:
the latest change now shows the robot that the trajectory belongs to, it can be used to improve the conflict popup https://github.com/osrf/rmf_schedule_visualizer/pull/43.

What do the different colors of the trajectory mean?

I'm trying to understand the code to find the conflicting segments, my understanding is that it checks all conflicting trajectory segments' (x, y) positions and mark segments with the same positions as conflicting. Would it catch a situation where the conflicts don't share segments, e.g.
```
   o
   |
o-----
   |
   |
```
if not, is there a better way to find the conflicting positions? Or is this situation even possible? @Yadunund

koonpeng:
On a side note, I noticed you also have clipping when using the "outline" trajectory animation, I thought its just because of my setup with fractional scaling, I'm going to open an issue on this, its not important but you are welcomed look into it if you have the time.

baviera08:
> the latest change now shows the robot that the trajectory belongs to, it can be used to improve the conflict popup [osrf/rmf_schedule_visualizer#43](https://github.com/osrf/rmf_schedule_visualizer/pull/43).

Cool, I'll add that.



> What do the different colors of the trajectory mean?

It means the different robot trajectories. 
Robot A from P1 to P2 -> Color 1
Robot B from P1 to P3 -> Color 2
Robot C from P3 to P4 -> Color 3
Robot A from P2 to P5 -> Color 4 


> I'm trying to understand the code to find the conflicting segments, my understanding is that it checks all conflicting trajectory segments' (x, y) positions and mark segments with the same positions as conflicting. Would it catch a situation where the conflicts don't share segments, e.g.
> 
> ```
>    o
>    |
> o-----
>    |
>    |
> ```
> 
> if not, is there a better way to find the conflicting positions? Or is this situation even possible? @Yadunund

Yeah, you're right. In that case, it wouldn't catch it. I can change the whole trajectory color (as it's now in master) for that particular case. But it worth checking first with @Yadunund.

Thanks for the feedback.

baviera08:
> On a side note, I noticed you also have clipping when using the "outline" trajectory animation, I thought its just because of my setup with fractional scaling, I'm going to open an issue on this, its not important but you are welcomed look into it if you have the time.

Yeah sure! I'll tackle this in another PR.

Yadunund:
> 'm trying to understand the code to find the conflicting segments, my understanding is that it checks all conflicting trajectory segments' (x, y) positions and mark segments with the same positions as conflicting. Would it catch a situation where the conflicts don't share segments, e.g.
> 
> ```
>    o
>    |
> o-----
>    |
>    |
> ```
> 
> if not, is there a better way to find the conflicting positions? Or is this situation even possible? @Yadunund

Two trajectories from participants A & B  are in conflict if A's [`Profile::footprint`](https://github.com/osrf/rmf_core/blob/master/rmf_traffic/include/rmf_traffic/Profile.hpp) when swept along A's trajectory overlaps with B's vicinity swept along B's trajectory or vice versa. This can be checked using this [API](https://github.com/osrf/rmf_core/blob/0f155e9efea5e82d0da6723c45ee54964eb3cef5/rmf_traffic/include/rmf_traffic/DetectConflict.hpp#L54)
Checking for matching pair of (x,y) segments between two trajectories is not a good approach
* Trajectories are represented as a collection of cubic splines which are interpolated between these segment knots. Hence, two trajectories can intersect in between these segment knots. 
* While a set of segments may overlap in position, they may not do so in the time domain which can lead to incorrect conflict assumption with the above approach.

Afik the time of conflict is not reported in any of the messages published by the `rmf_traffic_schedule`. Correct me if i'm wrong @mxgrey 

As a solution, we have have the websocket server use the above mentioned API to determine the time of conflict and communicate this to the front end. With the time of conflict, you should be able to determine the position of conflict for each of the conflicting trajectories. Let me know what you think.

baviera08:
> > 'm trying to understand the code to find the conflicting segments, my understanding is that it checks all conflicting trajectory segments' (x, y) positions and mark segments with the same positions as conflicting. Would it catch a situation where the conflicts don't share segments, e.g.
> > ```
> >    o
> >    |
> > o-----
> >    |
> >    |
> > ```
> > 
> > 
> > if not, is there a better way to find the conflicting positions? Or is this situation even possible? @Yadunund
> 
> Two trajectories from participants A & B are in conflict if A's [`Profile::footprint`](https://github.com/osrf/rmf_core/blob/master/rmf_traffic/include/rmf_traffic/Profile.hpp) when swept along A's trajectory overlaps with B's vicinity swept along B's trajectory or vice versa. This can be checked using this [API](https://github.com/osrf/rmf_core/blob/0f155e9efea5e82d0da6723c45ee54964eb3cef5/rmf_traffic/include/rmf_traffic/DetectConflict.hpp#L54)
> Checking for matching pair of (x,y) segments between two trajectories is not a good approach
> 
> * Trajectories are represented as a collection of cubic splines which are interpolated between these segment knots. Hence, two trajectories can intersect in between these segment knots.
> * While a set of segments may overlap in position, they may not do so in the time domain which can lead to incorrect conflict assumption with the above approach.
> 
> Afik the time of conflict is not reported in any of the messages published by the `rmf_traffic_schedule`. Correct me if i'm wrong @mxgrey
> 
> As a solution, we have have the websocket server use the above mentioned API to determine the time of conflict and communicate this to the front end. With the time of conflict, you should be able to determine the position of conflict for each of the conflicting trajectories. Let me know what you think.


Yeah, you're right, thanks for the feedback. The API sounds good, I'll be able to point the collision point with that. 
Oh right! I'll create a GH issue and I'll tag you two for awareness. And let's tackle the position of conflict on another PR. I'll leave the collision markers as they were originally.

src/components/schedule-visualizer/robot-trajectories-overlay.tsx
```diff
@@ -26,6 +28,26 @@ export default function RobotTrajectoriesOverlay(
...
+
+        robotName && robotNames.push(robotName);
+      });
+      return robotNames;
+    }
+
+    if (conflicts.length !== 0) {
+      notificationDispatch &&
+        notificationDispatch({
+          message: `Trajectory conflict between ${getConflictRobotsName()}`,
```
koonpeng, line 45: If I understand this correctly, it prints robots in all conflicting trajectories. For example if robot A conflicts with robot B and robot C conflicts with robot D, it would print A, B, C, D conflicting when there is actually 2 separate conflicts going on.

src/components/schedule-visualizer/robot-trajectory.tsx
```diff
@@ -19,51 +20,44 @@ export const RobotTrajectory = React.forwardRef(function(
...
-    return d;
-  }, [trajectory]);
-
-  const color = React.useMemo(
+  const pathColor = React.useMemo(
     () =>
-      conflicts.includes(trajectory.id) ? theme.palette.error.main : theme.palette.success.main,
+      conflicts.includes(trajectory.id)
+        ? theme.palette.error.main
+        : ColorManager.getPathColor(trajectory.id),
```
koonpeng, line 27: If we are using a fixed color for the conflicting trajs and a calculated color for non-conflicting trajectories we should take care that the non-conflict color does not end up being too similar to the conflicting color.

src/robot-trajectory-manager.ts
```diff
@@ -95,9 +97,14 @@ export class DefaultTrajectoryManager {
     return resp as TimeResponse;
   }
 
+  static getRobotNameFromPathId(pathId: number, trajectories: readonly Trajectory[]): string | undefined {
```
koonpeng, line 100: this shouldn't be needed now that the trajectory contains the robot name.

src/components/schedule-visualizer/colors.ts
```diff
@@ -41,6 +37,23 @@ export default class ColorManager {
     return color ? color : null;
   }
 
+  /**
+   * Gets a light color for the path. The hash is an async function that takes a while to give a response. Since our application should graph the path in real time, the hash was discarded and the path id was chosen to calculate its color. An attempt was made on robot-trajectory.tsx to get the colors with the hash by wrapping it in a useEffect but the update time was not fast enough.
+   * @param pathId : id of the path
+   */
+  static getPathColor(pathId: number): string {
```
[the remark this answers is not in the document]
baviera08:
``` typescript
  async robotColor(name: string, model: string): Promise<string> {
    const key = this._robotKey(name, model);
    let color = this._robotColorCache[key];
    if (!color) {
      const modelHash = new Uint16Array(await _hash(model));
      const nameHash = new Uint16Array(await _hash(name));
      color = ColorManager._getLightColor(modelHash[0], nameHash[0]);
      this._robotColorCache[key] = color;
      this._robotColorCache[name] = color;
    }
    return color;
  }
```


Ok, I'm using the `robotColorFromCache` now, but the original function was using the `name` and `model` of the robot to build a key. In the trajectory case, we just have the `robot_name` and the `fleet_name`. What I did in the meanwhile it's to create a new key with just the name of the robot and with that, the trajectory has the same color as the Robot that as you said makes sense.

So, I need to ask some questions here:
* Is the name of the robot unique or just the key-value pair of robot + model are unique.
* Can I use the robot + model to create the color of the Robot and just use the robot_name as a key?. In case the robot names are unique of course.
* I think this is the less valid option and doesn't make a lot of sense but still worth asking, can we add to the robot model to the trajectory?

src/components/schedule-visualizer/robot-trajectories-overlay.tsx
baviera08: Yeah, that's right. I'll remove the word between. 

We can just mention the conflicting trajectories for now. Unless I check the collision points by coordinates, but we discard that option [here](https://github.com/osrf/romi-dashboard/pull/40#issuecomment-649438833).

Let me know what you think about it.

src/components/schedule-visualizer/robot-trajectory.tsx
baviera08: Makes sense. Will do. Thanks

src/components/schedule-visualizer/robot-trajectories-overlay.tsx
```diff
@@ -26,6 +28,26 @@ export default function RobotTrajectoriesOverlay(
   // FIXME: hardcode for now, as the source of the footprint is expected to change.
   const footprint = 0.5;
 
+  useEffect(() => {
+    function getConflictRobotsName(): string[] {
```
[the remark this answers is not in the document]
baviera08: Maybe I'm not following you, I should have left a comment there. What that function does is the following

If this, for example, is our trajectory message
``` JSON
{
  "response":"trajectory",
  "conflicts": [310,311]
  "values":[
{
    "robot_name":"magni1",
    "fleet_name:"magni",
    "shape":"circle",
    "dimensions":0.3,
    "id":310,
    "segments":[
      {"t":336857,"v":[0.018886815933120995,0.4996431607843137,0.0],"x":[11.610485884950174,-8.054053944114406,-1.6085801124572754]},
      {"t":338700,"v":[0.01888681593558913,0.4996431608496359,0.0],"x":[11.645306833719362,-7.132879672805034,-1.6085801124572754]}]},

{
    "robot_name":"magni2",
    "fleet_name:"magni",
    "shape":"circle",
    "dimensions":0.3,
    "id":311,
    "segments":[
      {"t":336857,"v":[0.018886815933120995,0.4996431607843137,0.0],"x":[11.610485884950174,-8.054053944114406,-1.6085801124572754]},
      {"t":338700,"v":[0.01888681593558913,0.4996431608496359,0.0],"x":[11.645306833719362,-7.132879672805034,-1.6085801124572754]}]},

{id:312, ...},
],
}
```

Context:
We have these components in which the RobotTrajectoryOverlay pass their properties to its child RobotTrajectory (it receives just one trajectory).
`RobotTrajectoryOverlay -> RobotTrajectory`

`getConflictRobotsName` gets just the conflicting trajectories (310,311) and after that, gets the name of the robots inside each trajectory (values: [{},{}]). The mentioned function returns a string[] with the name of robots with conflicts.
After getting those values, the `useEffect` dispatches a message to the notification Bar to show the message. In this case, _RobotTrajectoryOverlay_ component is in charge of getting the conflict values and handle the notification. 

Once said that, Maybe what you are trying to said to me it's for me pass the `traj` prop that already contains the `robot_name` to the trajectory component and handle the message in the child component _RobotTrajectory_. There is a couple of issues here:
* I'll just have access to the name of the current trajectory on _RobotTrajectory_.
* I can trigger the Notification Bar from there with just one conflict trajectory.
* In case we want to RobotTrajectoryOverlay handle the Notification Bar just to show one message with all conflicts, we can send a dispatcher function from `RobotTrajectoryOverlay -> RobotTrajectory`. With this dispatcher, the _RobotTrajectory_ component would be able to create and modify a list of names of robots with conflicting paths. But I believe that will be unnecessary expensive because every RobotTrajectory should have to control one prop more and the logic inside RobotTrajectory will be replicated as many times as trajectories exist. 

``` typescript
 {trajs.map(traj => (
          <trajectoryContext.Component
            key={traj.id}
            trajectory={traj}
            footprint={footprint}
            conflicts={conflicts}
            colorManager={colorManager}
          />
        ))}

```


Hey sorry for writing so much, I just want to make sure that we are on the same page here.

src/robot-trajectory-manager.ts
baviera08: Related to https://github.com/osrf/romi-dashboard/pull/40#discussion_r451169117

src/components/schedule-visualizer/robot-trajectory.tsx
baviera08: Done!

src/components/schedule-visualizer/colors.ts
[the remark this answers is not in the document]
koonpeng: > Is the name of the robot unique or just the key-value pair of robot + model are unique.

Yes, they should be unique.

> Can I use the robot + model to create the color of the Robot and just use the robot_name as a key?. In case the robot names are unique of course.

That makes sense, the names are unique so there shouldn't be any issues.

> I think this is the less valid option and doesn't make a lot of sense but still worth asking, can we add to the robot model to the trajectory?

There has been a bit of discussion to have a service to query robot information, it's not a priority so it probably won't be available in the current release of rmf. When we eventually implement that service, the `RobotState` will likely no longer contain the model, instead we will use that service to query it based on the robot name.

src/components/schedule-visualizer/robot-trajectories-overlay.tsx
koonpeng: Thats a good point, the current conflict data can't differentiate between multiple separate conflicts. @Yadunund Is the proposal possible to implement in the trajectory server?

src/components/schedule-visualizer/robot-trajectories-overlay.tsx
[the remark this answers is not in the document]
koonpeng: Thanks for the explanation! You are right, I missed that the conflict only contains the trajectory id so we still need to resolve that to the trajectory and get the robot name. The current way it's doing the notification looks good.

src/components/schedule-visualizer/colors.ts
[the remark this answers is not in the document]
baviera08: the color of robots are generated by name + model and the cache key is the name of the robot

baviera08:
Thanks for the feedback @koonpeng, changes done as requested.

koonpeng:
feel free to create an issue here or at rmf_schedule_visualizer to better facilitate discussions. In the meantime, I will put this on hold until the conflict information is resolved.

baviera08:
> feel free to create an issue here or at rmf_schedule_visualizer to better facilitate discussions. In the meantime, I will put this on hold until the conflict information is resolved.

Done! https://github.com/osrf/romi-dashboard/issues/60

matiasbavera:
@koonpeng ready to go once osrf/rmf_schedule_visualizer#65 is merged

matiasbavera:
osrf/rmf_schedule_visualizer#65 is merged. We should be ready to merge.

koonpeng:
Thanks for these, can we revert the changes on the trajectory colors and tackle it in its own PR? I feel we should get more input if we should do the changing colors.

matiasbavera:
> Thanks for these, can we revert the changes on the trajectory colors and tackle it in its own PR? I feel we should get more input if we should do the changing colors.

Makes sense! Done!

PR #47, opened 2020-06-30
Feature/create delivery request form#42

matiasbavera:
## What's new
* Added request delivery form
* Added Unit tests
* Added e2e test for delivery request.

delivery request


baviera08:
Hi @Yadunund and @koonpeng, can you give me any feedback on this? Thanks!

koonpeng:
After some discussion, I realized that it's not possible to target specific robots for delivery and loop requests, as such we have decided to move the request forms to its own panel. @vallq is working on this in the `feat/commands-panel` branch. At some point we will probably move the delivery request form over there as well.

baviera08:
> After some discussion, I realized that it's not possible to target specific robots for delivery and loop requests, as such we have decided to move the request forms to its own panel. @vallq is working on this in the `feat/commands-panel` branch. At some point we will probably move the delivery request form over there as well.

Thanks for the feedback. In that case, makes sense. Ok, once that `feat/commands-panel` is addressed I'll continue with this PR.

src/components/tests/delivery-form.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+  test('Dispensers cannot be empty', () => {
+    const wrapper = buildWrapper('FleetB', onClick);
+    wrapper.find('form').simulate('submit');
+    expect(
+      wrapper
+        .find('#pickupDispenser-helper-text')
+        .first()
+        .find('p')
+        .props().children,
+    ).toBe('Cannot be empty');
```
koonpeng, line 73: I think it may be too pedantic to check for the exact error string, it can cause problems if we ever have localization in the future, changing the error message would also cause it to fail. Is checking for the existence of error enough? Should we starting looking at a localization library like https://www.npmjs.com/package/i18next and check the error message matches a translation dictionary?

src/mock/data/dispensers.ts
```diff
@@ -0,0 +1,10 @@
+export default function fakeDispensers(): Record<string, string[]> {
+    /**
```
koonpeng, line 2: please reformat to 2-space indents

e2e/tests/deliveryRequest.test.ts
```diff
@@ -0,0 +1,54 @@
+import { RmfLauncher } from '../rmf-launcher';
+import { overwriteClick, removeTextFromAutocomplete, getRobotLocations } from './utils';
+
+describe('Delivery request', () => {
+    const launcher = new RmfLauncher();
```
koonpeng, line 5: please reformat to 2-space indents

src/components/tests/delivery-form.test.tsx
matiasbavera: Oh wow, I didn't know that, you're right about the localization problems. Ok, I changed it and I'm checking just the error existence for now. 
I thought the following when I wrote those tests, if an input could have more than one possible error, how do I ensure that is failing in the correct scenario? but now that you've mentioned I think that if this level of detail isn't required right now, especially if we'll add localization. 
I agree with you, we should start looking for a localization library and implement it before checking the error messages in the tests.
I'll create a new GH issue about localization.

Thanks for the feedback

src/mock/data/dispensers.ts
matiasbavera: Thank for noticing this. Done, this should be solved by husky in the future.

e2e/tests/deliveryRequest.test.ts
matiasbavera: Done!

src/components/tests/delivery-form.test.tsx
matiasbavera: https://github.com/osrf/romi-dashboard/issues/73

koonpeng:
e2e tests are failing for me, adding this seems to fix it
```ts
$('input[name=pickupPlace]').waitForClickable();
```

matiasbavera:
> e2e tests are failing for me, adding this seems to fix it
> 
> ```ts
> $('input[name=pickupPlace]').waitForClickable();
> ```

Thanks for the feedback. Added! It could be because of the animation I guess

PR #57, opened 2020-07-10
Add storybook to romi-dashboard

matiasbavera:
## What's new
* Added storybook
* Updated Typescript version
* Added storybook script

image

We added something call Storybook to the project. The storybook is a user interface, software development environment, and playground for UI components. The main purpose of it is to allow you to create components independently in an isolated development environment outside the main application. We can also use this to show our components in their different states.

Storybook runs on its own web server and provides a catalog view for the components for our application. We can run the storybook project with `npm run-script storybook` command.

Using the Storybook will also force us to split the component into two: (1) presentational components (the component with no logic behind); and (2) container components (where the component logic is. E.g: coordinates calculations, etc.). Which is a widely used design.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,0 +1,8 @@
+module.exports = {
+  stories: ['../src/**/*.stories.js'],
+  addons: [
+    '@storybook/preset-create-react-app',
+    '@storybook/addon-actions',
+    '@storybook/addon-links',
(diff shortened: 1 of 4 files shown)


matiasbavera:
> I'm just gonna merge this now to make future PRs easier
Cool! That's the idea.

PR #64, opened 2020-07-16
Feat/omnipanel styling

ChawinTan:
Major changes to the styling and design that I made:
- Change the robot item information font
- Replace Tab with Ant Tab in lift component to ensure consistency with Robot menu
- Made sure the 2 columns align in dispenser queue when overflow
- Add a border to "UNKNOWN" states
- Hide text and show ellipsis when the item name and states are too long to be displayed in the field and buttons

Attached are some photos of visual changes that I made. Appreciate your feedback on what can I improve on and if these styles make sense/suits your taste :)

dispenser_overflow_alignment
Dropdown_list
Lift_Ant_tab


koonpeng:
maybe it will look better with less padding on the contents and the tab bar filling the spaces and the tabs spaced evenly?

ChawinTan:
Removed some padding around the contents but I think having some padding around the content makes it looks nice. Centered the AntTabs so that the heading fills in evenly - is this what you are thinking about?

ChawinTan:
I see. So the dash below the AntTab info should be fully extended

ChawinTan:
Changed the variant of the AntTab to full width and made some changes to the content padding to ensure alignment with contents wrapped in the `Box` element

matiasbavera:
LGTM! 
Maybe it just my impression but, the color of the doors with the status Unknown really catches my eye when I'm looking at the list of doors. Is this the intention? Or the Unknown state should go more unnoticed (like some scale of gray). Should or can the end-user take some action knowing that the door status is unknown? do we need to get the end user's attention in this case?

src/components/dispenser-item.tsx
```diff
@@ -36,20 +37,28 @@ export const DispenserItem = React.forwardRef(function(
       case RomiCore.DispenserState.OFFLINE:
         return `${classes.dispenserLabel} ${dispenserModeLabelClasses.offline}`;
       default:
-        return '';
+        return `${classes.dispenserLabel} ${dispenserModeLabelClasses.unknown}`;
     }
   }
 
-  function dispenserRequestQueueId(): React.ReactElement{
+  function dispenserRequestQueueId(): JSX.Element {
```
matiasbavera, line 44: Just checking, Is this JSX.Element or React.ReactElement?

ChawinTan: Should be React.Element. Just changed it

ChawinTan:
I guess for the `unknown` button, I wanted to find a color that looks different from the rest of the border colors so that users know that is of unknown state

https://material-ui.com/customization/palette/

matiasbavera:
> I guess for the `unknown` button, I wanted to find a color that looks different from the rest of the border colors so that users know that is of unknown state
> 
> https://material-ui.com/customization/palette/

Yeah, I get it, let's with @koonpeng too. He can give us more context.

koonpeng:
I'm fine with the current colors (grey), can't remember what is the previous color. But I noticed that it is ellipsis-ed, do you have any suggestion to avoid this?

I can think of the following:
* use "N/A"
* expand omnipanel width (need to think about how it affects mobile ui)
* decrease font size
* auto scroller

The door/lift etc names also suffer the same problem.

ChawinTan:
Yup, I deliberately ellipsis them if their name gets too long. I think reducing font size or shortening the name may not be the best since there might be longer names or modes in the future.

Another way if ellipsis is not desired is to use word wrap and make the extra characters flow to the next line

koonpeng:
another thing we could do is have the summary be ellipsed but add another field in the details to show the full name. We should probably add this to the storybook as well to show how long names are rendered.

ChawinTan:
That's a good idea!

ChawinTan:
Changed Unknown to N/A
Changed the fonts on the summary to h6
added a `name` column to the field details.

PR #65, opened 2020-07-17
Feature/storybook components

ChawinTan:
Keeping this as a draft first. 

@matiasbavera don't mind if you take a look at my current design and implementation of the lift component. Attached is a screenshot of what it currently looks like.

We can use this PR to standardize the features that would be included in other components. 

LiftStories
storybook_screen

### DIFF
--- a/src/stories/lift.stories.js
+++ b/src/stories/lift.stories.js
@@ -82,6 +82,24 @@ const UNKOWNState = {
     motion_state: 3
 }
 
+const HUMANState = {
+    ... liftStates,
+    current_mode: 1,
+    current_floor: 'L3',
+    destination_floor: 'L1',
+    motion_state: 2
+}
+
+const OFFLINEState = {
+    ... liftStates,
+    current_mode: 4,
+}
+
+const EMERGENCYState = {
+    ... liftStates,
+    current_mode: 5
+}
+
 const renderInfoPanel = (mode, doorState, motionState) => {
     return (
         <div style={styles.modeInfoPanel}>
@@ -151,6 +169,54 @@ export const State_UNKNOWN = () => (
                     lift={lift}
                     currentFloor={UNKOWNState.current_floor}
                     liftState={UNKOWNState}
+                />
+            </svg>
+        </div>
+    </div>
+)
+
+export const State_HUMAN = () => (
+    <div style={styles.display}>
+        { renderInfoPanel('HUMAN', 'Closed', 'Down') }
+
+        <div>
+            <svg viewBox={'0 0 25.794363144785166 14.53525484725833'}>
+                <Lift 
+                    lift={lift}
+                    currentFloor={HUMANState.current_floor}
+                    liftState={HUMANState}
+                />
+            </svg>
+        </div>
+    </div>
+)
+
+export const State_OFFLINE = () => (
+    <div style={styles.display}>
+        { renderInfoPanel('OFFLINE', 'Closed', 'Stopped') }
+
+        <div>
+            <svg viewBox={'0 0 25.794363144785166 14.53525484725833'}>
+                <Lift 
+                    lift={lift}
+                    currentFloor={OFFLINEState.current_floor}
+                    liftState={OFFLINEState}
+                />
+            </svg>
+        </div>
+    </div>
+)
+
+export const State_EMERGENCY = () => (
+    <div style={styles.display}>
+        { renderInfoPanel('EMERGENCY', 'Closed', 'Stopped') }
+
+        <div>
+            <svg viewBox={'0 0 25.794363144785166 14.53525484725833'}>
+                <Lift 
+                    lift={lift}
+                    currentFloor={EMERGENCYState.current_floor}
+                    liftState={EMERGENCYState}
                 />
             </svg>
         </div>
--- a/src/stories/lift.stories.js
+++ b/src/stories/lift.stories.js
@@ -1,9 +1,25 @@
 import React from 'react';
+import { Divider, Typography } from '@material-ui/core';
 
 import Lift from '../components/schedule-visualizer/lift';
 
 export default { 
     title: 'Lift',
+};
+
+const styles = {
+    display: {
+        display: 'grid',
+        gridTemplateColumns: '1fr 3fr'
+    },
+    modeInfoPanel: {
+        padding: '2rem'
+    },
+    modeInfoItem: {
+        display: 'flex',
+        justifyContent: 'space-between',
+        padding: '0.5rem'
+    }
 };
 
 const lift = {
@@ -20,7 +36,7 @@ const lift = {
             v2_y: -4.5
         },
     ],
-    levels: ['L1', 'L2', 'L3'],
+    levels: ['L1', 'L2', 'L3', 'L4'],
     name: 'Lift1',
     ref_x: 7.1,
     ref_y: -2.8,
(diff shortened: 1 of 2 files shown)


src/stories/lift.stories.js
```diff
@@ -0,0 +1,224 @@
+import React from 'react';
```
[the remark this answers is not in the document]
ChawinTan: ah yes ...

src/stories/lift.stories.js
```diff
@@ -0,0 +1,224 @@
...
+            <svg viewBox={'0 0 25.794363144785166 14.53525484725833'}>
+                <Lift
+                    lift={lift}
+                    currentFloor={AGVState.current_floor}
+                    liftState={AGVState}
+                />
+            </svg>
+        </div>
+    </div>
+);
```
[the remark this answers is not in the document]
ChawinTan: Sure!

ChawinTan:
@matiasbavera @koonpeng  added some more details to the component story. Do you think it is necessary? Also, do you want to commit all the components together or do it one by one? eg: (lift first, followed by doors ... etc).

matiasbavera:
> @matiasbavera @koonpeng added some more details to the component story. Do you think it is necessary? Also, do you want to commit all the components together or do it one by one? eg: (lift first, followed by doors ... etc).

Hello @ChawinTan, as it suits you best. (This is just my opinion and maybe I'm wrong) I think it depends a lot on the situation, eg:
* if you are working on something that has a strong correlation between the functionalities:
     - If you receive constant and quick reviews you could have related PRs without problems. But if you don't, for example, if you have a PR A, and a PR B created from PR A and a PR C created from PR B, if a change is requested for PR A you will have to rebase B and C respectively. 
- If you are not going to receive constant and quick reviews, it is advisable to do a related large PRs. To avoid the problem mentioned before.
* If you are working with decoupled things, it is advisable to do small PRs. "Divide and conquer". =)

Well, let me know if you find some problem with my logic.  =)

I hope this helps.

src/stories/lift-component.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import Lift from '../components/schedule-visualizer/lift';
+
+export interface LiftComponentProps {
+    currentFloor: string;
+    lift: RomiCore.Lift;
+    liftState: RomiCore.LiftState;
+    renderInfoPanel(): JSX.Element;
+    currentMode(): JSX.Element;
+    motionState(): JSX.Element;
```
matiasbavera, line 12: Are these type JSX.Element or React.ReactElement?

This is an interesting link:
https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement

koonpeng: Should we also prefix these with `render*` to make things easier to understand?

src/stories/lift.stories.tsx
```diff
@@ -0,0 +1,250 @@
...
+    lift_name: "Lift1",
+    lift_time: {sec: 0, nanosec: 0},
+    motion_state: 0,
+    session_id: "",
+}
+
+const AGVState = {
+    ... liftStates,
+    current_mode: 2,
+    door_state: 2,
```
koonpeng, line 65: Can we use romi-js constants instead of literals? See https://osrf.github.io/romi-js-core-interfaces/classes/liftstate.html#door_closed.

src/stories/lift-component.tsx
ChawinTan: The functions are JSX.Element since they have `any` for their props and type

src/stories/lift.stories.tsx
ChawinTan: Yup sure. Thanks for pointing that out

ChawinTan:
https://storybook.js.org/docs/testing/structural-testing/

Is there a need for unit testing of these storybook components?

matiasbavera:
> https://storybook.js.org/docs/testing/structural-testing/
> 
> Is there a need for unit testing of these storybook components?

I think there is no need to do that with the Storybook components. I mean, there is no need because we are already doing snapshot tests to the components itself. 

You can check an example here https://github.com/osrf/romi-dashboard/blob/master/src/components/schedule-visualizer/tests/lift.test.tsx

You are more than welcome to add more test cases if you see it's necessary.

HTH.

ChawinTan:
I think these are the most important components for now.

Here are the stories that I have documented so far:
- Doors
- Lift
- Robots
- Notification Bar
- Button colors for various states
- display for long names
- display for Unknown states

Are there any more components that you guys feel should be added? @matiasbavera  @koonpeng

src/stories/BaseComponents/button-colors.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+    const { state, style } = props;
+
+    return (
+        <div style={rootStyles}>
+            <Typography variant="body1">{state}</Typography>
+            <Typography style={style} variant="button">{state}</Typography>
+        </div>
+    )
+}
```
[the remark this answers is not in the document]
ChawinTan: You mean refactor out the code to display the name and button on the expansion panel? I think that's a good idea. Let me create a task and look into it.

ChawinTan:
> Thanks @ChawinTan ! Looking good! Left a couple of comments.
> 
> Also, what do you think about this:
> 
> * We can create a basic palette of colors to use. (F.E: the ones you used when you re-styled the Omni panel)
> * We can create a story page for the fonts we use. . I'm thinking more like one example and a link to the material-ui  to not reinventing the wheel.
> * Create a story page for the Form components we use:
>   
>   * Input we use. `<TextField>`
>   * The Combobox component  we use.
>   * Button's color

Will start work on the form components once valerie's command-panel is merged. Initially wanted to do it in another pull request but seems like progress there is faster than I expected!

src/stories/BaseComponents/robot-component.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import Robot from '../../components/schedule-visualizer/robot';
+import ColorManager from '../../components/schedule-visualizer/colors';
+
+export interface RobotComponentProps {
+    robot: RomiCore.RobotState;
+    footprint: number;
+    colorManager: ColorManager;
+    renderInfoPanel(): JSX.Element;
+}
```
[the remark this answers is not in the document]
ChawinTan: I don't think extending will work in this situation. If I extend `Romicore.RobotState`, the interface will end up with the properties of `RobotState`, eg, name, model, path etc ... which we are not passing in as props. 

Perhaps I misunderstood your intention. Could you provide me with an example?

src/stories/BaseComponents/button-colors.tsx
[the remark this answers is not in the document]
matiasbavera: Yeah, that's right!

src/stories/BaseComponents/robot-component.tsx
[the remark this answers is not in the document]
matiasbavera: Sure, something like the following 

``` typescript
export interface RobotProps {
  robot: RomiCore.RobotState;
  footprint: number;
  colorManager: ColorManager;
  onClick?(e: React.MouseEvent<SVGGElement>, robot: RomiCore.RobotState): void;
}

interface RobotComponentProps extends RobotProps {
  renderInfoPanel(): React.Element;
}

```

This interface is defined in robot.tsx

ChawinTan:
[Reminder to myself] - work on Form buttons tommorrow

ChawinTan:
@matiasbavera @koonpeng I think these are all the major components and stories that needs to be included for now.

src/stories/form.stories.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
```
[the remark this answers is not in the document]
ChawinTan: @matiasbavera what is your take on this?

ChawinTan:
```
Some components like doors, lifts, robots are clickable, we should look at showing the events in storybook, this will probably need some refactoring.
```
For this, I suppose we should use a mock function to show the events and there is no need to show the contents of the actual functions passed in?

koonpeng:
In a purely presentation component, there would be a generic react onclick event. I believe the current code uses a specialized handler callback, I don't know how action handlers in storybook work and if they can be made to handle arbitrary callbacks. But yeah, the idea is to just show that the component fired the callback.

ChawinTan:
Will continue development after https://github.com/osrf/romi-dashboard/pull/72 has been merged

src/stories/form.stories.tsx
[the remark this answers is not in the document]
matiasbavera: One of the multiple objectives of the Storybook is to use it like "style guide". If someone wants to contribute to the project, it can see how we design our components (in this case which components we use and how `Material-ui `designed its components).

I think it will not hurt to have this here and in case we want to eventually put our own touch of style on `Material-ui `components this is going to be a good place to demonstrate it.

You can check storybook examples here https://storybook.js.org/docs/examples/

If none of the reasons convinces you, you can remove it without hesitation =)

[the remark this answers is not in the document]
ChawinTan: Make sense. I will leave it here for future changes.

ChawinTan:
Did some refactoring of the existing code:
- brought out the states and object representation into `Utils.tsx` so that we can reuse them
- abstract out common styles for components and other design

src/components/lift-item/lifts-panel.tsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 import { LiftItem } from './lift-item';
 import { SpotlightValue } from '../spotlight-value';
 
-interface LiftsPanelProps {
+export interface LiftsPanelProps {
```
ChawinTan, line 6: Just a note to inform that I exported the `LiftPanelProps` since I need it and it was not exported before.

src/stories/design-decision.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+    <div style={styles.example}>
+      <Typography variant="body1">An example:</Typography>
+      <Typography style={styles.panelButton} data-role="state" variant="button">
+        N/A
+      </Typography>
+    </div>
+  </div>
+);
+
+export const DoorButtonColors = () => (
```
koonpeng, line 89: These now render the full info panels instead of just the label, might make more sense to rename it to door panel or something similar and move it to the door section. Same for lifts, robots, dispensers etc.

src/stories/design-decision.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+    <div style={styles.example}>
+      <Typography variant="body1">An example:</Typography>
+      <Typography style={styles.ellipsis} variant="body1">
+        A very very long name
+      </Typography>
+    </div>
+  </div>
+);
+
+export const HandleUnknown = () => (
```
koonpeng, line 69: This still uses hardcoded values, either refactor to have a single source of truth for the design of unknown labels or render a door/lift item with unknown state.

src/stories/design-decision.stories.tsx
```diff
@@ -0,0 +1,127 @@
...
+  panelButton: {
+    border: '2px solid #CCCCCC',
+    padding: 5,
+    width: '4rem',
+    textAlign: 'center',
+    borderRadius: '4px',
+  },
+};
+
+export const HandleLongName = () => (
```
koonpeng, line 50: This still uses hardcoded values, either refactor to have a single source of truth for the design of long labels or render a door/lift item with long names.

src/stories/colorpalette.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+    display: 'flex',
+    justifyContent: 'space-between',
+    margin: '1rem 0',
+  },
+};
+
+const sampleStyles: SampleStyleTyping = {
+  colorSample: {
+    unknown: {
+      backgroundColor: '#cccccc',
```
koonpeng, line 22: Don't hardcode this, if we are standardizing the color palette, we should add a class to reference the colors.

src/stories/BaseComponents/robot-component.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <div style={componentDisplayStyle.display}>
+      {renderInfoPanel()}
+
+      <svg viewBox={viewBoxCoords}>
+        <Robot
+          robot={robot}
+          footprint={footprint}
+          colorManager={colorManager}
+          {...actions('onClick')}
```
koonpeng, line 23: Clicking the robot causes 2 events to fire, can you investigate this?

ChawinTan: There are 2 `onClicks` within `robots.tsx`, one on the `g` element and another on the `circle` element. I removed the one on `circle` as it fails the unit tests and is inconsistent with the other components.

ChawinTan:
@koonpeng @matiasbavera Noticed that there is support for image icons already. Should I address it in this pr? If so, are there some default icons that you guys have already chosen? Don't think I found any of such assets in the project yet.

koonpeng:
let's keep it out for now to prevent scope creep

src/stories/baseComponents/color.tsx
```diff
@@ -0,0 +1,35 @@
...
+export default function Colors() {
+  return (
+    <React.Fragment>
+      {Object.keys(colorPalette).map((color, index) => {
+        return (
+          <React.Fragment>
+            <div style={styles.example} key={index}>
+              <Typography variant="body1">Unknown</Typography>
+              <div>
+                <Typography variant="body1">{`Color code: ${colorPalette[color].backgroundColor}`}</Typography>
```
ChawinTan, line 25: Just to confirm @koonpeng , the hardcoded color that you were referring to is on this line?

src/stories/baseComponents/utils.tsx
```diff
@@ -0,0 +1,294 @@
...
+  aTag: {
+    textDecoration: 'none',
+    color: 'rgb(20, 116, 243)',
+  },
+};
+
+// colors used outside material ui
+export const colorPalette: StyleTyping = {
+  unknown: {
+    backgroundColor: '#cccccc',
```
koonpeng, line 280: We need to have the actual components and the storybook components reference the same value, the actual components are still using hardcoded value, only the `design-decisions` story is using this.

koonpeng (changes requested):

koonpeng (approved):
thanks for the hard work

PR #66, opened 2020-07-17
Added support for png and jpg on robots

matiasbavera:
## What's new

* Added support for png and jpg on robots
* Split robot component into `robot-img-icons` and `robot-default-icons`
* Got the configurations of icons from an auto served JSON file
* Added a pre-build script to get the icons from an external resource.
* New command added to get the resource folder `npm run setup`
* Added a border to robot names.
* Added shadow to robot icon.

image

rebased

image

## Related to
https://github.com/osrf/rmf_demos/pull/109


matiasbavera:
https://developer.mozilla.org/en-US/docs/Web/SVG/Element/image
https://developer.mozilla.org/en-US/docs/Web/SVG/Tutorial/SVG_Image_Tag

matiasbavera:
Hi @koonpeng, can you give me your opinion on the structure presented in this PR to handle the robot icons and other possible icons that may appear in the future?

In simple words:
1) On system start, gets all the icons path from a file. (I'm faking this for now).
2) Sets an IconContext with the icon paths.
3) On the robot component, checks if a path for that model of robot exists.
4) If exists it renders the component with the Icon
5) If not exists or fails, it goes straight to the old robot icon.

koonpeng:
yep, that sounds right. On another note, would it be better to put the context provider in the schedule visualizer? Reason is because we might want to reuse the map on another project.

matiasbavera:
https://stackoverflow.com/questions/7106012/download-a-single-folder-or-directory-from-a-github-repo

matiasbavera:
Used URL https://github.com/matiasbavera/rmf_demos/trunk/feat/createResourceFolder/resources

matiasbavera:
Hi @koonpeng I'm in the process of creating a folder of resources as we discussed in last meeting https://github.com/osrf/rmf_demos/pull/109.
I'm getting that specific folder (resources/) and copying its content to the public folder of Romidashboard, this happens as a result of executing the `npm run setup` command which asks for an URL to get those resources.
We can use this URL for now
https://github.com/matiasbavera/rmf_demos/branches/feat/createResourceFolder/resources

I just want to check something with you, because I think something like this could happen:
Imagine we have the resource folder with:
resource/robots
resource/dispensers
resource/someFolderThatWeFon'tWantToBeOnRomiDashboard

To solve that I think we can have a list of the folders need it in the RomiDashboard and just bring those folders to the `public/` folder. 

What do you think about it?

koonpeng:
That can work, we just need to document down the directories that we expect.

matiasbavera:
Related to https://github.com/osrf/rmf_demos/pull/109

src/components/schedule-visualizer/robot.tsx
```diff
@@ -1,84 +1,87 @@
...
-  React.useLayoutEffect(() => {
-    if (robotColor) {
-      return;
-    }
-    (async () => {
-      setRobotColor(await colorManager.robotColor(robot.name, robot.model));
-    })();
-  }, [robot, robotColor, colorManager]);
+  useEffect(() => {
+    setRenderDefaultIcon(!iconPath || !!iconError);
```
koonpeng, line 47: can this be a state to avoid double rendering?

src/components/schedule-visualizer/robot.tsx
```diff
@@ -1,84 +1,87 @@
...
-      )}
-    </g>
+    <>
+      <g
+        ref={ref}
+        data-component="Robot"
+        aria-label={robot.name}
+        onClick={e => onClick && onClick(e, robot)}
+      >
+        {!!iconPath && !iconError && (
```
koonpeng, line 58: I think this resolves to `!renderDefaultIcon`, using that will make the code more readable.

src/components/schedule-visualizer/robot-image-icon.tsx
```diff
@@ -0,0 +1,63 @@
...
+};
+
+const RobotImageIcon = React.forwardRef(function(
+  props: RobotImageIconProps,
+  ref: React.Ref<SVGGElement>,
+): React.ReactElement {
+  const classes = useStyles();
+  const { robot, footprint, iconPath, dispatchIconError, onClick } = props;
+  // Hardcoded delta of 2 to set the size of the robot. With the double size of the footprint, we achieved a similar size to the robot default svg icon.
+  const [imgIconWidth, imgIconHeigth] = useMemo(() => [footprint * 2, footprint * 2], [footprint]);
```
koonpeng, line 27: The default icon uses footprint as the radius, so we * 2 here because the width/height is in a square, or diameter. Might want to describe this instead of saying it is hardcoded.

src/components/schedule-visualizer/robots-overlay.tsx
```diff
@@ -1,25 +1,41 @@
...
-import React from 'react';
+import React, { useMemo } from 'react';
 import ColorManager from './colors';
 import Robot from './robot';
 import SVGOverlay, { SVGOverlayProps } from './svg-overlay';
 
 export interface RobotsOverlayProps extends SVGOverlayProps {
   robots: readonly RomiCore.RobotState[];
   colorManager: ColorManager;
+  fleets: readonly RomiCore.FleetState[];
```
koonpeng, line 11: Since we are receiving the list of fleets here, we might not need `robots` anymore, it should contain the list of robots already.

src/resource-manager.ts
```diff
@@ -0,0 +1,41 @@
+// We set any here to be more flexible when receiving the configuration structure
+type SubResourceConfigurationsType =
+  | string
+  | Record<string, string | any>
+  | Record<string, Record<string, string | any>>
+  | any;
+export type ResourceConfigurationsType = Record<string, SubResourceConfigurationsType>;
+const axios = require('axios').default;
```
koonpeng, line 8: use the import pattern for typescript

src/util/calculation-helpers.ts
```diff
@@ -0,0 +1,16 @@
+export function radiansToDegrees(radians: number) {
+    var pi = Math.PI;
```
koonpeng, line 2: nit: avoid var, also we could just use `Math.PI` directly since the formula is so simple

scripts/setup/get-icons.js
```diff
@@ -0,0 +1,36 @@
...
+  console.log(chalk`{green The icons have been successfully obtained. Check public/assets/icons/}`);
+};
+
+let rawdata = fs.readFileSync('./.resources.json');
+let data = JSON.parse(rawdata);
+
+if (data.hasOwnProperty('url')) {
+  exec('[ -d "public/assets/icons/" ] && rm -rf public/assets/icons');
+  exec('mkdir -p public/assets/icons');
+  exec(`svn checkout ${data.url} public/assets/icons/`, (error, stdout, stderr) =>
```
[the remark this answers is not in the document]
matiasbavera: In our last meeting, I thought we agreed that the resources should be in a git repo. Each particular project on a different repo. E.g.:
rmf_demos/rmf_dashboard_resources/
rmf_implementationA/rmf_dashboard_resources/

But actually, with the current structure, we can point at any repo folder we want.
Also, I added a functionality for copying the resources from a specific folder inside the server in case necessary

Is it ok? Am I missing something? 

Thinking that I use svn to extract files from a specific folder inside the repo.

. SVN Checkout – Create working copy
Checkout command is used to download sources from SVN repository to working copy. If you want to access files from the SVN server, checkout is the first operation you should perform.

`SVN checkout` creates the working copy, from where you can do edit, delete, or add contents. You can checkout a file, directory, trunk or whole project. To checkout you should know URL of the components you want to checkout.

src: https://www.javatpoint.com/svn-commands and https://www.thegeekstuff.com/2011/04/svn-command-examples/#:~:text=SVN%20stands%20for%20Subversion.,placed%20into%20a%20central%20repository.&text=This%20article%20explains%20some%20basic%20SVN%20commands%20with%20examples.

scripts/setup/setup.js
```diff
@@ -0,0 +1,59 @@
+const chalk = require('chalk');
+const inquirer = require('inquirer');
+var fs = require('fs');
+
+if (fs.existsSync('./.resources.json')) {
+  console.log(chalk`
+  {blue You already configured an icon source!} 👋`);
```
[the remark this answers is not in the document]
matiasbavera: Will do!

src/app-config.ts
```diff
@@ -3,10 +3,13 @@ import { SossTransport } from '@osrf/romi-js-soss-transport';
 import FakeTrajectoryManager from './mock/fake-traj-manager';
 import { FakeTransport } from './mock/fake-transport';
 import { DefaultTrajectoryManager, RobotTrajectoryManager } from './robot-trajectory-manager';
+import fakeAppResources from './mock/data/fake-app-resources';
```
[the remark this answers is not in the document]
matiasbavera: Sure. I'll rename a fleet to `magni` on the mock scenario to show an icon too.

src/components/schedule-visualizer/robot.tsx
matiasbavera: I totally agree with you but I need to check if the `iconPath` exists. If not I got this error. 
```
Type 'string | null' is not assignable to type 'string'.
Type 'null' is not assignable to type 'string'
``` 

Should I use (it's working on both cases):
*  // @ts-ignore on top of `iconPath={iconPath}`
* `!renderDefaultIcon && !!iconPath`. 
* `!!iconPath && !iconError`

src/components/schedule-visualizer/robots-overlay.tsx
matiasbavera: Yeah, we can do that if `RobotsOverlayProps` receives the `currentLevel` too. With the current level, we will be able to get the robots corresponding to that level. Which I think is a good idea, to encapsulate the logic of the robot on the overlay, also we'll be removing some of the robot logic of the ScheduleVisualizer component, which is getting bigger and bigger.

src/resource-manager.ts
matiasbavera: You're right Mr. My bad! Thanks!

src/util/calculation-helpers.ts
matiasbavera: You're right! Thanks!

src/components/schedule-visualizer/robot.tsx
matiasbavera: Makes sense. Done!

src/app-config.ts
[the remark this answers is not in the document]
matiasbavera: Done

src/components/schedule-visualizer/robot-image-icon.tsx
matiasbavera: Makes sense. Thank you.

scripts/setup/get-icons.js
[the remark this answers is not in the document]
koonpeng: I didn't know you can checkout a git repo with svn, but I would still prefer to avoid bringing it in as a dependency.

src/components/schedule-visualizer/robot.tsx
koonpeng: I see, in that case I'm ok with either option, but I don't like adding ts-ignore, you can just leave this as is.

src/components/schedule-visualizer/robots-overlay.tsx
koonpeng: That can work. Currently it only receives a list of robots you that it only receives what it is going to render, receiving all robots would mean we need bring that logic in, there is a bit of refactoring but there shouldn't be any problems with that.

koonpeng:
Can you also add a README.md to describe how to import resource packs?

matiasbavera:
> Can you also add a README.md to describe how to import resource packs?

Yeah! Make sense.

scripts/setup/get-icons.js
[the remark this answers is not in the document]
matiasbavera: Ok! I'll find a way to do it without svn. As a last resort, we could create a repo (let's say `rmf-dashboard-resources`) and add it as a git sub-module to the `rmf-demos` project and make a direct clone in this project (`romi-dashboard`)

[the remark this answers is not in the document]
matiasbavera: Update. I found this information:
* GH does not support git-archive https://stackoverflow.com/questions/7106012/download-a-single-folder-or-directory-from-a-github-repo#:~:text=Step1%3A%20Input%20github%20url%20to,File%22%20button%20to%20get%20files.
* Using the oldest version of git sparse-checkout will still download the whole repository from the server – only the checkout is reduced in size. At the moment it is not possible to clone only a single directory. But if you don't need the history of the repository, you can at least save on bandwidth by creating a shallow clone. 

* The newest (better I found discarding svn option) is to use the new version of sparse-checkout, with no checkout and with no history. Note that it requires git version 2.25 installed. Read more about it here: https://github.blog/2020-01-17-bring-your-monorepo-down-to-size-with-sparse-checkout/

scripts/setup/setup.js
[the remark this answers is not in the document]
matiasbavera: Done!

scripts/setup/get-icons.js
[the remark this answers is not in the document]
matiasbavera: The `git sparse-checkout` was implemented. What do you think about it?

matiasbavera:
I tested with this information

```
{
"repoUrl":"https://github.com/matiasbavera/rmf_demos.git",
"folder":"rmf_dashboard_resources/office/",
"branch":"feat/createResourceFolder"
}
```

scripts/setup/get-icons.js
```diff
@@ -0,0 +1,80 @@
...
+  execSync(
+    `git clone "${resourcesData.repoUrl}" --no-checkout  --depth=1 --single-branch --branch ${resourcesData.branch} ${tempFolder} -o repo`,
+    {
+      stdio: [0, 1, 2],
+      cwd: path.resolve(__dirname, '../../'),
+    },
+  );
+
+  execSync(
+    `git sparse-checkout init --cone &&
```
koonpeng, line 61: `sparse-checkout` is great but its only available in 2.25. ubuntu 20.04 does have it but 18.04 is on 2.17. It is still too new so we should have a fallback for older git versions.

README.md
```diff
@@ -19,6 +19,40 @@ cd ~/.
...
+In the case of choosing to obtain the data from a repository, you must define: (1) The repository URL and (2) the folder within the repository. In case of choosing to obtain the data from a local directory, the absolute path must be defined.
+
+This folder obtained must contain at its root a JSON file calling `main.json`, with the _RomiDashboard_ actors (robots, dispensers, etc.) and the path to their corresponding resources. E.g.:
+
+```json
+{
+  "robots":
+    "magni": {
+      "icons": {
+        "magni": "/robots/magni/magni.png"
```
koonpeng, line 44: can we clarity if the key is the robot model or name?

scripts/setup/get-icons.js
matiasbavera: Thanks for the comments! Oh ok! Well, I refactored a bit the `get-icons` file, so we can add multiple implementations for cloning a specific folder. 
Now we're asking for the git version and for git =>25  we are going to use the new version of `sparse-checkout`. And for  git < 25 we are going to use the old version of `sparse-checkout `(available from git `version 1.17`).

README.md
matiasbavera: Makes sense! Thanks!

scripts/setup/get-icons.js
```diff
@@ -0,0 +1,175 @@
...
+            git checkout`,
+      {
+        stdio: [0, 1, 2],
+        cwd: path.resolve(__dirname, `../../${this.tempFolder}`),
+      },
+    );
+  };
+}
+
+class ISparseCheckoutGitV217 extends IconManagerBase {
```
koonpeng, line 56: Although this is not typescript, but let's still try to follow the style guide https://basarat.gitbook.io/typescript/styleguide#interface

scripts/setup/get-icons.js
```diff
@@ -0,0 +1,175 @@
...
+class IconManagerBase {
+  constructor(resourcesData, iconFolder, tempFolder) {
+    this.resourcesData = resourcesData;
+    this.iconFolder = iconFolder;
+    this.tempFolder = tempFolder;
+    this.rootPath = '../../';
+  }
+}
+
+class ISparseCheckoutGitV225 extends IconManagerBase {
```
koonpeng, line 29: Although this is not typescript, but let's still try to follow the style guide https://basarat.gitbook.io/typescript/styleguide#interface

src/components/app.tsx
```diff
@@ -111,10 +112,11 @@ const viewMap = makeViewMap();
 
 export default function App(props: AppProps): JSX.Element {
   const classes = useStyles();
-  const { transportFactory, trajectoryManagerFactory } = props.appConfig;
+  const { transportFactory, trajectoryManagerFactory, appResources } = props.appConfig;
   const [transport, setTransport] = React.useState<RomiCore.Transport | undefined>(undefined);
   const [buildingMap, setBuildingMap] = React.useState<RomiCore.BuildingMap | undefined>(undefined);
   const trajManager = React.useRef<RobotTrajectoryManager | undefined>(undefined);
+  const ResourceManager = React.useRef<ResourceConfigurationsType | undefined>(undefined);
```
koonpeng, line 119: I don't think this should be PascalCase, `ResourceConfigurationsType` is a `Record<string, SubResourceConfigurationsType>`, usually we should only use `PascalCase` for newable objects or special cases like react functional components.

scripts/setup/get-icons.js
matiasbavera: Got it! Thanks for the feedback!

scripts/setup/get-icons.js
matiasbavera: Got it! Thanks for the feedback!

src/components/app.tsx
matiasbavera: Sorry, you're right, that's a leftover from the refactor. Thanks!

matiasbavera: Just to check, you're talking me about `const ResourceManager` -> `const resourceManager`, right?

Because the type `ResourceConfigurationsType` should be in PascalCase accordingly to https://basarat.gitbook.io/typescript/styleguide#type. Example: https://basarat.gitbook.io/typescript/styleguide#type-vs-interface

matiasbavera:
Comments addressed and branch rebased and added a shadow to the robot icon. 
 rebased

src/components/app.tsx
koonpeng: yep

koonpeng (approved):

PR #71, opened 2020-08-03
Add red shadow to robot icon when there are conflicting trajectories

vallq:
Task: Make robot bigger when it has conflicting trajectories
https://app.asana.com/0/1141736835949017/1183830989063585

photo6143120959767423546

@matiasbavera for your review!


src/components/schedule-visualizer/index.tsx
```diff
@@ -215,23 +220,45 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
...
+          });
+          robotNames && conflictRobotNames.push(robotNames);
+        });
+      }
+      return conflictRobotNames;
+    }
+    if (curMapFloorLayer) {
+      setCurMapTrajectories(getTrajectory(curMapFloorLayer.level.name));
+      setCurMapConflicts(getConflicts(curMapFloorLayer.level.name));
+      setConflictRobotNames(getConflictRobotsName(curMapConflicts, curMapTrajectories));
```
matiasbavera, line 258: I would be extremely careful with this because setState is an async function so maybe `getConflictRobotsName(curMapConflicts, curMapTrajectories) ` can be executed with old values.

src/components/schedule-visualizer/index.tsx
```diff
@@ -215,23 +220,45 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
...
+    }
+  }, [curMapFloorLayer, trajectories, curMapConflicts, curMapTrajectories]);
```
matiasbavera, line 260: Just checking, are we making sure that the components are only rendered when we want? Are we making sure it doesn't go into a state update loop?

koonpeng:
I'm guessing this is to highlight robots currently in conflict? If/When we have the robot dimensions and render each robot with different sizes, this could make things confusing. I think we can use better indicators, adding a warning badge for example.

matiasbavera:
> I'm guessing this is to highlight robots currently in conflict? If/When we have the robot dimensions and render each robot with different sizes, this could make things confusing. I think we can use better indicators, adding a warning badge for example.

Oh I see, that sounds good. What do you think about replacing the current robot’s black shadow for a red one too (In case of conflict of conflict of course)?

vallq:
Hm okay just so I understand the conclusion:
we are looking at using a highlight/shadow to indicate the robot being in conflict instead of a change in size since it will be messy once we have many robots with different dimensions.

> Could you create a function with the common logic?
> Just checking, are we making sure that the components are only rendered when we want? Are we making sure it doesn't go into a state update loop?
@matiasbavera sorry what do you mean by creating function with common logic? and also, I'm not sure how to check and prevent a state update loop.

matiasbavera:
> we are looking at using a highlight/shadow to indicate the robot being in conflict instead of a change in size since it will be messy once we have many robots with different dimensions.

Yeah, I understand that too, but let's check with @koonpeng .



> @matiasbavera sorry what do you mean by creating function with common logic? and also, I'm not sure how to check and prevent a state update loop.

Sorry for not being more clear. Maybe we should discard this logic after the discussion but I'll explain it either way, let me try this way, the following logic is repeated for both test `Robot overlay increases in size when it is in trajectory conflict`  and `Robots Overlay`. I was just saying to create a function to set this to default values, not to copy and paste these values. Something like what you did in the `loopRequest.test.tsx`

```
const robots = fakeFleets()[0].robots;
    const bounds = new L.LatLngBounds([0, 25.7], [-14, 0]);
    const colorManager = new ColorManager();
    // TextEncoder is not available in node
    colorManager.robotColor = jest.fn(async () => 'black');
    colorManager.robotColorFromCache = jest.fn(() => 'black');
    const conflictRobotNames = [fakeFleets()[0].robots[0].name];

    const wrapper = mount(
      <LMap>
        <RobotsOverlay bounds={bounds} colorManager={colorManager} robots={robots} conflictRobotNames={conflictRobotNames} />
      </LMap>,
    );

```

src/components/schedule-visualizer/robot.tsx
```diff
@@ -57,7 +58,12 @@ const Robot = React.forwardRef(function(
             rotate(${-(robot.location.yaw * 180) / Math.PI})`}
           >
             <filter id={`${robot.name}-shadow`} x="-20%" y="-20%" width="140%" height="140%">
-              <feDropShadow dx="0" dy="0" stdDeviation={footprint * 0.15} floodColor="black" />
+              <feDropShadow
+                dx="0"
+                dy="0"
+                stdDeviation={footprint * 0.15}
+                floodColor={inConflict ? 'red' : 'black'}
```
matiasbavera, line 65: Can you use `theme.palette.error.main` (same color of the trajectory) instead of 'red'?

src/components/schedule-visualizer/tests/robots-overlay.test.tsx
```diff
@@ -7,20 +7,46 @@ import ColorManager from '../colors';
...
+      <LMap>
+        <RobotsOverlay
+          bounds={bounds}
+          colorManager={colorManager}
+          robots={robots}
+          conflictRobotNames={conflictRobotNames}
+        />
+      </LMap>,
+    );
+    expect(wrapper.containsMatchingElement(<feDropShadow floodColor={'red'} />)).toBeTruthy();
```
matiasbavera, line 49: If you use theme.palette.error.main maybe you would have a problem here?

src/components/schedule-visualizer/index.tsx
matiasbavera: Can you do something like this?

``` typescript
 const mapTrajectories = getTrajectory(curMapFloorLayer.level.name);
 const mapConflicts = getConflicts(curMapFloorLayer.level.name);
 setCurMapTrajectories(mapTrajectories);
 setCurMapConflicts(mapConflicts);
 setConflictRobotNames(getConflictRobotsName(mapConflicts, mapTrajectories));
```

To make sure we are passing the updated values to `setConflictRobotNames`?

src/components/schedule-visualizer/index.tsx
```diff
@@ -215,23 +220,45 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
...
+          let robotNames: string[] = [];
+          conflictPair.forEach(conflictId => {
+            const robotName = DefaultTrajectoryManager.getRobotNameFromPathId(conflictId, trajs);
+            robotName && robotNames.push(robotName);
+          });
+          robotNames && conflictRobotNames.push(robotNames);
+        });
+      }
+      return conflictRobotNames;
+    }
```
matiasbavera, line 254: What do you think if we do this? It looks less convoluted, right?

``` typescript
      if (conflicts.length === 0) {
        return conflictRobotNames;
      }

      conflicts.forEach(conflictPair => {
        let robotNames: string[] = [];
        conflictPair.forEach(conflictId => {
          const robotName = DefaultTrajectoryManager.getRobotNameFromPathId(conflictId, trajs);
          robotName && robotNames.push(robotName);
        });
        robotNames && conflictRobotNames.push(robotNames);
      });
```

matiasbavera:
> Task: Make robot bigger when it has conflicting trajectories
> https://app.asana.com/0/1141736835949017/1183830989063585
> 
> photo6143120959767423546
> 
> @matiasbavera for your review!

Maybe you should update the PR description.

src/components/schedule-visualizer/index.tsx
vallq:
```
      if (conflicts.length !== 0) {
        return conflictRobotNames;
      }

      conflicts.forEach(conflictPair => {
        let robotNames: string[] = [];
        conflictPair.forEach(conflictId => {
          const robotName = DefaultTrajectoryManager.getRobotNameFromPathId(conflictId, trajs);
          robotName && robotNames.push(robotName);
        });
        robotNames && conflictRobotNames.push(robotNames);
      });
      return conflictRobotNames;
```

if the repeating return line is alright, then yea we can go with separating them!

src/components/schedule-visualizer/tests/robots-overlay.test.tsx
vallq:
```
expect(wrapper.containsMatchingElement(<feDropShadow floodColor="#f44336" />)).toBeTruthy();
```
this is the simplest workaround that I can think of: to take the actual value of the red used in the theme object. Would it be bad practice?

src/components/schedule-visualizer/index.tsx
matiasbavera: Yeah, we can return an empty array of type string[][] to avoid confusion on the first `conflictRobotNames`.

src/components/schedule-visualizer/tests/robots-overlay.test.tsx
matiasbavera: Mmm.. how about creating a new theme since we are using the default 
`const theme = createMuiTheme();`
With that, you will have access to `theme.palette.error.main`.
`floodColor={theme.palette.error.main}`
If in the future we want to create a custom theme, we would also be able to access that configuration in this test (if it's exported, of course).  

HTH,

src/components/schedule-visualizer/tests/robots-overlay.test.tsx
```diff
@@ -6,21 +6,59 @@ import RobotsOverlay from '../robots-overlay';
...
 
-  wrapper.unmount();
+describe('Robots Overlay', () => {
+  const theme = createMuiTheme({
+    palette: {
+      error: {
+        main: '#f44336',
+      },
+    },
+  });
```
matiasbavera, line 26: I think we can use just `const theme = createMuiTheme();` since we are using the default theme.

vallq: oh okay, i thought we would need to specify props for a new theme!

vallq:
thanks for all the thorough checking and comments @matiasbavera!
@koonpeng ready for merging

PR #72, opened 2020-08-03
Refactor Accordion labels into one component

ChawinTan:
Refactoring accordion labels into a separate component. 

https://github.com/osrf/romi-dashboard/pull/65#discussion_r461124966

Asana task link:
https://app.asana.com/0/1141736835949017/1186395258093417/f


src/components/labels.tsx
```diff
@@ -0,0 +1,35 @@
...
+          {liftState ? liftState.current_floor : 'N/A'}
+        </Typography>
+        :
+        <Typography className={modalLabelClass} variant='button'>
+          {modeText}
+      </Typography>
+      }
+    </React.Fragment>
+  )
+}
```
matiasbavera, line 35: Components like these ([presentational](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0)) should ideally be independent of the logic of other components. In other words, the component should be responsible for rendering a label with plain data passed to it and not knowing if there is an elevator status. So I would recommend that the elevator status be handled in the parent component. I mean sending 
`{liftState? liftState.current_floor: 'N / A'} ` in the `modeText` parameter.

src/components/labels.tsx
```diff
@@ -0,0 +1,35 @@
...
+interface LabelProps {
+  hideText: string;
+  modalLabelClass: string;
+  modeText?: string;
+  liftState?: Readonly<RomiCore.LiftState>;
+  liftName?: string;
+  name?: string;
+}
+
+export default function Labels(props: LabelProps) {
```
matiasbavera, line 14: `Labels` feel like a too generic name. Can we be more descriptive with that name it something like OmnipanelStateLabel?

src/components/labels.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import { Typography } from '@material-ui/core';
+import * as RomiCore from '@osrf/romi-js-core-interfaces';
+
+interface LabelProps {
+  hideText: string;
```
matiasbavera, line 6: When I read `hideText` feels like it should be receiving a boolean. Can we renamed to something like hideTextStyle or hideTextClass? Or can we add the following 

```
    hideText: {
      overflow: "hidden",
      textOverflow: "ellipsis",
      whiteSpace: "nowrap",
      maxWidth: "10rem",
    }
```

to the component default style with the option to overwrite it?

koonpeng: I agree, `hideText` is passed in from various components all using the same values. There are some ways we can improve on it

1. as suggested, have a default style and have a `className` (or similar) prop to overwrite/append it.
2. in the case where we want to have multiple pre-defined styles, we can have a `variant` prop similar to how `Typography` does it.
3. If we want to use custom styling for a particular label, we can export the styles in the module so we can render it in storybook without resorting to hardcoded values.

src/components/labels.tsx
koonpeng: Yea, the label shouldn't need information on the lift/door/robot state, it should just receive a string for the text. I don't think we need 3 `Typography` components that render conditionally either.

src/components/omni-panel-status-labels.tsx
```diff
@@ -0,0 +1,34 @@
...
+  }
+}));
+
+export default function OmniPanelStatusLabels(props: LabelProps) {
+  const classes = useStyles();
+  const { modalLabelClass, modeText, name, hideTextStyleOverride } = props;
+
+  return (
+    <React.Fragment>
+      <Typography variant="h6" className={hideTextStyleOverride ? hideTextStyleOverride : classes.hideText}>
```
ChawinTan, line 26: Just curious, is this the right way to do style overriding? Seems like I am violating one of the rules of presentational components again

matiasbavera: Simple answer: yes.

[Not useful for now] Extra information for the future
It's likely that in the future we'll want to have some components with a part of the style immutable and that only certain things such as colors, etc. can be overwritten. In that case, we shouldn't replace the entire style as we are doing now and we should probably also check the entered text type with custom Typescript types.

ChawinTan:
I'm happy with writing snapshot tests! Initially, I saw that there were only a few components that were tested in the `tests` folder so I though we don't do unit tests on presentational components. Maybe I should have asked earlier, but are there any rules for components that we need to or do not need to test, or is that still a work in progress and we should be writing unit test for every single component from now on?

koonpeng:
In general, its always good to have tests whenever possible, any lack of tests is usually a result of the initial development when the work flow has not settled down yet.

src/components/tests/__snapshots__/omni-panel-status-labels.test.tsx.snap
```diff
@@ -0,0 +1,216 @@
...
+          "noWrap": "MuiTypography-noWrap",
+          "overline": "MuiTypography-overline",
+          "paragraph": "MuiTypography-paragraph",
+          "root": "MuiTypography-root",
+          "srOnly": "MuiTypography-srOnly",
+          "subtitle1": "MuiTypography-subtitle1",
+          "subtitle2": "MuiTypography-subtitle2",
+        }
+      }
+      variant="h6"
```
matiasbavera, line 49: I got to say that this was weird for me, but actually happened the same when I executed this

```
test('test', () => {
  expect(toJson(mount(<Typography variant="h4"> test </Typography>))).toMatchSnapshot();
});
```

src/components/tests/omni-panel-status-labels.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+  Component: (props: LabelProps) => React.ReactElement,
+  modeText: string,
+  modalLabelClass: string,
+  name: string,
+  hideTextStyleOverride?: string,
+) => {
+  return mount(
+    getStatusLabelsComponent(Component, modeText, modalLabelClass, name, hideTextStyleOverride),
+  );
+};
```
matiasbavera, line 36: I think you are sending the same parameters without any modification to both of the functions. Can we simplify the logic?. Something like this

```
const buildLabelsWrapper = (
  modeText: string,
  modalLabelClass: string,
  name: string,
  hideTextStyleOverride?: string,
) => {
  return mount(
// I think it is ok to put here the component because we are just testing one component in this file
    <OmniPanelStatusLabels
      modeText={modeText}
      modalLabelClass={modalLabelClass}
      name={name}
      hideTextStyleOverride={hideTextStyleOverride}
    />,
  );
}; 
```

src/components/tests/omni-panel-status-labels.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+  expect(
+    toJson(
+      buildLabelsWrapper(OmniPanelStatusLabels, 'State', 'mock-style', 'Robot', 'override-style'),
+    ),
+  ).toMatchSnapshot();
+});
+
+test('Component default style is being override when alternative style is provided', () => {
+  const overrideStyle: string = 'override-style';
```
matiasbavera, line 51: nit: I think you can send 'override-style' directly as a parameter, Looks descriptive enough to me.

ChawinTan: I get what you mean, but I think it is better to assign it to a variable so as to avoid hardcoding values in the following segments should we want to change the value for some reason in the future.

```  
const wrapper = buildLabelsWrapper(
    OmniPanelStatusLabels,
    'State',
    'mock-style',
    'Robot',
    overrideStyle,  <----- here
  );

  const isOverride: boolean = wrapper
    .find(OmniPanelStatusLabels)
    .at(0)
    .find('h6')
    .hasClass(overrideStyle);   <---- and here
```

src/components/tests/__snapshots__/omni-panel-status-labels.test.tsx.snap
ChawinTan: Hmmmm .... I suppose you mean that when you run the following test with the `h4` variant, the new test results are not reflected in the `.snap` file?

For me, when I ran this test, the results are appended at the end of the file, after the previous snapshot test.

src/components/tests/omni-panel-status-labels.test.tsx
matiasbavera: whoops, I didn't see that, you're right =)

src/components/tests/__snapshots__/omni-panel-status-labels.test.tsx.snap
matiasbavera: Sorry this was just a comment, it was weird to me to see that the component to prints its whole API/ options

```
classes={
        Object {
          "alignCenter": "MuiTypography-alignCenter",
          "alignJustify": "MuiTypography-alignJustify",
          "alignLeft": "MuiTypography-alignLeft",
          "alignRight": "MuiTypography-alignRight",
          "body1": "MuiTypography-body1",
          "body2": "MuiTypography-body2",
          "button": "MuiTypography-button",
          "caption": "MuiTypography-caption",
          "colorError": "MuiTypography-colorError",
          "colorInherit": "MuiTypography-colorInherit",
          "colorPrimary": "MuiTypography-colorPrimary",
          "colorSecondary": "MuiTypography-colorSecondary",
          "colorTextPrimary": "MuiTypography-colorTextPrimary",
          "colorTextSecondary": "MuiTypography-colorTextSecondary",
          "displayBlock": "MuiTypography-displayBlock",
          "displayInline": "MuiTypography-displayInline",
          "gutterBottom": "MuiTypography-gutterBottom",
          "h1": "MuiTypography-h1",
          "h2": "MuiTypography-h2",
          "h3": "MuiTypography-h3",
          "h4": "MuiTypography-h4",
          "h5": "MuiTypography-h5",
          "h6": "MuiTypography-h6",
          "noWrap": "MuiTypography-noWrap",
          "overline": "MuiTypography-overline",
          "paragraph": "MuiTypography-paragraph",
          "root": "MuiTypography-root",
          "srOnly": "MuiTypography-srOnly",
          "subtitle1": "MuiTypography-subtitle1",
          "subtitle2": "MuiTypography-subtitle2",
        }
```

PR #74, opened 2020-08-05
Simplified way to install the `rmf_demos` as a backend

matiasbavera:
Hey @koonpeng 

What do you think about having this for the Romi-dashboard too?

## What's new
Simplified the way to install the `rmf_demos` project to be used as a backend for this project. 
* RMF dependencies are added as git submodules
* npm install will update and clone (if necessary) the submodules
* npm run `install:backend` now runs colcon to build RMF dependencies

ROS still needs to be sourced before building.


koonpeng:
I'm not really sure if we should have this here, romi-dashboard doesn't really depend on rmf_demos and other rmf repos.

matiasbavera:
Oh, I get it, maybe it's better to have Romi-Dashboard in rmf_demo project right?
Closing this

PR #78, opened 2020-08-12
Feature/robot trajectory differentiation

ChawinTan:
All my changes here are temporary and is just for demoing purposes. I also added some packages to help with the creation of patterns and quick prototyping. 

@koonpeng @matiasbavera would need your opinion on this one. I currently hardcoded 3 different patterns on the path of the test robots, RobotA, RobotB and RobotC and found some issues with this approach:

1) The patterns are being overlapped and it looks very messy. 
2) Currently, the paths are narrow so it is hard to see the patterns.
3) Because of the way `viewbox` works, it is quite hard to scale the patterns. I think it might also be that I am using a package and there are constants in the packages which cause the scaling issues.

I think the existing implementation that @matiasbavera  had done by differentiating the trajectories through robot colors is actually good enough to tackle this issue. What are your opinions?

I think another improvement that can be made to the trajectories is when the trajectories conflict. Right now, they are all the same color, `theme.palette.error.main`. I am thinking of highlighting them by implementing a pulsating shadow around the conflicted trajectories. What's your take on this? Is this necessary with valerie working on shadow around the conflicting robots?


matiasbavera:
Adding a screenshot 
image

matiasbavera:
> I think the existing implementation that @matiasbavera had done by differentiating the trajectories through robot colors is actually good enough to tackle this issue. What are your opinions?

I think this is a good solution for differentiating paths. I'm also thinking (and this is just an idea) that we can add some geometric figures inside the trajectory
image
Mmm.. but this will also be a kind of messy on the interceptions.

 

> I think another improvement that can be made to the trajectories is when the trajectories conflict. Right now, they are all the same color, `theme.palette.error.main`. I am thinking of highlighting them by implementing a pulsating shadow around the conflicted trajectories. What's your take on this? Is this necessary with valerie working on shadow around the conflicting robots?

Yeah, I think that shadow on the robot complements the warning structure but is not as visible as a red path. Maybe we need to try if that red shadow actually catches your eye (in a good way) on the screen when you see a conflict (instead of the red path ).

koonpeng:
I think the current "Follow" animation actually does a pretty good job of differentiating the trajectories. But let's see if adding patterns will improve it without making it look too cluttered and confusing.

There's some problem with identifying the trajectories by colors that stops me from merging the current implementation:
* To me it looks confusing for each trajectory to have different colors, and it is hard to relate which color belongs to which robot.
* There's no logic to avoid unintuitive colors like black, grey, white, red etc.

ChawinTan:
My thoughts too. I think adding too much details in the path can make it look unnecessarily cluttered.

@matiasbavera I used the squares and circles shape for the paths of RobotB and RobotC actually, but I guess it wasn't too obvious due to the configurations. Added a fill and reduced the pattern size to make them standout and swap robotA with robotC.

ChawinTan:
Added some options on the settings-drawer so that it is easier for you guys to toggle between different configurations to see which is better.

Trajectory size:
- default (0.4)
- robot (size of robot, currenlty hardcoded at 0.5)

Trajectory color:
- plain (current implementation)
- robot (@matiasbavera  implementation of robot colors)
- shapes (with patterns)

matiasbavera:
Thanks for this @ChawinTan . Yeah, as you guys said adding too much detail in the path can make it look unnecessarily cluttered. I agreed with @koonpeng on the `follow` animation. I have a new proposal, let me know what you think:

 Use the `follow` animation by default. For the trajectory, work with a single color (such as green) and use its different shades (light green to dark green). For example, we could always start assigning dark green and assign a lighter hue to each new active trajectory. E.g:
1. Trajectory 1 (dark green 100%)
2. Trajectory 2 (dark green 80%)
..
20. Trajectory 20 (light green 30%)

As the trajectories are completed or finished, we can reuse the free colors.

I was also going to propose that the conflicting trajectories be a little larger than the normal trajectories so that they gain more visibility in the interface. But as we already discussed with @koonpeng , if we have different sizes of robots and the trajectory is also going to be reflected by the size of the robot, increasing the size of the trajectory will tend to confuse the end-user.

ChawinTan:
I can write an implementation to test out using the same color for paths but would need to find some way to avoid colors like white/black.

I think making conflicting trajectories larger can work if we have an option for the path size to be a constant value (0.4 at the moment). If users chose the option for path size to be the same size as the robot, then we can disable this feature.

ChawinTan:
I made a simpler implementation of the above proposal. I calculated the different shades of green and cache them for use later, much like the current implementation of Robot colors.

The range of the hue currently ranges from 90-150 and the luminance between 20-80 while keeping the saturation at 100% (full green) so that it does not veer too close to black, white or grey colors. Looking at the current test trajectories alone, I think it can be a viable alternative if all we want to do is just to differentiate the trajectories.

matiasbavera:
> I made a simpler implementation of the above proposal. I calculated the different shades of green and cache them for use later, much like the current implementation of Robot colors.
> 
> The range of the hue currently ranges from 90-150 and the luminance between 20-80 while keeping the saturation at 100% (full green) so that it does not veer too close to black, white or grey colors. Looking at the current test trajectories alone, I think it can be a viable alternative if all we want to do is just to differentiate the trajectories.

Thanks for this @ChawinTan . Looks like a viable alternative for me too. The paths can be easily differentiated and at least for me, it is not confusing. Let's check with @koonpeng too.

koonpeng:
This adds a few extra settings, let's discuss the direction we want to go in regards to the design plan. Should we go the apple way (minimal design/features), or the kde way (slap a setting for everything). Personally I prefer just having settings that can arguably provide some benefits.

I think at some point we should vote on the default implementation and which alternatives to include or reject. Having different trajectory scenarios in a storybook would be very useful for this.

ChawinTan:
Hmmm ... ok, here is what I think.

With regards to the size of the trajectory, I think it would be good to include another option for a fix trajectory size just in case the robot sizes are too big or small as the path size might look awkward on the map.

For the design of the trajectory, I think I would prefer the apple way. When there are many robots moving around, too many different colors and patterns might not be as useful as we think since there is just too many paths to track.

ChawinTan:
Here's a poll I created ... and also to test if it works :smiley: 

Poll to include options for trajectory size

[](https://api.gh-polls.com/poll/01EFXQ1XXGBMTS810W5XSGAVZ8/Current%20implementation%20based%20on%20Robot%20Size%20will%20do/vote)
[](https://api.gh-polls.com/poll/01EFXQ1XXGBMTS810W5XSGAVZ8/Should%20include%20another%20option%20for%20a%20fix%20sized%20path/vote)

Poll for trajectory color and design/patterns

[](https://api.gh-polls.com/poll/01EFXQAQWTC41947JGHMF3PB2Y/Different%20shades%20of%20one%20Green/vote)
[](https://api.gh-polls.com/poll/01EFXQAQWTC41947JGHMF3PB2Y/Current%20implementation%20of%20just%20one%20color%20for%20un-conflicting%20paths/vote)
[](https://api.gh-polls.com/poll/01EFXQAQWTC41947JGHMF3PB2Y/Path%20color%20will%20be%20the%20same%20as%20Robot%20color/vote)
[](https://api.gh-polls.com/poll/01EFXQAQWTC41947JGHMF3PB2Y/Include%20shapes%20and%20patterns%20in%20the%20path/vote)

matiasbavera:
> Personally I prefer just having settings that can arguably provide some benefits.

I agree with you.


> With regards to the size of the trajectory, I think it would be good to include another option for a fix trajectory size just in case the robot sizes are too big or small as the path size might look awkward on the map.

Sounds good, in case we implement that now, we can also evaluate this approach later with the final user (for keeping it or removing it). (Just a thought) Maybe the fixed size can be useful in situations where aren't many robots and we don't need that extra differentiating pattern (like size). 

> For the design of the trajectory, I think I would prefer the apple way. When there are many robots moving around, too many different colors and patterns might not be as useful as we think since there is just too many paths to track.

Agreed

ChawinTan:
Looks like most people have already voted.

Based on the results, these are the additional options that I will add:

1) Include an additional option to have a fix trajectory size. We will showcase this feature to the end user once the app is ready to see if they think this is necessary or not.

For the trajectory design, 2 options currently stand out with the default implementation a close third, so I will include the following 3 options:

1) Current implementation
2) Different shades of green (or same color, other than grey, white, black or red)
3) Robot trajectory to be the same as robot color

Finally, include the different options for trajectory representation in story book.

Is everyone ok with the following arrangement?

ChawinTan:
One more thing ....

Which of the options would you guys prefer to have as default?

Trajectory size:

[](https://api.gh-polls.com/poll/01EFZKJHS9Q5Q5PJ2EW5F71CKF/Robot%20size/vote)
[](https://api.gh-polls.com/poll/01EFZKJHS9Q5Q5PJ2EW5F71CKF/fix%20size/vote)

Trajectory design:

[](https://api.gh-polls.com/poll/01EFZKNDXC5XJZ8ZCQE8MVA1NC/Current%20implementation/vote)
[](https://api.gh-polls.com/poll/01EFZKNDXC5XJZ8ZCQE8MVA1NC/Different%20shades%20of%20green/vote)
[](https://api.gh-polls.com/poll/01EFZKNDXC5XJZ8ZCQE8MVA1NC/Robot%20colors/vote)

ChawinTan:
As requested by Matthew on Slack:

Current implementation:
current_trajectory

Different shades of green:
different_shades_of_green

With Robot colors:
robot_colors

With Patterns:
with_patterns

koonpeng:
The general consensus is that we go with robot color as default. I think that can look nice but the current implementation has some problems:

 * conflicting trajectories are fixed as red, we should look at other ways to show it. Preferably some ways that allows us to still differentiate conflicting trajectories.
* The path colors are based on the default robot icon colors, in the case where we have a robot icon the colors would not match. See if we can analyze the icon to select a color that matches it, e.g. magni would always be blue-ish, mir would be white-ish. White is a very common color though so this might not work out very well, @matiasbavera @ChawinTan what do you think about it?

ChawinTan:
> conflicting trajectories are fixed as red, we should look at other ways to show it. Preferably some ways that allows us to still differentiate conflicting trajectories.

One way I can think of is to make the conflicting trajectories blink or flash, kind of like those warning signals.

> The path colors are based on the default robot icon colors, in the case where we have a robot icon the colors would not match. See if we can analyze the icon to select a color that matches it, e.g. magni would always be blue-ish, mir would be white-ish. White is a very common color though so this might not work out very well, @matiasbavera @ChawinTan what do you think about it?

Agree .... this is going to be big issue. I never thought about robots being white/black or red as I did not take too much notice on @matiasbavera pr on supporting robot image icons. One way I can think of is to disable this option when there are image icons, fully or partially and allow only when all robots are using default icons. 

Or maybe we can simply remove that option and that would be easier. :sweat_smile:

ChawinTan:
Added an animation to make the path blink (fade in and out). It is currently only visible under the choice `Robot_Color`. I set the [calcMode](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/calcMode) to `paced` which gives it the fading effect, although `linear` looks no different visually. The color is still left as red for now since everything is not confirmed.

What do you guys think?

matiasbavera:
> * conflicting trajectories are fixed as red, we should look at other ways to show it. Preferably some ways that allows us to still differentiate conflicting trajectories.

As you mentioned before I think no robot should be red, black or white because it'll be counter-intuitive. Once said that, I don't see a problem for the conflicting trajectories being fixed as red. I think the red color will help us to catch the end-user attention.


> Added an animation to make the path blink (fade in and out). It is currently only visible under the choice `Robot_Color`. I set the [calcMode](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/calcMode) to `paced` which gives it the fading effect, although `linear` looks no different visually. The color is still left as red for now since everything is not confirmed.
> 
> What do you guys think?
Hey, actually, it looks good, I thought it was going to come out of the style we used, but it looks pretty natural to me. 

> * The path colors are based on the default robot icon colors, in the case where we have a robot icon the colors would not match. See if we can analyze the icon to select a color that matches it, e.g. magni would always be blue-ish, mir would be white-ish. White is a very common color though so this might not work out very well, @matiasbavera @ChawinTan what do you think about it?

Yeah, that sounds good. If the processing that we must do to obtain the color of the icon is too much to let the client handle that, we could also send a default color for the icon on `.resources.json`

mxgrey:
What if conflicting trajectories get a red border around the path instead of completely turning red? As long as no robots/paths are being drawn with a red-ish shade, that should make it pretty clear which are in conflict.

matiasbavera:
> What if conflicting trajectories get a red border around the path instead of completely turning red? As long as no robots/paths are being drawn with a red-ish shade, that should make it pretty clear which are in conflict.

Yeah! That could work too, what do you think @ChawinTan ?

ChawinTan:
> What if conflicting trajectories get a red border around the path instead of completely turning red? As long as no robots/paths are being drawn with a red-ish shade, that should make it pretty clear which are in conflict.

Yup, I think that could work, although the borders need to be thick else if you encounter colors like the dark blue of the existing magni robot icons, it might be not be easily distinguishable.

ChawinTan:
> Yeah, that sounds good. If the processing that we must do to obtain the color of the icon is too much to let the client handle that, we could also send a default color for the icon on .resources.json

I actually think this might be a better idea. The current way of doing it using a canvas is quite hacky and requires a more robust implementation for it to work consistently. I will see if I can make enough improvements on the current implementation first though.

koonpeng:
We can process the colors at startup and save them to persistent storage since they are not expected to change often. Or better yet we can process it when we import the resource pack and save it in a json which the client read at runtime.

This library might help https://jariz.github.io/vibrant.js/.

ChawinTan:
shadow_outline_conflict

Here's how adding a red border looks like. I used a shadow in this case.

So currently, we have proposed adding a flashing animation and also highlighting the border of the path in red. @koonpeng @matiasbavera what is your opinion of each approach? Which one would you prefer or you have other suggestions in mind?

ChawinTan:
> We can process the colors at startup and save them to persistent storage since they are not expected to change often. Or better yet we can process it when we import the resource pack and save it in a json which the client read at runtime.
> 
> This library might help https://jariz.github.io/vibrant.js/.

Thanks! will go try it out later

mxgrey:
> Here's how adding a red border looks like. I used a shadow in this case.

Would it be possible to make the outline thicker and/or brighter?

And bonus points if it can flash like an alarm, but maybe that's going overboard.

ChawinTan:
> Would it be possible to make the outline thicker and/or brighter?
> 
> And bonus points if it can flash like an alarm, but maybe that's going overboard.

Yup, but right now the outline is causing some trajectories to not display correctly (eg: trajectory becomes thin and breaking up in some cases) so that will be my priority for now

ChawinTan:
`Something I noticed and I think we need to be careful, is with the number of renders we need to show the animations. With these changes, at least in my computer, the normal trajectories run smoothly as always, but the conflicting trajectories animations are too slow.`

I noticed this too but it was also before adding all the extra animations. The animations tend to be jerky around the turning points. Will take a look at that and see what I can do

koonpeng:
Sometimes the animation looks jerky because the robot itself is jerky. We are rendering a prediction of the robot's path, sometimes it predicts that it will slow/stop for a while due to turning etc which results in jerky animations. That being said, the lag shown here is not normal and should be looked at.

ChawinTan:
> Would Looks good, Can we also try it with a stronger red?.

What do you think of `secondary.main`?

So far, the animations have been running quite smoothly on my computer with no visible lag but it may be that my computer is new or it is not running too much processe. I think the lag might be caused by the `feDropShadow` as it is several filters merged together and is computationally intensive. I tried out [this](https://codepen.io/jonkwheeler/pen/mpJVWr/) method but it did not work too well with our implementation due to the `opacity` being less than 1.

Anyone has more ideas on how to approach this issue? I was thinking of using the `outline` animation logic for a red outline instead.

matiasbavera:
Now that we change the name from `magni` to `tinyRobot` we are making the robot color pink. This pink does not differ well from our shadow red. We have two options here, one is to make the robot not be able to take this pink color (and place a default one), or to darken the colors (of the shadow or of the trajectory).

Maybe with the option that @ChawinTan proposed, to use the border of the `outline` animation, can work in this case too.

image

koonpeng:
@matiasbavera Currently the mock trajectories are obtained from a live run so it would contain the stalls that happens. But since the mock are looped, the micro stalls should always happen at the same spot everytime.

---

@ChawinTan I have just merged in the storybook PR, I suggest that we create a bunch of trajectories in storybook to test the visual and performance. You might want to check with @Yadunund on how to create the trajectories, it involves some maths so to make it easier in the future we can have a script to help build a trajectory. The story can contain more complex scenarios to help visualize how different animations look. We can have a story for benchmark also, just spam hundreds of trajectories and measure the fps.

ChawinTan:
Found a work around that seems to improve the performance. I created a separate path without the `ref` and applied the shadow on the path. @matiasbavera can you try it on the machine to see if it improves while I go learn up on bezier curves and implement the story book?

Also restricted the range of the hue to 50-270 to prevent any colors that is too similar to red. Now, the colors range from yellowish orange to purple.

matiasbavera:
Hi @ChawinTan 

> Also restricted the range of the hue to 50-270 to prevent any colors that is too similar to red. Now, the colors range from yellowish orange to purple.

Cool! Looks better now.

> Found a work around that seems to improve the performance. I created a separate path without the `ref` and applied the shadow on the path. @matiasbavera can you try it on the machine to see if it improves while I go learn up on bezier curves and implement the story book?

Thanks @ChawinTan, It's way better now. There are small white lines between the trajectory and the shadow.
withoutpink


> @ChawinTan I have just merged in the storybook PR, I suggest that we create a bunch of trajectories in storybook to test the visual and performance. You might want to check with @Yadunund on how to create the trajectories, it involves some maths so to make it easier in the future we can have a script to help build a trajectory. The story can contain more complex scenarios to help visualize how different animations look. We can have a story for benchmark also, just spam hundreds of trajectories and measure the fps.

Good idea. +1

ChawinTan:
Still working on the script to generate the waypoints of trajectories.

ChawinTan:
storybook_conflict_Traj

Now that the infrastructure for creating random trajectories flexibly is more or less done, I am going to start on the test cases. Here is the list of cases that has been mentioned and that I can think of thus far:

1) Spam trajectories and benchmark the performance (both conflict and non-conflicting) by measuring fps.
2) Create visual test case for each type of animation, (conflicting and non-conflicting). We can do this by rendering a good amount of trajectories (10-20) and see how they look like when they overlap each other.
3) Create visual test case for each type of animation when all are conflicting at once.
4) Create visual test case for each type of animation when there are partial conflicts
5) (TBC) Test out other ways to represent conflicting trajectories. (Have we confirmed on the design to represent conflicting trajectories?)
6) Different color schemes?

I am currently creating trajectories by generating the waypoints during runtime and as such, the trajectories are different on every reload. Do you guys think this is a good idea or should we use the script as a standalone (maybe commit it to a separate github repo) and use it to generate a bunch of trajectories on it's own, then copy the output of the script into `romi--dashboard`?

koonpeng:
That looks awesome! imo, we should have some fixed trajectories in the storybook, as a bonus we can also have a dynamic runtime generated trajectories as a story. The script can live in this repo if you are happy with it.

mxgrey:
When writing tests, there are a few cases you might want to account for that our current work hasn't really covered yet. These are cases that are supported by the scheduling system but just haven't been used yet:

* Curving trajectories
* Simultaneous translation + rotation
* Third-order trajectories (i.e. trajectories that experience jerk rather than simply acceleration)

ChawinTan:
Video quality is bad cause I used an online converter

There are a total of 150 trajectories for both cases. When all 150 are conflicted, the animation is definitely a lot slower.

spam_conflict
spam-1

Here is the link to the script:
https://github.com/ChawinTan/RomiH-Trajectory-Generator

koonpeng:
And they say tokyo metro is complex 😆

matiasbavera:
> Video quality is bad cause I used an online converter
> 
> There are a total of 150 trajectories for both cases. When all 150 are conflicted, the animation is definitely a lot slower.
> 
> 
> Here is the link to the script:
> https://github.com/ChawinTan/RomiH-Trajectory-Generator

This is awesome. Thanks @ChawinTan, this is really helpful, it raises a lot of things to think about it.

image

If something like this can happen in real life, we will necessarily have to take action. Like for example putting a border (like the outline animation, but it could be something fixed, since we have an animation that indicates the route and direction, this will help to differentiate each path a lot). I'm getting the idea from this which is much easier to understand
image

Adding a simple shadow has a huge performance impact when many paths are added. Should we continue down the path of adding shadows?
1. I think we can see if performance improves with the library Val was testing.
2. If it is feasible to have 100+ robots in conflict at the same time in real life, we should think about having a simpler animation to optimize performance.

koonpeng:
https://www.chromium.org/developers/design-documents/image-filters this might be interesting, maybe the filters are not using gpu.

koonpeng:
> > Video quality is bad cause I used an online converter
> > There are a total of 150 trajectories for both cases. When all 150 are conflicted, the animation is definitely a lot slower.
> > Here is the link to the script:
> > https://github.com/ChawinTan/RomiH-Trajectory-Generator
> 
> This is awesome. Thanks @ChawinTan, this is really helpful, it raises a lot of things to think about it.
> 
> image
> 
> If something like this can happen in real life, we will necessarily have to take action. Like for example putting a border (like the outline animation, but it could be something fixed, since we have an animation that indicates the route and direction, this will help to differentiate each path a lot). I'm getting the idea from this which is much easier to understand
> image
> 
> Adding a simple shadow has a huge performance impact when many paths are added. Should we continue down the path of adding shadows?
> 
> 1. I think we can see if performance improves with the library Val was testing.
> 2. If it is feasible to have 100+ robots in conflict at the same time in real life, we should think about having a simpler animation to optimize performance.

The rmf traffic negotiation system currently doesn't scale to hundreds of robots, I think it most it can handle now is 5-10 robots conflicting with each other. It would be great if we can improve the shadow performance but if we can't it should be fine since we should never expect so much conflicts happening at once.

ChawinTan:
> The rmf traffic negotiation system currently doesn't scale to hundreds of robots, I think it most it can handle now is 5-10 robots conflicting with each other. It would be great if we can improve the shadow performance but if we can't it should be fine since we should never expect so much conflicts happening at once.

Phew! So far the shadow can work pretty well with around 20-30 robots conflicting each other so I think that is actually a good buffer. 

I am going to look at the performance of `feDropShadow` vs the current work around to see if there is any significant difference or not. There are currently some white edges along conflicting trajectories that Matias pointed out on the current work around of the `feDropShadow`.

koonpeng:
> https://www.chromium.org/developers/design-documents/image-filters this might be interesting, maybe the filters are not using gpu.
> 
> Edit: Some more insight
> It says SVG-on-SVG is enabled if "ganesh" is enabled, which seems to be the GPU rasterization, in my pc chrome://gpu says "Rasterization: Software only. Hardware acceleration disabled". Does anyone know if there is a way to determine if the effects are using GPU?
> 
> More edit:
> And it appears gpu rasterization is not supported in linux, but supported in all other platforms (more rant on why linux desktop sucks). Maybe we can test it in windows or mac or maybe even on android and see the performance?

Just tried it with the force gpu rasterization flag and that indeed improve the performance massively.

ChawinTan:
> > https://www.chromium.org/developers/design-documents/image-filters this might be interesting, maybe the filters are not using gpu.
> > Edit: Some more insight
> > It says SVG-on-SVG is enabled if "ganesh" is enabled, which seems to be the GPU rasterization, in my pc chrome://gpu says "Rasterization: Software only. Hardware acceleration disabled". Does anyone know if there is a way to determine if the effects are using GPU?
> > More edit:
> > And it appears gpu rasterization is not supported in linux, but supported in all other platforms (more rant on why linux desktop sucks). Maybe we can test it in windows or mac or maybe even on android and see the performance?
> 
> Just tried it with the force gpu rasterization flag and that indeed improve the performance massively.

Wow! Just tried it and indeed all the animations were so much better! However, there seem to be greyed out boxes in the background once gpu rasterization is active, sort of like the viewboxes changed colors. Is that an intentional side effect or is that a bug?

koonpeng:
> Wow! Just tried it and indeed all the animations were so much better! However, there seem to be greyed out boxes in the background once gpu rasterization is active, sort of like the viewboxes changed colors. Is that an intentional side effect or is that a bug?

I'm not surprised that it's buggy given that it's not enabled by default.

mxgrey:
> The rmf traffic negotiation system currently doesn't scale to hundreds of robots, I think it most it can handle now is 5-10 robots conflicting with each other. It would be great if we can improve the shadow performance but if we can't it should be fine since we should never expect so much conflicts happening at once.

A few points of clarification on this:
1. The planner can be inefficient with memory at the moment, which means it doesn't scale as well as it should right now, but I expect that problem to be resolved by the end of this month.
2. Because of that scaling issue, a single negotiation also has some difficulty scaling beyond 4-5 participants. However, it's possible to have multiple independent negotiations happening at the same time in different locations. For example, there could be 10 negotiations happening with 4-5 participants per negotiation, which would require us to highlight 40-50 paths.

It's true that we don't need to worry about a scale beyond ~20 right now, but we shouldn't assume that we **never** will.

> Should we continue down the path of adding shadows?

Would it be more efficient to do a simpler solid red outline instead of this shadow that has the fade-away gradient?

ChawinTan:
> Would it be more efficient to do a simpler solid red outline instead of this shadow that has the fade-away gradient?

Probably, I am still trying to find a way to implement this. Right now, the paths are represented by strokes on an svg path which are essentially borders. Trying to draw a border on a border may not be that straight forward so I started off using some form of filters to implement shadows around the stroke. 

I am currently trying out `mask`. It works for some cases but not in others so am still trying to fix that.

koonpeng:
I think the best solution is to try to use gpu acceleration, according to the chromium doc it should be possible even on linux.

mxgrey:
> I think the best solution is to try to use gpu acceleration

I'm open to this, but we might want to examine what the target platforms will be in production. Will customers be irritated if they need to buy dedicated GPUs to use our dashboard?

ChawinTan:
> > I think the best solution is to try to use gpu acceleration
> 
> I'm open to this, but we might want to examine what the target platforms will be in production. Will customers be irritated if they need to buy dedicated GPUs to use our dashboard?

I think he means enable the flags (`GPU rasterization` and ` Out of process rasterization`) in chrome?

mxgrey:
> I think he means enable the flags (GPU rasterization and Out of process rasterization) in chrome?

I'm assuming those flags won't help on a computer that doesn't have a dedicated GPU, but maybe that's a bad assumption on my part.

ChawinTan:
> Thanks @ChawinTan, It's way better now. There are small white lines between the trajectory and the shadow.

@matiasbavera Just fixed this.

koonpeng:
According to the docs it only requires GPU rasterization to be enabled for "SVG-in-SVG" filters, "SVG-in-CSS" and "CSS shorthand" filters should work on all platforms.

matiasbavera:
> A few points of clarification on this:
> 
> 1. The planner can be inefficient with memory at the moment, which means it doesn't scale as well as it should right now, but I expect that problem to be resolved by the end of this month.
> 2. Because of that scaling issue, a single negotiation also has some difficulty scaling beyond 4-5 participants. However, it's possible to have multiple independent negotiations happening at the same time in different locations. For example, there could be 10 negotiations happening with 4-5 participants per negotiation, which would require us to highlight 40-50 paths.
> 

Oh! Good to know that. Thanks for the explanation. Thinking ahead, would you have something else in mind that should be represented with these trajectories? that could result in an extra animation or some variation? Like for example some kind of timer?

> It's true that we don't need to worry about a scale beyond ~20 right now, but we shouldn't assume that we **never** will.

I totally agree, we must find a solution that scales correctly.

matiasbavera:
> I'm open to this, but we might want to examine what the target platforms will be in production.

Do you have an idea of this? that will be good to know, we can tackle this offline.

> Will customers be irritated if they need to buy dedicated GPUs to use our dashboard?

I don't want to be in the shoes of the person who has to tell them that =).

> I'm assuming those flags won't help on a computer that doesn't have a dedicated GPU, but maybe that's a bad assumption on my part.

Mmm.. I'm talking without bases here, but I think by enabling this flag, we can also have more intensive use of the integrated GPU. But I should definitely read more about that. If that the case, most of the processors nowadays (not counting server processors and some AMD ones) already has a "decent" (for doing what we need at least), integrated graphic processors.

I'll find a laptop with windows without a dedicated GPU and test it there too.

matiasbavera:
> > Thanks @ChawinTan, It's way better now. There are small white lines between the trajectory and the shadow.
> 
> @matiasbavera Just fixed this.

Awesome. 
image

ChawinTan:
Right now, the current implementation is able to handle up to ~140 conflicting trajectories without a significant slow down. `feDropShadow` scales much poorly, capping out at ~40 conflicting trajectories before a noticeable decline in performance. These 2 results are without enabling the 2 flags for gpu rasterization

Once I enabled the above 2 flags in chrome, the current implementation is able to handle up to ~ 280-300 trajectories before I find the jitters unacceptable, although it is still much smoother than the previous results at 140 trajectories. Performance seems to be about the same across different types of animations. @mxgrey do you think a scale of ~250 conflicting trajectories is acceptable? Or should the implementation scale even further?

ChawinTan:
Some more photos to show the performance with and without gpu rasterization

Without gpu:
No_gpu_rasterization

With Gpu:
gpu_rasterization

Defintely reduces the time spent painting and scripting by a huge margin

koonpeng:
@ChawinTan have you tried the "SVG-in-CSS" and "CSS shorthand" filters? Does it use GPU even without the flag?

mxgrey:
I think by the time we're dealing with a scale of hundreds of vehicles being in conflict with each other, we'll have had several new generations of computers, which will have many times the computing power, so this sounds good to me :+1:

ChawinTan:
> @ChawinTan have you tried the "SVG-in-CSS" and "CSS shorthand" filters? Does it use GPU even without the flag?

I did, but the results are similar to pure SVG filters, capping out at around 140 trajectories. Now that you reminded me of it, I tried it on a different browser `firefox` and it was way better.

ChawinTan:
Hmmm perhaps we can do something like this?

https://stackoverflow.com/questions/32567156/why-dont-css-filters-work-on-svg-elements-in-chrome

koonpeng:
I am still getting white boxes on some trajectories
image

I also looked into the performance, the current implementation is still very slow, especially when zoomed in. Even the mock scenario with 2 trajectories overlapping causes frame rate to drop to < 10 fps for me. I think the slowdown is more to do with the amount of pixels that needs to be rasterized, looking at the chrome profiler the bottleneck is definitely caused by rasterization.

I tried the various form of filter declarations but that doesn't seem to change anything. At this point we may just have to accept that svg filters simply have bad performance in chrome+linux. This affects the robot shadows as well, zooming in to the line of "geths" and panning the map also drops the fps down massively. We may have to forego svg filters and use other tricks like masking, gradient colors etc.

In the long term maybe we can consider looking at canvas/webgl.

ChawinTan:
`anime.js` seem to be not working correctly in storybook for some reason. The trajectory is moving in the "hidden" parts it seems. Tried it with an existing trajectory from the `.json` file but still encountered this issue. 

Will be looking into this for the next couple of days to see if I can fix this.

anime

matiasbavera:
> At this point we may just have to accept that svg filters simply have bad performance in chrome+linux. This affects the robot shadows as well, zooming in to the line of "geths" and panning the map also drops the fps down massively. We may have to forego svg filters and use other tricks like masking, gradient colors etc.
> 
> In the long term maybe we can consider looking at canvas/webgl.

Yeah, sadly we have to accept it. I agree with you.

Here is one guy that has the same issue, and came to your same conclusion. + He added a hacky way of solving the problem too:
"_The easiest thing to do, is to render the shadow to an image an include that in your svg. Starting from that there could be solutions which generate that shadow in different resolutions and pull in the appropriate size on demand_."
https://stackoverflow.com/questions/38581031/svg-filterres-alternative-replacement-strategy

I think we can do that with robots, but definitely can't do it with trajectories.

As for the robots, I think we can rid out of the shadow having a simple border stroke E.g:
image
On conflict:
image

And for the robots with icons, we can do something like this:
image

Isn't as aesthetic as the shadow, but It'd be best in terms of performance.

As for the trajectories itself, we can use tricks like masking, gradient colors as you said. We mix the robot's color + red in a gradient for the color of a conflict trajectory.  

HTH,

ChawinTan:
mask_path

This is my result when using masking. I have been trying this since a while back but could not get it to work on certain trajectories for some reason. It seems like they are not being masked. For those that do work, it looks pretty good though. Any feedback on where could have went wrong?

ChawinTan:
maskpath_fix

Found the fix to the issue finally.

Here's how it looks like in storybook with multiple trajectories. Aesthetically, it is not as pleasing as shadows, but performance is definitely a lot better even without gpu rasterization. Which implementation would you you guys prefer @koonpeng @matiasbavera .

koonpeng:
im fine with how it looks

> And if we decide to use this, maybe we need to modify the trajectry-animations for the outline animation to be red as the current animation would simply overlap the normal outline animation

yea, makes sense, or i am also fine with just burning the outline animation :fire:.

matiasbavera:
Thanks for this @ChawinTan . It feels way smoother, I'm fine with how it looks too.

> yea, makes sense, or i am also fine with just burning the outline animation .
+1 to this.

ChawinTan:
Here are some results that i did with the original implementation (pure javascript) vs anime.js. I compared the results for each of them @300 trajectories and then increase the number of trajectories until the fps is < 10.

<b>Scripting</b>
Anime.js ~ 820 - 850ms
Javascript ~ 400ms

In terms of scripting performance wise, the current Javascript implementation is much better anime.js. This is especially true when the number of trajectories gets very large (>700) as the animation would have started before anything gets painted on the screen. 

<b>Painting</b>
Anime.sj ~ 2000ms
Javascript ~ 2550ms

It seems like anime.js is more optimal when it comes to repainting screen during the animation as it is consistently ~ 500ms faster than just using pure javascript.

<b>Fps</b>
Despite the differences above both implementations have around the same fps performance.

300 trajectories:
~ 21-24 fps for both

400 trajectories
~ 17-18 fps for both

500 trajectories
~ 10-13 fps for both

600 trajectories
~ 10 -12 fps for both

700 trajectories
~ 9-11 fps for both

As mention above, the original Javascript implementation seems to be working better for large number of trajectories since it requires less time to load, while animejs seems to cause the trajectories to only be loaded after the animation has started partially although we probably do not need to work at these scales though.

If you want to play with anime in storybook, here is the implementation to make it work:

```
      const keyFrames = offsets.map(offset => ({
        offset: offset,
        strokeDashoffset: Math.max(2 - offset, strokeDash + 1),
      }));
      anime({
        targets: pathAnim,
        keyframes: keyFrames,
        easing: 'linear',
        duration: animationDuration(trajectory, animationScale),
        loop: false,
      })
```
Base on the results, I do not have any preference over either implementation, although the current implementation seems to be slightly better at larger scales. However, after experimenting with animejs and its api, I can see that it can make implementing new animations, especially complicated ones that have multiple targets less tricky. 

Personally though, I feel that current implementation is good enough for now but we can add in animejs if we have plans for more complicated animations in the future.

What do you guys think @matiasbavera @koonpeng @val ?

matiasbavera:
> Base on the results, I do not have any preference over either implementation, although the current implementation seems to be slightly better at larger scales. However, after experimenting with animejs and its api, I can see that it can make implementing new animations, especially complicated ones that have multiple targets less tricky.
> 
> Personally though, I feel that current implementation is good enough for now but we can add in animejs if we have plans for more complicated animations in the future.
> 
> What do you guys think @matiasbavera @koonpeng @Val ?

Awesome!! Thanks @ChawinTan for the benchmarks and this useful information. I totally agree with your conclusion. 

Just to put it in writing, I think our possible options are the following:
1) Stick with simple and scalable animations.
2) Stick with simple and scalable animations now and change in the future if we need complicated animations. We would have a tradeoff between performance and simplicity.
3) Take time now to move to Animejs, in case in the future we have to add more complicated animations.

I would go with option 1, and in the future, it could be evaluated again whether to continue with 1 or move to 2. Going with option 3 seems to be the riskiest option (for me), since (just for now) we are not clear about the complexity of the animations that we may need. I mean risky, to the amount of time we have to invest to move to Animejs now and the time that it would take to move again to simpler animations in the future in case we need better performance.

vallq:
Hello I just looked through the current updates! Nice work with all the testing @ChawinTan !! @-@ 300 paths..... And also the animejs documentation was probably not the most verbose to work with as well haha. Animejs is not ~the best~ but it definitely is a lot simpler for trickier animations. 

On my side I don't see any big problems with the current animations with JS, but changing to animejs doesn't require much effort! Maybe if we are expecting to put a lot more different animations patterns or tricky paths then this might be worth moving towards.

Regarding canvas and webgl, I did experiment a little with canvas while working on animejs too. It's doable but slightly more complicated that it is now. So it could be a future consideration.

koonpeng:
Note that we can always use both raw web animation and animejs, since web animation is built in on the browser it doesn't have any payload so there is no effect on the app bundle size. We can use the current implementation for simple things and use animejs when we want to do more advanced animations.

ChawinTan:
> Note that we can always use both raw web animation and animejs, since web animation is built in on the browser it doesn't have any payload so there is no effect on the app bundle size. We can use the current implementation for simple things and use animejs when we want to do more advanced animations.

Will remove animejs for now until we need it for more complicated animations. Like what Valerie said, it is easy to port over the animations from animejs

ChawinTan:
@matiasbavera @koonpeng  I think this branch is ready for review

src/components/schedule-visualizer/colors.ts
```diff
@@ -21,29 +8,81 @@ async function _hash(s: string): Promise<ArrayBuffer> {
 
 export default class ColorManager {
   async robotColor(name: string, model: string): Promise<string> {
```
koonpeng, line 10: this doesn't appear to be used anymore

src/components/schedule-visualizer/colors.ts
```diff
@@ -21,29 +8,81 @@ async function _hash(s: string): Promise<ArrayBuffer> {
...
+    }
+    return color;
   }
 
-  private _robotColorCache: Record<string, string> = {};
+  // use vibrant package
+  async robotImageColor(path: string, name: string) {
+    let color = this._robotColorCache[name];
+    if (!color) {
+      const imgHolder = new Image(400, 400);
```
koonpeng, line 36: Do we need to make sure the image dimensions are correct?

src/components/schedule-visualizer/colors.ts
```diff
@@ -21,29 +8,81 @@ async function _hash(s: string): Promise<ArrayBuffer> {
...
+      const nameHash = new Uint16Array(await _hash(name));
+      color = ColorManager._gePathColor(modelHash[0], nameHash[0]);
+      this._pathColorCache[name] = color;
+    }
+    return color;
   }
 
-  private _robotColorCache: Record<string, string> = {};
+  // use vibrant package
+  async robotImageColor(path: string, name: string) {
```
koonpeng, line 33:
```suggestion
  async robotImageColor(path: string, name: string): Promise<string | undefined> {
```

src/components/schedule-visualizer/colors.ts
```diff
@@ -21,29 +8,81 @@ async function _hash(s: string): Promise<ArrayBuffer> {
...
+      color = ColorManager._getLightColor(modelHash[0], nameHash[0]);
+      this._robotColorCache[name] = color;
     }
     return color;
   }
 
-  robotColorFromCache(name: string, model: string): string | null {
-    const color = this._robotColorCache[this._robotKey(name, model)];
-    return color ? color : null;
+  async robotTrajectoryColor(name: string, model: string): Promise<string> {
```
koonpeng, line 21: We could probably combine this with `robotImageColor`

```ts
async robotPrimaryColor(fleet: string, name: string, model: string, image?: string): Promise<string | undefined>
```

What do you think of this?

src/components/schedule-visualizer/colors.ts
```diff
@@ -21,29 +8,81 @@ async function _hash(s: string): Promise<ArrayBuffer> {
...
+          if (rgb) {
+            const colorHolder = `rgb(${rgb[0]}, ${rgb[1]}, ${rgb[2]})`;
+            this._robotColorCache[name] = colorHolder;
+            return colorHolder;
+          }
+        });
+    }
   }
+
+  robotColorFromCache(name: string): string | null {
```
koonpeng, line 56: we might want to make this fleet + name to prevent the super rare case of a name clash.

src/components/schedule-visualizer/robot-trajector-conflict.tsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
koonpeng, line 1: typo in file name

src/settings.ts
```diff
@@ -11,9 +11,22 @@ export enum AnimationSpeed {
   Fast,
 }
 
+export enum TrajectoryDiameter {
+  Fix_size,
```
koonpeng, line 15: enum members should use PascalCase

src/components/schedule-visualizer/tests/colors.test.tsx
```diff
@@ -0,0 +1,33 @@
...
+    const colorPromise = await colorManager.robotColor(mockRobot.name, mockRobot.model);
+    expect(colorPromise).toEqual(colorManager.robotColorFromCache(mockRobot.name));
+  });
+
+  it('should store and return a color when robotTrajectoryColor is called', async () => {
+    const colorPromise = await colorManager.robotTrajectoryColor(mockRobot.name, mockRobot.model);
+    expect(colorPromise).toEqual(colorManager.pathColorFromCache(mockRobot.name));
+  });
+
+  it('should return a promise when robotImageColor is called', done => {
```
koonpeng, line 24: Did you test this? https://create-react-app.dev/docs/adding-images-fonts-and-files/
Not sure if it will work for unit tests.

src/components/schedule-visualizer/tests/robot-trajectories-overlay.test.tsx
```diff
@@ -0,0 +1,69 @@
...
+  it('renders without crashing', () => {
+    const wrapper = createWrapper(
+      RobotTrajectoriesOverlay,
+      mapBound,
+      trajectoryConflict,
+      colorManager,
+      [trajectoryValue],
+      [[]],
+    );
+    wrapper.unmount();
```
koonpeng, line 67: You can add a snapshot test here.

```ts
expect(wrapper.html()).toMatchSnapshot();
```

src/components/schedule-visualizer/tests/robot-trajectory.test.tsx
```diff
@@ -1,33 +1,128 @@
...
+    });
+
+    trajectoryValue = trajectoryData.values[0];
+    trajectoryConflict = trajectoryData.conflicts;
+  });
+
+  it('renders without crashing', () => {
+    const root = createWrapper(RobotTrajectory, trajectoryValue, trajectoryConflict, 0.5, 'black');
+
+    root.unmount();
```
koonpeng, line 69: You can add a snapshot test here.

```ts
expect(root.html()).toMatchSnapshot();
```

src/components/schedule-visualizer/robot-trajectory.tsx
```diff
@@ -22,36 +27,76 @@ export const RobotTrajectory = React.memo(
...
-        fill={'none'}
-        pathLength={1}
-        strokeDasharray={2}
-        strokeDashoffset={0}
-        {...otherProps}
-      />
+      <>
+        {color && (
+          <path
+            id="myPath"
```
koonpeng, line 77: Give it a more meaningful name? e.g. `robot-trajectory-path`.

src/components/schedule-visualizer/tests/robot-trajectory.test.tsx
```diff
@@ -1,33 +1,128 @@
...
     </svg>,
   );
+};
+
+describe('Robot Trajectory', () => {
+  let trajMgr;
+  let trajectoryData;
+  let trajectoryValue: Trajectory;
+  let trajectoryConflict: Conflict[];
+  const themeColor = '#4caf50';
```
koonpeng, line 48: Is it possible to use the actual theme? Something like the equivalent of 
```tsx
const theme = React.useTheme();
```
outside of a react component.

src/components/schedule-visualizer/tests/robot-trajectory.test.tsx
```diff
@@ -1,33 +1,128 @@
...
+  const fixPathSize = 0.4;
```
koonpeng, line 49: We could export this as a const value in robot-trajectory.tsx and check that the value matches.

src/components/schedule-visualizer/tests/robot-trajectory.test.tsx
```diff
@@ -1,33 +1,128 @@
...
+      .find('#myPath')
+      .props().stroke;
+
+    expect(trajWidth).toEqual(fixPathSize);
+    expect(trajColor).toEqual(themeColor);
+
+    root.unmount();
+  });
+
+  it('should set color to a certain shade of green', () => {
```
koonpeng, line 102: There is no expectation in this test.

src/components/schedule-visualizer/tests/robot-trajectory.test.tsx
```diff
@@ -1,33 +1,128 @@
...
+      trajectoryConflict,
+      0.5,
+      'black',
+      mockContext,
+    );
+
+    root.unmount();
+  });
+
+  it('should change path color to conflicting color', () => {
```
koonpeng, line 119: This test doesn't actually check the path color. Maybe rename it to something like `render conflicting path for conflicting trajectories`.

src/components/settings-drawer.tsx
```diff
@@ -21,20 +21,40 @@ export interface SettingsDrawerProps extends DrawerProps {
...
+  );
+
+  const drawerAnchor = useMediaQuery('(max-aspect-ratio: 8/10') ? 'bottom' : 'right';
 
   function handleTrajectoryAnimationChange(ev: React.ChangeEvent<HTMLInputElement>): void {
     const newSettings: Settings = { ...settings, trajectoryAnimation: Number(ev.target.value) };
     onSettingsChange && onSettingsChange(newSettings);
   }
 
+  function handleTrajectoryDiameterChange(ev: React.ChangeEvent<HTMLInputElement>): void {
```
koonpeng, line 48: Some tests for this would be nice. You can use `jest.fn()` to provide a mock, simulate a click in enzyme, and use `toHaveBeenCalledTimes(1)` to check if it works.

src/components/settings-drawer.tsx
```diff
@@ -21,20 +21,40 @@ export interface SettingsDrawerProps extends DrawerProps {
...
+    const newSettings: Settings = { ...settings, trajectoryDiameter: Number(ev.target.value) };
+    onSettingsChange && onSettingsChange(newSettings);
+  }
+
+  function handleTrajectorColorChange(ev: React.ChangeEvent<HTMLInputElement>): void {
```
koonpeng, line 53: Some tests for this would be nice. You can use `jest.fn()` to provide a mock, simulate a click in enzyme, and use `toHaveBeenCalledTimes(1)` to check if it works.

src/settings.ts
```diff
@@ -11,9 +11,22 @@ export enum AnimationSpeed {
...
+  Robot_size,
+}
+
+export enum TrajectoryColor {
+  Theme,
+  Robot_Color,
```
koonpeng, line 21: enum members should use PascalCase

src/setupTests.ts
```diff
@@ -2,3 +2,14 @@ import Enzyme from 'enzyme';
...
+
+const globalAny: any = global;
+class MockEncoder {
+  encode() {
+    return jest.fn();
+  }
+}
+const mockCrypto = { subtle: { digest: () => jest.fn() } };
+
+globalAny.window.TextEncoder = MockEncoder;
```
koonpeng, line 14: How much of the tests require these mocks? If it only affects the robot tests its better to set them up in their respective test cases.

src/stories/baseComponents/trajectory-performance.tsx
```diff
@@ -0,0 +1,67 @@
...
+import Trajectory from './trajectory';
+import { createTrajectories } from '../baseComponents/utils-trajectory';
+import { TextField, makeStyles } from '@material-ui/core';
+import MenuItem from '@material-ui/core/MenuItem';
+import Select from '@material-ui/core/Select';
+import InputLabel from '@material-ui/core/InputLabel';
+import { mapBound } from '../baseComponents/utils';
+import ColorManager from '../../components/schedule-visualizer/colors';
+import { defaultSettings } from '../../settings';
+
```
koonpeng, line 12: you need to import the leaflet css to the map to work properly

```tsx
import 'leaflet/dist/leaflet.css';
```

src/stories/baseComponents/trajectory.tsx
```diff
@@ -0,0 +1,135 @@
...
+  RobotTrajectoryContext,
+  RobotTrajectoriesOverlayProps,
+} from '../../components/schedule-visualizer/robot-trajectories-overlay';
+import {
+  withFillAnimation,
+  withFollowAnimation,
+  withOutlineAnimation,
+} from '../../components/schedule-visualizer/trajectory-animations';
+import { SettingsContext } from '../../components/app-contexts';
+
```
koonpeng, line 20: you need to import the leaflet css to the map to work properly

```tsx
import 'leaflet/dist/leaflet.css';
```

src/stories/baseComponents/utils-default-traj.tsx
```diff
@@ -0,0 +1,7901 @@
+import { Conflict, Trajectory } from '../../robot-trajectory-manager';
```
koonpeng, line 1: Can you provide instructions to generate this file? Also would be good to add a comment at the top to indicate that this file is generated.

src/stories/trajectory.stories.tsx
```diff
@@ -0,0 +1,181 @@
...
+  followAnim: 'Trajectory with follow animation',
+  followAnimMixConflict: 'Follow animation with some trajectories conflicting',
+  followAnimConflict: 'Follow animation with all trajectories conflicting',
+  outlineAnim: 'Trajectory with outline animation',
+  outlineAnimMixConflict: 'Outline animation with some trajectories conflicting',
+  outlineAnimConflict: 'Outline animation with all trajectories conflicting',
+  fillAnim: 'Trajectory with fill animation',
+  fillAnimMixConflict: 'Fill animation with some trajectories conflicting',
+  fillAnimConflict: 'Fill animation with all trajectories conflicting',
+  spam: 'Spam trajectories to test performance (Currently 100 conflicting trajectories)',
```
koonpeng, line 38: This is not used anymore.

koonpeng (changes requested):

src/components/schedule-visualizer/tests/robot-trajectories-overlay.test.tsx
matiasbavera: Adding this to KP comment toJson(wrapper..

src/components/schedule-visualizer/tests/robot-trajectories-overlay.test.tsx
```diff
@@ -0,0 +1,69 @@
+import { createMount } from '@material-ui/core/test-utils';
+import React from 'react';
+import { Map as LMap } from 'react-leaflet';
+import RobotTrajectoriesOverlay, {
+  RobotTrajectoriesOverlayProps,
+} from '../robot-trajectories-overlay';
+import { Conflict, Trajectory } from '../../../robot-trajectory-manager';
+import ColorManager from '../colors';
+import FakeTrajectoryManager from '../../../mock/fake-traj-manager';
+import { mapBound } from '../../../stories/baseComponents/utils';
```
matiasbavera, line 10: Hmm.. not sure if this `mapBound` should be in `src/util` since it's been used here, instead of being located in `src/stories` and import it from there.

src/components/schedule-visualizer/tests/robot-trajectory.test.tsx
matiasbavera: Yeap, an example can be found in `robots-overlay.test.tsx`

src/settings.ts
```diff
@@ -11,9 +11,22 @@ export enum AnimationSpeed {
...
+  Shades,
+}
+
 export interface Settings {
   trajectoryAnimation: TrajectoryAnimation;
   trajectoryAnimationSpeed: AnimationSpeed;
+  trajectoryDiameter: TrajectoryDiameter;
+  trajectoryColor: TrajectoryColor;
```
matiasbavera, line 29: JFYI,

```
// Enums are not type safe =)
enum Num {
  zero = 0,
  one = 1,
  two = 2,
}
let num: Num = 3;
console.log(num)

```

matiasbavera (commented):
Left a couple of comments. HTH

src/components/schedule-visualizer/colors.ts
ChawinTan: Still in use in `robot-default-icon`.

src/components/schedule-visualizer/colors.ts
ChawinTan: These were placeholder before the `node-virbrant` packages were introduced. Removed them.

src/components/schedule-visualizer/colors.ts
ChawinTan: I'm good with this, but maybe without the `fleet` parameter because I dun see it used in either methods.

src/settings.ts
koonpeng: hmm interesting, not sure if this is a design decision or a bug in typescript. I think enum is appropriate for this use case though, the alternative would be to use string literals, but it makes passing in variables hard.

```ts
function foo(bar: 'bar'): void {}
const bar = 'bar';
foo(bar); // error, 'bar' is of type 'string', but expected type 'bar'
```

src/components/schedule-visualizer/tests/colors.test.tsx
ChawinTan: Tried importing the icons in the public assets but seems like it is out of scope

src/components/schedule-visualizer/tests/robot-trajectory.test.tsx
ChawinTan: I will remove this for now as it requires a more detailed mock of the colorManager.

src/setupTests.ts
ChawinTan: It affects any test with colorManager instance involved. Currently it only affects robotTrajectoryOverlay and color tests but I will leave it here for now as we might need to access colorManager in other test in the future when improving code coverage or testing components that require colorManager.

src/stories/baseComponents/utils-default-traj.tsx
ChawinTan: Right now I have the script inside my own github. Would it necessary to create one under `osrf`?

src/components/settings-drawer.tsx
ChawinTan: I'm removing the snapshot test for this component for now. The name on the `input` field has random numbers that change on every run.

src/components/schedule-visualizer/robot-image-icon.tsx
```diff
@@ -16,12 +16,28 @@ const RobotImageIcon = React.forwardRef(function(
...
+  const [robotColor, setRobotColor] = useState<string | null>(() =>
+    colorManager.robotColorFromCache(robot.name),
+  );
+
+  React.useLayoutEffect(() => {
+    if (robotColor) {
+      return;
+    }
+    (async () => {
+      await colorManager.robotImageColor(iconPath, robot.name).then(color => {
```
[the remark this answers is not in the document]
ChawinTan: Correct me if I am wrong, but what you trying to say is if vibrant fails to extract a color then we can directly load the default icon?

Because I thought the check for the loading of image icons is already done in `robot.tsx`?

src/components/schedule-visualizer/tests/robot-trajectories-overlay.test.tsx
ChawinTan: I created a new mapbound in the test itself since it is only for this particular test module

src/components/schedule-visualizer/robot-image-icon.tsx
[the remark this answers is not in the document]
matiasbavera: > Because I thought the check for the loading of image icons is already done in `robot.tsx`?

Mmmhh.. `robots.tsx` checks if a path exists for a fleet or fleet model and it doesn't check for loading success. `robot-image-icon.tsx` is where the image is loaded and the <image> component is the one in charge of notifying us if the load failed `onError=`(due to a bad path or a missing icon). If the load fails, it throws an error and the `robot.tsx ` component is in charge of switching to the default icon.

> Correct me if I am wrong, but what you trying to say is if vibrant fails to extract a color then we can directly load the default icon?

Yeah, if vibrant fails because it wasn't able to get the image. If it fails for something else, we can either set a default color or as you said load directly the default icon.

ChawinTan:
@koonpeng @matiasbavera Made the changes already

settings_pascalcase

Right now some of the options in settings are in PascalCase. I intend to change the display to normal english in a separate pr. Do you think this is necessary or it is fine as it is?

koonpeng:
> Right now some of the options in settings are in PascalCase. I intend to change the display to normal english in a separate pr. Do you think this is necessary or it is fine as it is?

Is it a big effort to change them in this PR? I haven't look at the details but it would seem to just be a small change.

ChawinTan:
> > Right now some of the options in settings are in PascalCase. I intend to change the display to normal english in a separate pr. Do you think this is necessary or it is fine as it is?
> 
> Is it a big effort to change them in this PR? I haven't look at the details but it would seem to just be a small change.

Initially I tried using string enums, but they were giving me quite a lot of problems further down. My current implementation is not the best but since it is only calculated once, I think it's ok for now.

matiasbavera (commented):
Thanks for the changes @ChawinTan, left a couple of comments. Thanks, for the clean code, it was easy to read.

Found a UI problem:
image
 
Something to improve (Maybe for another PR)

image

There isn't a back or close button here. So you need to click it outside to exit, which is not very intuitive.

HTH,

ChawinTan:
> There isn't a back or close button here. So you need to click it outside to exit, which is not very intuitive.

Created an issue @https://github.com/osrf/romi-dashboard/issues/111

src/stories/trajectory.stories.tsx
```diff
@@ -0,0 +1,180 @@
...
+  <Trajectory
+    bounds={mapBound}
+    conflicts={conflictingTraj.conflicts}
+    colorManager={colorManager}
+    conflictRobotNames={conflictingTraj.conflictingRobotName}
+    trajs={conflictingTraj.trajectories}
+    description={descriptions.conflictingTraj}
+    currSettings={defaultSettings()}
+  />
+);
```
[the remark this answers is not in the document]
ChawinTan: I noticed this too. Just realized that it was due to adding the `%` sign to the mask width that caused the issue. I removed it in this module and all the ones in `trajectory-animations`.

src/setupTests.ts
```diff
@@ -2,3 +2,14 @@ import Enzyme from 'enzyme';
 import Adapter from 'enzyme-adapter-react-16';
 
 Enzyme.configure({ adapter: new Adapter() });
+
+const globalAny: any = global;
```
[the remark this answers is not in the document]
ChawinTan: Added. Assigning global to an any type so that I can access the window object

src/components/schedule-visualizer/robot-trajectory.tsx
```diff
@@ -22,36 +28,82 @@ export const RobotTrajectory = React.memo(
...
+      <>
+        {color && (
+          <path
+            id="robotTrajectoryPath"
+            data-component="RobotTrajectory"
+            ref={ref}
+            d={pathD}
+            stroke={color}
+            opacity={0.8}
+            strokeWidth={isConflict && !isOutline ? trajectoryDiameter - 0.1 : trajectoryDiameter}
```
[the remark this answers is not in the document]
ChawinTan: Changed it for now. Only concern I have is that when we have different robots with different diameters, relating to percentage may cause some outlines to be too thin or too thick resulting in an inconsistent look across the UI. We will see how it goes when the actual footprint data for the robot comes in.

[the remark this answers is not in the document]
matiasbavera: Yeah, you're right. Ok, in case we need to change it in the future, what do you think about setting the default to 0.1 (as it was) and have a minimum diameter for the component (let's say it should be always greater than 0.2)

src/components/tests/settings-drawer.test.tsx
```diff
@@ -0,0 +1,101 @@
...
+          settings={settings}
+          onSettingsChange={newSettings => onSettingsChange(newSettings)}
+          open={true}
+        />
+      </SettingsContext.Provider>,
+    );
+
+    wrapper
+      .find('input')
+      .at(6)
```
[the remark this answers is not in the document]
matiasbavera: What I mean was that you can add something like this in the setting-drawer:

```tsx
  {trajColorText.map((text, i) => (
            <FormControlLabel
              key={i}
              className={classes.flexBasis}
              value={i}
              control={<Radio name={text} />}
              label={text}
            />
          ))} 
```

And find it this way on your tests:

```
    wrapper.find('[name="Theme"]').find('input').simulate('change', mockEvent);
    expect(onSettingsChange).toHaveBeenCalledTimes(1);
```

Because if you are using at(x) you are depending on the position of the inputs, which is not good because if in the future we change the positions of those (radio) inputs we will have to modify our tests

matiasbavera (commented):
Thanks for the changes! everything seems to be working properly. Left my last comment, once addressed that, it should be ready to go!

src/components/schedule-visualizer/robot-trajectory.tsx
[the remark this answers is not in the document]
ChawinTan: Sounds good!

ChawinTan:
unit_test

> Because if you are using at(x) you are depending on the position of the inputs, which is not good because if in the future we change the positions of those (radio) inputs we will have to modify our tests

I tried to look for a unique identifier too when I was writing my unit tests initially, but material-ui would generate a random name that is different every time you run the test. This is also why I excluded snapshot testing for this component. At the moment, I can't find any other way other than using the position of the buttons to access them.

koonpeng:
> I tried to look for a unique identifier too when I was writing my unit tests initially, but material-ui would generate a random name that is different every time you run the test. This is also why I excluded snapshot testing for this component. At the moment, I can't find any way other than using the position of the buttons to access them at the moment

I think he means adding your own attributes so that you can find them, the first `find` will look for your component, then the 2nd `find` will look for html component that is descendants of that component, e.g.

```html
<div id='myDiv'>
  <input type="radio" />
</div>
```

First you find `#myDiv`, then you find `input[type="radio"]`. I think in css you can combine them to a single selector, not sure how is it in enzyme.

For the snapshot problem, are you able to find any solutions online? I'm sure many material ui users would have the same problem.

src/components/settings-drawer.tsx
```diff
@@ -21,20 +21,54 @@ export interface SettingsDrawerProps extends DrawerProps {
...
+    const keys = Object.keys(TrajectoryDiameter).slice(
+      Object.keys(TrajectoryDiameter).length * 0.5,
+    );
+    return keys.map(key => {
+      let text = '';
+      key.split(/(?=[A-Z])/).forEach(k => {
+        text += k + ' ';
+      });
+      return text.trim();
+    });
```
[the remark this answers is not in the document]
koonpeng: Since `TrajectoryDiameter` is a enum, you can just "hardcode" the text in an object.

https://www.typescriptlang.org/play?#code/KYOwrgtgBAKgTgQwFbAMYBcD2cCeARASwQmHWDigG8AoKKAMQIA8BlAgL2ABpaoAlTACNM6Npx4BfatVSYQAZ3RR0iFBmz4iJMnBjAm6AFxUoAbQDWUAiFiq0WXIWKlyAXWOK41gOZQJUAF4qXlN4ZHsNJ21yADpGVg5gdygAcnjgABMoMWAUnjpQu3VHLRc4GIFhUUTklMqRbMS86gkAbmlZBUwAG2AY7sxvAAoVcOLNZx09A0Kxhwno8vic1wBKds75Hr6B4dG1eaiy6fRZg8jSnQqhERX16iA

ChawinTan:
> > I tried to look for a unique identifier too when I was writing my unit tests initially, but material-ui would generate a random name that is different every time you run the test. This is also why I excluded snapshot testing for this component. At the moment, I can't find any way other than using the position of the buttons to access them at the moment
> 
> I think he means adding your own attributes so that you can find them, the first `find` will look for your component, then the 2nd `find` will look for html component that is descendants of that component, e.g.
> 
> ```
> <div id='myDiv'>
>   <input type="radio" />
> </div>
> ```
> 
> First you find `#myDiv`, then you find `input[type="radio"]`. I think in css you can combine them to a single selector, not sure how is it in enzyme.
> 
> For the snapshot problem, are you able to find any solutions online? I'm sure many material ui users would have the same problem.

I did try for a whole day, but this was as close as I got https://github.com/mui-org/material-ui/issues/9492 . 

Oh  I see ... Thanks for the clarification.  Populating the `name` attribute on the `FormControlLabel` seems to have solved this issue. :smiley:

src/components/tests/settings-drawer.test.tsx
```diff
@@ -27,6 +28,7 @@ describe('Settings Drawer', () => {
         <SettingsDrawer settings={settings} onSettingsChange={onSettingsChange} open={true} />
       </SettingsContext.Provider>,
     );
+    expect(toJson(wrapper)).toMatchSnapshot();
```
matiasbavera, line 31: This is generating a massive snapshot file (+5000). I suggest we do something like this

``` ts
  it('should render correctly', () => {
    const wrapper = shallow(
      <SettingsContext.Provider value={settings}>
        <SettingsDrawer settings={settings} onSettingsChange={onSettingsChange} open={true} />
      </SettingsContext.Provider>,
    );
    expect(toJson(wrapper.dive())).toMatchSnapshot();
    wrapper.unmount();
  });

```
https://enzymejs.github.io/enzyme/docs/api/ShallowWrapper/dive.html

That will generate this:

```
// Jest Snapshot v1, https://goo.gl/fbAQLP

exports[`Settings Drawer should render correctly 1`] = `
<WithStyles(ForwardRef(Drawer))
  PaperProps={
    Object {
      "className": "makeStyles-drawer-1",
    }
  }
  anchor="right"
  open={true}
>
  <WithStyles(ForwardRef(FormControl))
    component="fieldset"
  >
    <WithStyles(ForwardRef(FormLabel))
      className="makeStyles-legendLabel-2"
      component="legend"
    >
      Trajectory Animation
    </WithStyles(ForwardRef(FormLabel))>
    <WithStyles(ForwardRef(Divider)) />
    <ForwardRef(RadioGroup)
      className="makeStyles-trajGroup-3"
      onChange={[Function]}
      value={2}
    >
      <WithStyles(ForwardRef(FormControlLabel))
        className="makeStyles-flexBasis-4"
        control={<ForwardRef(WithStyles) />}
        key="0"
        label="None"
        name="None"
        value={0}
      />
      <WithStyles(ForwardRef(FormControlLabel))
        className="makeStyles-flexBasis-4"
        control={<ForwardRef(WithStyles) />}
        key="1"
        label="Fill"
        name="Fill"
        value={1}
      />
      <WithStyles(ForwardRef(FormControlLabel))
        className="makeStyles-flexBasis-4"
        control={<ForwardRef(WithStyles) />}
        key="2"
        label="Follow"
        name="Follow"
        value={2}
      />
      <WithStyles(ForwardRef(FormControlLabel))
        className="makeStyles-flexBasis-4"
        control={<ForwardRef(WithStyles) />}
        key="3"
        label="Outline"
        name="Outline"
        value={3}
      />
    </ForwardRef(RadioGroup)>
    <WithStyles(ForwardRef(FormLabel))
      className="makeStyles-legendLabel-2"
      component="legend"
    >
      Trajectory Diameter
    </WithStyles(ForwardRef(FormLabel))>
    <WithStyles(ForwardRef(Divider)) />
    <ForwardRef(RadioGroup)
      className="makeStyles-trajGroup-3"
      onChange={[Function]}
      value={1}
    >
      <WithStyles(ForwardRef(FormControlLabel))
        className="makeStyles-flexBasis-4"
        control={<ForwardRef(WithStyles) />}
        key="0"
        label="Fix Size"
        name="Fix Size"
        value={0}
      />
      <WithStyles(ForwardRef(FormControlLabel))
        className="makeStyles-flexBasis-4"
        control={<ForwardRef(WithStyles) />}
        key="1"
        label="Robot Size"
        name="Robot Size"
        value={1}
      />
    </ForwardRef(RadioGroup)>
    <WithStyles(ForwardRef(FormLabel))
      className="makeStyles-legendLabel-2"
      component="legend"
    >
      Trajectory Color
    </WithStyles(ForwardRef(FormLabel))>
    <WithStyles(ForwardRef(Divider)) />
    <ForwardRef(RadioGroup)
      className="makeStyles-trajGroup-3"
      onChange={[Function]}
      value={1}
    >
      <WithStyles(ForwardRef(FormControlLabel))
        className="makeStyles-flexBasis-4"
        control={<ForwardRef(WithStyles) />}
        key="0"
        label="Theme"
        name="Theme"
        value={0}
      />
      <WithStyles(ForwardRef(FormControlLabel))
        className="makeStyles-flexBasis-4"
        control={<ForwardRef(WithStyles) />}
        key="1"
        label="Robot Color"
        name="Robot Color"
        value={1}
      />
      <WithStyles(ForwardRef(FormControlLabel))
        className="makeStyles-flexBasis-4"
        control={<ForwardRef(WithStyles) />}
        key="2"
        label="Shades"
        name="Shades"
        value={2}
      />
    </ForwardRef(RadioGroup)>
  </WithStyles(ForwardRef(FormControl))>
</WithStyles(ForwardRef(Drawer))>
`;

```

With this method, we will be able to understand what's the component rendering by just looking at the snapshot.

matiasbavera (commented):
Thanks for the changes @ChawinTan, one last change and it should be ready to go.

matiasbavera (approved):
LGTM!

PR #80, opened 2020-08-14
[ready] Negotiation status panel

ddengster:
changes:
- Added negotiation status panel, which displays a tree of conflict negotiations
- Ability to display trajectories for a certain negotiation upon selecting a tree item
- Moved websocket to appconfig from robot trajectory manager to accomodate websocket message flow

to build, git clone these dependencies into your workspace: 
- osrf/rmf_core
- osrf/rmf_schedule_visualizer ~~(edit: with this branch https://github.com/osrf/rmf_schedule_visualizer/tree/negotiations_subscribe)~~ master
- osrf/romi-dashboard this branch
- osrf/soss
- osrf/traffic_editor

to run:
~~- do the same to display a map and get robots moving, with 1 exception: run `ros2 launch visualizer server_keephistory.xml` instead of `ros2 launch visualizer server.xml`~~
`ros2 launch visualizer server.xml ` has been updated use that!


src/components/app.tsx
```diff
@@ -9,7 +9,10 @@ import DispenserStateManager from '../dispenser-state-manager';
 import DoorStateManager from '../door-state-manager';
 import FleetManager from '../fleet-manager';
 import LiftStateManager from '../lift-state-manager';
-import { RobotTrajectoryManager } from '../robot-trajectory-manager';
+import * as NegotiationStatusManager from '../negotiation-status-manager';
```
matiasbavera, line 12: You can also do this for avoid importing the whole package.
`import NegotiationStatusManager from '../negotiation-status-manager';`

and use just `NegotiationStatusManager` instead of `NegotiationStatusManager.default`

src/components/app.tsx
```diff
@@ -285,11 +306,18 @@ export default function App(props: AppProps): JSX.Element {
     setLiftSpotlight(undefined);
     setRobotSpotlight(undefined);
     setDispenserSpotlight(undefined);
+    setNegotiationSpotlight(undefined);
   }
 
+  function removeAllNegotiationTrajectories(): void {
+    for (var key of Object.keys(negotiationTrajStore))
+      delete negotiationTrajStore[key];
+  }
```
matiasbavera, line 315: If I'm understanding right and the objective it's to remove all negotiations, you can do the following:
```
const [negotiationTrajStore, setNegotiationTrajStore] = React.useState
    <Record<string, NegotiationTrajectoryResponse>>({});
```
on line 165 and 
you can use 
`setNegotiationTrajStore({})` to remove them all.


When you're doing this 
```
const [negotiationTrajStore, setNegotiationTrajStore] = React.useState
    <Record<string, NegotiationTrajectoryResponse>>({});
```
You're defining a state that it'll be a store here `negotiationTrajStore` and the store modifier which is this `setNegotiationTrajStore`, in theory, you should modify the state only with its state modifier.

HTH.

src/negotiation-status-manager.ts
```diff
@@ -0,0 +1,152 @@
...
+    
+    this.removeOldConflicts();
+  }
+
+  allConflicts(): Record<number, NegotiationConflict> {
+    return this._conflicts;
+  }
+
+  startSubscription() {
+    var negotiation_status_mgr = this;
```
matiasbavera, line 113: const?

matiasbavera, line 113: Not for now, but you can check the arrow functions and its lexical scope if you want to get rid of this references. 

https://www.freecodecamp.org/news/learn-es6-the-dope-way-part-ii-arrow-functions-and-the-this-keyword-381ac7a32881/#:~:text=While%20in%20ES5%20'this'%20referred,method%20or%20the%20object%20itself.

src/negotiation-status-manager.ts
```diff
@@ -0,0 +1,152 @@
...
+  public participant_ids_to_status : Record<string, NegotiationStatusData> = {};
+  public resolved : ResolveState = ResolveState.UNKNOWN;
+};
+
+export default class NegotiationStatusManager extends EventEmitter<Events> {
+  constructor(private _backend_ws ?: WebSocket) {
+    super();
+  }
+
+  updateNegotiationStatus(msg : any) {
```
matiasbavera, line 43: Let's try not to use `any`, `any` it's our last resource. You can create custom types to accept different kind of inputs. 
type Test = string|number|record<string, string>

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
+import React from 'react';
+//import DispenserItem from './dispenser-item';
+import { makeStyles } from '@material-ui/core/styles';
+import TreeView from '@material-ui/lab/TreeView';
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import ChevronRightIcon from '@material-ui/icons/ChevronRight';
+import TreeItem from '@material-ui/lab/TreeItem';
+import { SpotlightValue } from './spotlight-value';
+import * as NegotiationStatusManager from '../negotiation-status-manager';
```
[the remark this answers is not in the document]
ddengster: using `import NegotiationStatusManager from '../negotiation-status-manager'` won't compile with `NegotiationStatusManager.NegotiationConflict`

src/app-config.ts
```diff
@@ -29,16 +30,20 @@ if (!process.env.REACT_APP_MOCK && process.env.NODE_ENV !== 'test') {
     throw new Error('REACT_APP_TRAJECTORY_SERVER env variable is needed but not defined');
   }
 
+  var ws = new WebSocket(trajServer);
```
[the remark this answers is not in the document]
ddengster: changed to const 789fed3

src/components/app.tsx
ddengster: replaced 789fed3

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
+import React from 'react';
+//import DispenserItem from './dispenser-item';
```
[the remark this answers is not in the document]
ddengster: removed comment 789fed3

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const { spotlight } = props;
+
+  React.useEffect(() => {
+    if (!spotlight) {
+      return;
+    }
+    // TODO: spotlight
+  }, [spotlight]);
+  
+  /**  Utility conversion functions **/
```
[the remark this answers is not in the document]
ddengster: pushed down 789fed3

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+  };
+
+  // keep track of parameters so we can send them as requests
+  var nodeid_to_parameters = new Map<string, Parameter>();
+
+  /** Render Negotiation panel contents **/
+  const renderNegotiations = (version : string, conflict : NegotiationStatusManager.NegotiationConflict) => {
+    let conflict_label = "Conflict #" + version + ", Participants: ";
+    var i = 0;
+    for (var name of Object.values(conflict.participant_ids_to_names))
```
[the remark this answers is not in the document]
ddengster: replaced with const 789fed3

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+      props.negotiationTrajStore["L1"] = resp;
+    };
+
+    updateNegotiationTrajectory();
+  };
+
+  return (
+    <TreeView
+      //className={classes.root}
```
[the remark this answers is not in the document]
ddengster: removed 789fed3

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+      negotiation_contents.push(contents);
+    });
+  }
+  else {
+    console.log("prop negotationstatus empty");
+  }
+  
+  // action callbacks
+  const handleSelect = (event: React.ChangeEvent<{}>, nodeIds: string) => {
+    //console.log("selected: " + nodeIds);
```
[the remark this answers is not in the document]
ddengster: removed 789fed3

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+    async function updateNegotiationTrajectory() {
+      if (!props.trajManager) {
+        return;
+      }
+
+      var traj_params = nodeid_to_parameters.get(nodeIds);
```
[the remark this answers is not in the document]
ddengster: changed to const 789fed3

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+      let contents = renderNegotiations(version, conflict);
+      negotiation_contents.push(contents);
+    });
+  }
+  else {
+    console.log("prop negotationstatus empty");
+  }
+  
+  // action callbacks
+  const handleSelect = (event: React.ChangeEvent<{}>, nodeIds: string) => {
```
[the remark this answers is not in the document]
ddengster: return void 789fed3

src/negotiation-status-manager.ts
```diff
@@ -0,0 +1,152 @@
...
+    {
+      if (status.resolved & ResolveState.RESOLVED)
+        resolved.push(version);
+    }
+    resolved.sort(); //ascending
+    
+    // pop from the front until you reach the desired retain count
+    while (resolved.length !== 0 && resolved.length > retain_count)
+    {
+      var key = resolved[0];
```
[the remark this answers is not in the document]
ddengster: replaced with const 789fed3

src/negotiation-status-manager.ts
```diff
@@ -0,0 +1,152 @@
...
+        else if (msg["type"] === "negotiation_conclusion")
+        {
+          negotiation_status_mgr.concludeNegotiationStatus(msg);
+        }
+      };
+    }
+  }
+
+  removeOldConflicts() {
+    var retain_count = 99; // increase this value for testing
```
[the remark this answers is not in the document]
ddengster: replaced with const 789fed3

src/negotiation-status-manager.ts
ddengster: replaced with const 789fed3

ddengster:
thanks for the review @matiasbavera, let me wait some of the dependencies to be updated to foxy before i continue with the other fixes

src/components/schedule-visualizer/index.tsx
```diff
@@ -302,6 +309,23 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
...
+                  bounds={curMapFloorLayer.bounds}
+                  trajs={props.negotiationTrajStore[curMapFloorLayer.level.name].values}
+                  conflicts={getConflicts(curMapFloorLayer.level.name)}
+                  colorManager={colorManager}
+                  conflictRobotNames={conflictRobotNames}
+                />
+              </RobotTrajectoryContext.Provider>
+            </Pane>
+          )}
+          </LayersControl.Overlay>
```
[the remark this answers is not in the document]
ddengster: yup, that's correct. It's another RobotTrajectoriesOverlay with a different trajectory store.

src/components/negotiations-panel.tsx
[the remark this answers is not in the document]
koonpeng:
```ts
import NegotiationStatusManager from '../negotiation-status-manager'
```
this does a default import, since this module uses something other than the default import, you need to do
```ts
import { ... } from '...'
```
or, use the import * form like the one used now.

I don't think theres anything wrong with the import * form.

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+    else if (status.rejected)
+      status_text += "  [REJECTED]";
+    else if (status.defunct)
+      status_text += "  [DEFUNCT]";
+    else if (parent_resolved === NegotiationStatusManager.ResolveState.RESOLVED)
+      status_text += "  [FINISHED]";
+    else
+      status_text += "  [ONGOING]";
+    return status_text;
+  };
```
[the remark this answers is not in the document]
koonpeng: I prefer it here, I think `NegotationStatusManager` shouldn't change things in the presentation layer.

koonpeng (commented):
There are many style problems that husky should fix, can you check if husky is working?

src/components/schedule-visualizer/index.tsx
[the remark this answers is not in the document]
matiasbavera: Cool, that's a great feature. =)

src/components/negotiations-panel.tsx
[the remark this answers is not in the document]
matiasbavera: Mmm .. yeah you're right in this case because we end up using the whole functionalities stored in the file. But in the case of a library (let's say Lodash), we only want to import the functionalities we want to use and not the whole library; if we import the whole library, our app bundle will end up being bigger. Webpack ends up playing its role here, Webpack is in charge of looking for the functionality we want on the library and its dependencies.  In our case, we access Webpack through react-script.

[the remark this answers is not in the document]
koonpeng: Some interesting info
https://stackoverflow.com/questions/58741044/why-webpack-doesnt-tree-shake-the-lodash-when-using-import-as
https://stackoverflow.com/questions/42051588/wildcard-or-asterisk-vs-named-or-selective-import-es6-javascript/42051641

It appears webpack does tree-shake `import *` but only under certain conditions. I'm not sure if it works for typescript modules, I guess it probably depends on the webpack config in create-react-app.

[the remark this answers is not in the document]
ddengster: shorten import names 5706218

src/robot-trajectory-manager.ts
```diff
@@ -97,6 +111,18 @@ export class DefaultTrajectoryManager {
...
 
+  async negotiationTrajectory(request : NegotiationTrajectoryRequest): 
+    Promise<NegotiationTrajectoryResponse> {
+    const event = await this._send(JSON.stringify(request));
+    const resp = JSON.parse(event.data);
+    this._checkResponse(request, resp);
+
+    if (resp.values === null) {
+      resp.values = [];
+    }
```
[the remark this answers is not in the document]
ddengster: ,so if i do:
```
if (!value) { ... }
```

will it consider all those values you mentioned above?

src/robot-trajectory-manager.ts
```diff
@@ -143,7 +169,7 @@ export class DefaultTrajectoryManager {
 
   private _checkResponse(request: Request, resp: Response): void {
     if (request.request !== resp.response) {
-      console.warn('received response for wrong request');
+      console.warn('received response for wrong request. Request: ' + request.request + 'Response: ' + resp.response);
```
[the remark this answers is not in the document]
ddengster: nice tip! 91ec211

src/components/schedule-visualizer/index.tsx
```diff
@@ -302,6 +309,23 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
               </Pane>
             )}
           </LayersControl.Overlay>
+
+          <LayersControl.Overlay name="Negotiation Trajectories" checked>
+          {curMapFloorLayer && props.negotiationTrajStore[curMapFloorLayer.level.name] && (
+            <Pane>
+              <RobotTrajectoryContext.Provider value={{ Component: TrajectoryComponent }}>
```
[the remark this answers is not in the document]
ddengster: If you mean the code below it gives me errors like `Cannot read property 'map' of undefined
    at Overlay.componentDidUpdate` and `TypeError: this.props.removeLayerControl is not a function
    at Overlay.componentWillUnmount`
```
          <RobotTrajectoryContext.Provider value={{ Component: TrajectoryComponent }}>
            <LayersControl.Overlay name="Robot Trajectories" checked>
            {curMapFloorLayer && (
              <Pane>
                  <RobotTrajectoriesOverlay
                    bounds={curMapFloorLayer.bounds}
                    trajs={curMapTrajectories}
                    conflicts={curMapConflicts}
                    colorManager={colorManager}
                    conflictRobotNames={conflictRobotNames}
                  />
              </Pane>
            )}
            </LayersControl.Overlay>
            <LayersControl.Overlay name="Negotiation Trajectories" checked>
              {curMapFloorLayer && props.negotiationTrajStore[curMapFloorLayer.level.name] && (
              <Pane>
                  <RobotTrajectoriesOverlay
                    bounds={curMapFloorLayer.bounds}
                    trajs={props.negotiationTrajStore[curMapFloorLayer.level.name].values}
                    conflicts={getConflicts(curMapFloorLayer.level.name)}
                    colorManager={colorManager}
                    conflictRobotNames={conflictRobotNames}
                  />
              </Pane>
              )}
            </LayersControl.Overlay>
          </RobotTrajectoryContext.Provider>
```

Note that I need checkboxes for negotiation trajectories so I need to have seperate `LayersControl.Overlay`

src/components/app.tsx
ddengster: replace function with `setNegotiationTrajStore({});` 6fb5321

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+
+export interface NegotiationsPanelProps {
+  conflicts : Readonly<Record<string, NegotiationStatusManager.NegotiationConflict>>;
+  spotlight?: Readonly<SpotlightValue<string>>;
+  trajManager?: Readonly<RobotTrajectoryManager>;
+  negotiationTrajStore : Record<string, NegotiationTrajectoryResponse>;
+}
+
+export default function NegotiationsPanel(props: NegotiationsPanelProps): JSX.Element {
+  const { spotlight } = props;
```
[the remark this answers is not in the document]
ddengster: destructure variables b133429

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+  defunct: {
+    backgroundColor: "red"
+  },
+  unresolved: {
+    backgroundColor: "red"
+  },
+  ongoing: {
+    backgroundColor: "yellow"
+  }
+});
```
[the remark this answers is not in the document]
ddengster: changed to palette b133429

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+    var style = classes.ongoing;
+    if (status.forfeited)
+      style = classes.forfeited;
+    else if (status.rejected)
+      style = classes.rejected;
+    else if (status.defunct)
+      style = classes.defunct;
+    else if (parent_resolved === NegotiationStatusManager.ResolveState.RESOLVED)
+      style = classes.finished;
+    return style;
```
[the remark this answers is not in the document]
ddengster: fix returns, add return type a76d219

src/components/negotiations-panel.tsx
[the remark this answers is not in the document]
ddengster: was a copypaste oversight, nice catch!
fix returns, add return type a76d219

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+  };
+
+  // keep track of parameters so we can send them as requests
+  var nodeid_to_parameters = new Map<string, Parameter>();
```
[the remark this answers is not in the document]
ddengster: align to code style 7076cbf

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+    let reversed_conflicts = Object.keys(props.conflicts).reverse();
+    reversed_conflicts.forEach(version => {
+      const conflict = props.conflicts[version];
+      let contents = renderNegotiations(version, conflict);
+      negotiation_contents.push(contents);
+    });
+  }
+  else {
+    console.log("prop negotationstatus empty");
+  }
```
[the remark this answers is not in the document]
ddengster: doing this stops the panel from updating.

Also, just curious, how is the function useEffect is called? is there a flowchart somewhere?

src/negotiation-status-manager.ts
ddengster: removed reference fd33d8f

src/negotiation-status-manager.ts
ddengster: these functions are only used in one place and I'd rather not make another type, so inlined functions c3fc9b7

src/app-config.ts
```diff
@@ -29,16 +30,20 @@ if (!process.env.REACT_APP_MOCK && process.env.NODE_ENV !== 'test') {
...
   }
 
+  const ws = new WebSocket(trajServer);
+
   appConfig = {
     transportFactory: () => SossTransport.connect(sossNodeName, sossServer, token),
-    trajectoryManagerFactory: () => DefaultTrajectoryManager.create(trajServer),
+    trajectoryManagerFactory: () => DefaultTrajectoryManager.create(ws),
     appResources: ResourceManager.getResourceConfigurationFile(),
+    trajectoryWebSocket : ws,
```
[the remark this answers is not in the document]
ddengster: > open a separate connection

the backend iterates through connections to send negotiation updates so doing this would be doubling bandwidth sent

> trajectory manager to support negotation status 

if we do this it sets up a precedent where anything that needs the websocket gets shoved into the trajectory manager. 

The only thing that will need tending to in future is the onmessage function binding inNegotiationStatusManager's`startSubscription()`, for the case where some other class needs a callback to another message type. ATM we should be fine sticking to this.

src/robot-trajectory-manager.ts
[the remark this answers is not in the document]
matiasbavera: Glad to help

src/robot-trajectory-manager.ts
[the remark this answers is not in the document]
matiasbavera: You're right! 

Please DO NOT use this to check empty objects (dict) {} or empty arrays []

src/components/schedule-visualizer/index.tsx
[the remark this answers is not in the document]
matiasbavera: Yeah, sorry, my bad, I didn't notice that you were sending different values 

``` typescript
<RobotTrajectoryContext.Provider value={RobotTrajContextValue}>
<RobotTrajectoryContext.Provider value={{ Component: TrajectoryComponent }}>
```

Leave it at it was

src/app-config.ts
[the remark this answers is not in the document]
koonpeng: > the backend iterates through connections to send negotiation updates so doing this would be doubling bandwidth sent

Can you explain more on this, do you mean that it will push the negotiation updates to all existing connections?

The trajectory protocol doesn't support interleaved requests atm, it always assumes that the response that it receives will be for the most recent requests sent, sharing connections will cause it interpret another instance's request for its own. This is also the case for negotiation push messages, it will cause it to interpret a negotiation message pushed by the server as a response to another request.

imo, without a proper protocol that can support interleaved messages, sharing connection would just cause more problems down the road.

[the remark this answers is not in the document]
ddengster: > Can you explain more on this, do you mean that it will push the negotiation updates to all existing connections?

See https://github.com/osrf/rmf_schedule_visualizer/blob/master/rmf_schedule_visualizer/src/Server.cpp#L90
Note that this is a server->romidashboard flow. It's not the same trajectory request->respond protocol that was previously setup.

> imo, without a proper protocol that can support interleaved messages, sharing connection would just cause more problems down the road.

Yea I'm very rarely getting exceptions related to that. I see your point; ill switch to use a second one.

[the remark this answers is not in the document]
ddengster: use 2 websockets 3472808
issue https://github.com/osrf/romi-dashboard/issues/102

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+    else if (parent_resolved === NegotiationStatusManager.ResolveState.RESOLVED)
+      style = classes.finished;
+    return style;
+  };
+
+  // keep track of parameters so we can send them as requests
+  var nodeid_to_parameters = new Map<string, Parameter>();
+
+  /** Render Negotiation panel contents **/
+  const renderNegotiations = (version : string, conflict : NegotiationStatusManager.NegotiationConflict) => {
```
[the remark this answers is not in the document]
ddengster: from what i'm reading about Components, you're saying to encapsulate this function and associated callbacks/text conversion functions into a Component? I'm not sure what benefits we can get out of this since it's used in only 1 spot

src/negotiation-status-manager.ts
matiasbavera: Cool! Just a comment (nit) on  https://github.com/osrf/romi-dashboard/commit/c3fc9b79067a3838fad71fb0cbb2b9642755f9d3

```
 const conflict_version : number = msg["conflict_version"];
 const conflict_version_str = conflict_version.toString()
```
conflict_version -> conflictVersion

src/components/negotiations-panel.tsx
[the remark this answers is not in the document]
matiasbavera: Well that's weird, did you put the dependencies correctly?
```
React.useEffect(()=>{

},[deps])
```

If you leave the dependencies empty it'll only run when the component is mounting.
```
React.useEffect(()=>{

},[])
```

Here is a flow chart
https://www.google.com/imgres?imgurl=https%3A%2F%2Fwww.bram.us%2Fwordpress%2Fwp-content%2Fuploads%2F2019%2F03%2Fhook-flow.png&imgrefurl=https%3A%2F%2Fwww.bram.us%2F2019%2F03%2F11%2Freact-hook-flow-diagram%2F&tbnid=FKtZ2-QzKUWBBM&vet=12ahUKEwjE1LqZztrrAhVxMbkGHRm2BAMQMygAegUIARCRAQ..i&docid=mnzBG5Ih5CIsZM&w=762&h=969&q=react%20useeffect%20flowchart&ved=2ahUKEwjE1LqZztrrAhVxMbkGHRm2BAMQMygAegUIARCRAQ

This lecture could help https://medium.com/swlh/react-lifecycle-hooks-71547ef4e7a8 (5 min)

src/negotiation-status-manager.ts
ddengster: nice catch, align code style for file 1912262

src/components/negotiations-panel.tsx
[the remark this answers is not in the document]
ddengster: I've put the functions `determineStatusText`, `determineStyle` and `renderNegotiations` into the useEffect function along with the code, and these are my dependencies `[conflicts, classes, negotiationContents, nodeidToParameters]`

These two `conflicts, classes` dependencies only call the useEffect function when I activate the panel, whereas the other two dependencies cause a redraw all the time with the panel activated.

In all cases, my negotiation panel does not update the rendering of new negotiations.

Let's fix it in a different PR? The rebasing/re-merging onto master is becoming a drag.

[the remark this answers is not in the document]
matiasbavera: Sure thing! Thanks for creating that new Issue =). Resolving this.

src/components/negotiations-panel.tsx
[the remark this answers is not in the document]
matiasbavera: > from what i'm reading about Components, you're saying to encapsulate this function and associated callbacks/text conversion functions into a Component? I'm not sure what benefits we can get out of this since it's used in only 1 spot

Yeah, I was thinking that it'll be easier to test smaller components mocking their props. But it should work either way. Don't worry.

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+    {
+      conflictLabel += name;
+
+      if (i !== (Object.keys(conflict.participantIdsToNames).length - 1))
+        conflictLabel += ", ";
+      ++i;
+    }
+    let conflictStyle = classes.ongoing;
+    if (conflict.resolved === ResolveState.RESOLVED)
+      conflictStyle = classes.finished;
```
matiasbavera, line 115: What do you think about creating a function `getConflictStyle` as you did with `determineStyle`?

```
const getConflictStyle =() =>{
  if (conflict.resolved === ResolveState.RESOLVED)
      return classes.finished;
    else if (conflict.resolved === ResolveState.FAILED)
      return classes.unresolved;

    return classes.ongoing
}
```

const conflictStyle = getConflictStyle()

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+  const renderNegotiations = (version : string, conflict : NegotiationConflict) => {
+    let conflictLabel = "Conflict #" + version + ", Participants: ";
+    let i = 0;
+    for (const name of Object.values(conflict.participantIdsToNames))
+    {
+      conflictLabel += name;
+
+      if (i !== (Object.keys(conflict.participantIdsToNames).length - 1))
+        conflictLabel += ", ";
+      ++i;
```
matiasbavera, line 111: What do you think about creating a function `getConflictLabel` ot handle this logic

```
 let conflictLabel = 'Conflict #' + version + ', Participants: ';
    let i = 0;
    for (const name of Object.values(conflict.participantIdsToNames)) {
      conflictLabel += name;

      if (i !== Object.keys(conflict.participantIdsToNames).length - 1) conflictLabel += ', ';
      ++i;
    }

```
 as you did with determineStyle?

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+      conflictStyle = classes.unresolved;
+
+    let tableDom : JSX.Element[] = [];
+    for (const [participantId, statusData] of Object.entries(conflict.participantIdsToStatus))
+    {
+      // status handling and background
+      const participantName = conflict.participantIdsToNames[participantId];
+        
+      //add 1 or 2 rows of data depending on the sequence
+
```
matiasbavera, line 126: nit: // add
maybe remove extra line?

matiasbavera: Mmmhh.. I'm seeing this pattern on the comments `//word` (without space) in some places, just out of curiosity do you have a rule to place those?

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,256 @@
...
+        const terminalStatus = statusData.terminal;
+        
+        //set text and style for terminal node
+        let terminalLabelText = participantName;
+        terminalLabelText += " -> [";
+        const lastIdx = (terminalStatus.sequence.length - 1);
+        for (let idx = 0; idx < lastIdx; ++idx)
+        {
+          let sequenceId = terminalStatus.sequence[idx].toString();
+          let sequnceIdName = conflict.participantIdsToNames[sequenceId];
```
matiasbavera, line 138: sequnceIdName-> **sequenceIdName**

src/negotiation-status-manager.ts
```diff
@@ -0,0 +1,149 @@
...
+            conflict.resolved = ResolveState.FAILED;
+          
+          this.removeOldConflicts();
+        }
+      };
+    }
+  }
+
+  removeOldConflicts() {
+    const retain_count = 99; // increase this value for testing
```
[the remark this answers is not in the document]
ddengster: style fix 

> Now if we refresh the page all the conflicts are wiped out, right?
yes, that's correct.

> Not for his PR but in the future, do you think we should save the conflicts in the browser localStorage?
The only problem with that is if you restart the gazebo simulation conflicts will be retained from the previous simulation

src/components/negotiations-panel.tsx
ddengster: `determineStyle` and its variants are used repeatedly multiple times hence the abstraction. In this case it's only used in a single place. We could make more abstraction but I'd generally stay away from that since there's more jumping around for definitions as you read the code.

If it still irks you and the team, let me know and I'll change it.

src/components/negotiations-panel.tsx
ddengster: see https://github.com/osrf/romi-dashboard/pull/80#discussion_r486030501

src/components/negotiations-panel.tsx
ddengster: nope. does this repo have a linter i can use?

src/negotiation-status-manager.ts
```diff
@@ -0,0 +1,149 @@
...
+    let resolved : string[] = [];
+
+    for (const [version, status] of Object.entries(this._conflicts))
+    {
+      if (status.resolved & ResolveState.RESOLVED)
+        resolved.push(version);
+    }
+    resolved.sort(); //ascending
```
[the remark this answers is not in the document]
ddengster: fix space f09e6ac

src/negotiation-status-manager.ts
```diff
@@ -0,0 +1,149 @@
...
+  allConflicts(): Record<number, NegotiationConflict> {
+    return this._conflicts;
+  }
+
+  startSubscription() {
+    if (this._backendWs) {
+      this._backendWs.send(
+        JSON.stringify({request: "negotiation_update_subscribe"}));
+
+      //only recv negotiation status messages
```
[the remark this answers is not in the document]
ddengster: remove comment f09e6ac

src/negotiation-status-manager.ts
```diff
@@ -0,0 +1,149 @@
...
+      this._backendWs = new WebSocket(url);
+  }
+
+  allConflicts(): Record<number, NegotiationConflict> {
+    return this._conflicts;
+  }
+
+  startSubscription() {
+    if (this._backendWs) {
+      this._backendWs.send(
```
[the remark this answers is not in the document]
ddengster: add warning f09e6ac

src/components/negotiations-panel.tsx
ddengster: fix variable name f09e6ac

src/components/negotiations-panel.tsx
koonpeng: it uses eslint, you need to set up your dev env to use it. However, the linter rules does not check the stylistic things, there is prettier to make sure the styles are right, which should be automatically ran on commit, but it doesn't seem to happen in this case, maybe something is wrong your ide/nodejs?

src/negotiation-status-manager.ts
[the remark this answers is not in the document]
koonpeng: storing it in localStorage would also miss any negotiations that happens while the page is not loaded. I think this would need a service to query the list of negotiations available.

koonpeng:
more testing would be nice, you can use the codecov report to see which parts are not tested, given that this affects connection to the rmf backend, some e2e tests would be needed also. The codecov now only covers unit test, feel free to add files that cannot be reasonably tested to the ignore list.

src/components/negotiations-panel.tsx
ddengster: apply prettier 75b99c8

ddengster:
> more testing would be nice
added unit test 8369d09

src/negotiation-status-manager.ts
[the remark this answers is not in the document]
matiasbavera: > The only problem with that is if you restart the gazebo simulation conflicts will be retained from the previous simulation

That's a good point. 




> storing it in localStorage would also miss any negotiations that happens while the page is not loaded. I think this would need a service to query the list of negotiations available.

Oh! you're right, having a service to query the list of negotiations available sounds like a good solution.

matiasbavera:
It was all going well but I got these errors after playing around for a couple of minutes. Let's try to find it and handle it properly so that it does not fail and the user cannot continue using the dashboard.
image


HTH.

src/components/negotiations-panel.tsx
matiasbavera: Cool, thanks for the feedback.

Actually, AFAIK there is nothing definite regarding this topic in this project, but it is good that you raised it. I guess as long as the code is understood and testable, it's fine by me. So feel free to leave it as is. Thank you.

matiasbavera:
A couple of ideas not for this PR.
traject


* Differentiate the stored path of the current path on the map. The stored path is appearing now above the robots as you can see in the video.
* Add a button to clear the map of the selected negotiations and clear the selected negotiation on the panel.
### On the panel:
* Place a separator between the conflicts.
* Make the tree levels difference more noticeable. Either by colors or by bold or by putting an explicit indentation.
* Try to use the typography of material UI.
* When pressing in a negotiation on the panel, mark or distinguish which is the selected negotiation.

src/components/tests/negotiations-panel.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    negotiationTrajStore={undefined} />);
+
+  const treeView = root.find(TreeView);
+  expect(treeView).toBeDefined();
+  
+  const treeItem = root.find(TreeItem);
+  expect(treeItem).toBeDefined();
+
+  root.unmount();
+});
```
matiasbavera, line 70: Maybe we can add a test to verify that the colors are being assigned correctly depending on the status of the negotiation. (you can leave it as a TODO if you want to)

You could also add a snapshot test with a fixed number of deals.

matiasbavera (commented):
Code looks good. Found an unhandled error described here https://github.com/osrf/romi-dashboard/pull/80#issuecomment-690619210

koonpeng:
> A couple of ideas not for this PR.
> traject
> 
> * Differentiate the stored path of the current path on the map. The stored path is appearing now above the robots as you can see in the video.
> * Add a button to clear the map of the selected negotiations and clear the selected negotiation on the panel.
> 
> ### On the panel:
> * Place a separator between the conflicts.
> * Make the tree levels difference more noticeable. Either by colors or by bold or by putting an explicit indentation.
> * Try to use the typography of material UI.
> * When pressing in a negotiation on the panel, mark or distinguish which is the selected negotiation.

Another idea, instead of just storing the trajectory, maybe we could save the entire state of the schedule visualizer, this will allow the negotiation playback to show the state of the robots/doors etc at the point of conflict.

ddengster:
> It was all going well but I got these errors after playing around for a couple of minutes. Let's try to find it and handle it properly so that it does not fail and the user cannot continue using the dashboard.

nice catch! this is caused by the sharing of websockets for trajectories, i've seperated them accordingly c7ce5cc
this should be handled when we address https://github.com/osrf/romi-dashboard/issues/102

matiasbavera:
> > It was all going well but I got these errors after playing around for a couple of minutes. Let's try to find it and handle it properly so that it does not fail and the user cannot continue using the dashboard.
> 
> nice catch! this is caused by the sharing of websockets for trajectories, i've seperated them accordingly [c7ce5cc](https://github.com/osrf/romi-dashboard/commit/c7ce5ccf4a3f9a12db69275966ca121c8296a092)
> this should be handled when we address #102

I got the same error but located in a different place

image
image


Steps to reproduce:
1. Comands -> Loop Request -> Number of Loops = 2
2. Comands -> Loop Request -> Number of Loops = 2

It crashes after a couple of minutes.

matiasbavera (commented):
Thanks for the changes, I left a comment. I got an error described here https://github.com/osrf/romi-dashboard/pull/80#issuecomment-691237113  and I wasn't able to reproduce it anymore after several tries. 

HTH,

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,255 @@
...
+        request: 'negotiation_trajectory',
+        param: trajParams,
+      });
+      if (resp.values === undefined) console.warn('values undefined!');
+      if (resp.response !== 'negotiation_trajectory') {
+        console.warn('wrong response, ignoring!');
+        return;
+      }
+
+      negotiationTrajStore['L1'] = resp;
```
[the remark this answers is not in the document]
ddengster: let's do it in another PR. the data flow is not easy to rework. besides, I don't think the server supports multilevel trajectories yet

src/components/tests/negotiations-panel.test.tsx
ddengster: add snapshot test c111ae7

ddengster:
```
Steps I did to produce it:

    Comands -> Loop Request -> Number of Loops = 2
    Comands -> Loop Request -> Number of Loops = 2

It crashes after a couple of minutes.
```
This is not happening on my setup. The commit that seperates the websocket should have made it not possible to get the errors. Also, you may be running on https://github.com/osrf/rmf_schedule_visualizer/ 2 commits ago; the latest commit should fix the problem. My setup is `npm run start:react-auth` and then starting up soss, the office demo and the websocket server seperately.

@koonpeng @matiasbavera Any more blockers?

koonpeng:
I haven't looked at this in detail, based on matias's review, these are the things that I would hope be addressed:

> Differentiate the stored path of the current path on the map. The stored path is appearing now above the robots as you can see in the video.

This seems like a non-trivial defect which I think should be resolve before merging.

> Try to use the typography of material UI.

Should be simple to address, helps improve consistency of the look n feel.

negotiations-panel.tsx#239
> I think this is the hardcoded level, can we replace it with the current level?

Do not wait till another PR to address this.

regarding the `start:react-auth` script, does this PR work?
#107 

more tests:
* codecov reports 45.53% of changes is testing, more tests could be added.
* There is no e2e test, since this feature involves multiple other repos (rmf_core, rmf_demos, rmf_schedule_visualizer), it should have tests to ensure the whole chain is working.

ddengster:
> This seems like a non-trivial defect which I think should be resolve before merging.

> Should be simple to address, helps improve consistency of the look n feel.

I was hoping to handover this to either @vallq or @ChawinTan. Also most notably there is a PR about trajectory differentiation that could use this as a use case while I work on more backend related stuff.

Also, the negotiation trajectories only appear if you click them. If you need to present it or something you can just not click on it for now.

> Do not wait till another PR to address this.

Why so? What are the expectations for the feature PRs? This shouldn't break anything and we can leave it as something to work on later.

koonpeng:
> Also, the negotiation trajectories only appear if you click them. If you need to present it or something you can just not click on it for now.

Merging in this PR means adding the negotiation panel feature, we can't just pretend it's not there by not clicking on it. If you would like to merge it as it is now, I would suggest to add a build flag that disables it by default, and make it only available by explicitly enabling it e.g. `ROMI_DASHBOARD_ENABLE_NEGOTATION=1 npm run build`.

> > Do not wait till another PR to address this.
> 
> Why so? What are the expectations for the feature PRs? This shouldn't break anything and we can leave it as something to work on later.

Because with this hardcoding, this feature doesn't actually work outside of the floor 'L1', we shouldn't merge in something that is known to not work.

Reason is because the way romi-dashboard is managed, `master` should always be deploy-able, merging in a feature that is known to be problematic breaks that, at least these features should be disabled by default if we really want to merge them.

ddengster:
thanks for taking the time to reply @koonpeng. given the current expectations, expect a little more time on the PR before it gets merged in.

matiasbavera:
> thanks for taking the time to reply @koonpeng. given the current expectations, expect a little more time on the PR before it gets merged in.

Count me in for the reviews and let me know if you need some help!

src/components/negotiations-panel.tsx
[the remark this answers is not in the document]
ddengster: 3683491

ddengster:
add e2e test 54abddf

src/components/schedule-visualizer/index.tsx
```diff
@@ -320,19 +320,20 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
...
-                  bounds={curMapFloorLayer.bounds}
-                  trajs={props.negotiationTrajStore[curMapFloorLayer.level.name].values}
-                  conflicts={getConflicts(curMapFloorLayer.level.name)}
-                  colorManager={colorManager}
-                  conflictRobotNames={conflictRobotNames}
-                />
-              </RobotTrajectoryContext.Provider>
-            </Pane>
-          )}
+            {curMapFloorLayer && (
```
matiasbavera, line 321: Maybe you can check here the existence of `props.negotiationTrajStore[curMapFloorLayer.level.name]` and we can keep the `RobotTrajectoriesOverlay` with `traj` as required.

E.g.:` curMapFloorLayer && props.negotiationTrajStore[curMapFloorLayer.level.name] && ....
`


Also, I think there is a `curLevelName` that you can use instead of `curMapFloorLayer.level.name`. To do something like `props.negotiationTrajStore[curLevelName]`

matiasbavera: Just tell if I'm missing something here

src/components/negotiations-panel.tsx
```diff
@@ -61,6 +63,10 @@ export default function NegotiationsPanel(props: NegotiationsPanelProps): JSX.El
     // TODO: spotlight
   }, [spotlight]);
 
+  let curLevelName = "";
+  if (mapFloorLayerSorted)
+    curLevelName = mapFloorLayerSorted[0];
+
```
matiasbavera, line 69: What do you think about doing something like this?

``` ts
  const curLevelName = React.useMemo(() => {
    if (!mapFloorLayerSorted) {
      return ''; // or maybe you can return null
    }
    return mapFloorLayerSorted[0];
  }, [mapFloorLayerSorted]);
```

The idea here is to memoize the result in case the input is the same. In this case, this assignation (curLevelName=return mapFloorLayerSorted[0]) will not occur again until the value mapFloorLayerSorted changes.

e2e/tests/negotiations.test.ts
```diff
@@ -0,0 +1,90 @@
+import { makeLauncher } from '../rmf-launcher';
+import { login, overwriteClick, removeTextFromAutocomplete } from './utils';
+
+import { createMount } from '@material-ui/core/test-utils';
+import React from 'react';
+import { Redirect } from 'react-router';
+import { BrowserRouter } from 'react-router-dom';
```
matiasbavera, line 7: Doesn't seen to be used anymore

e2e/tests/negotiations.test.ts
```diff
@@ -0,0 +1,90 @@
...
+
+  // Disabled till we find out a way to locate that floor icon
+  // Also figure out a way for a non-rejected negotiation status click
+  /*
+  it('renders negotiation trajectory', () => {
+    $('[data-component=viewOptions]').moveTo();
+    $('[data-component=NegotiationTrajCheckbox]').click();
+    expect($('[data-component=RobotTrajectory]')).toBeVisible();
+  });
+  */
```
matiasbavera, line 89: Just to check, this test will fail every time unless we run r`os2 launch visualizer server_keephistory.xml` instead of `ros2 launch visualizer server.xml` right? If that right we shouldn't include it in our e2e CI. If I'm right please ask KP regarding this issue.

matiasbavera (commented):
Thanks for the changes. Left a couple of comments.  

> traject
> 
> * Differentiate the stored path of the current path on the map. The stored path is appearing now above the robots as you can see in the video.


Regarding this, now the trajectory appears under the robots, which is good but now we can notice the difference between the current trajectory and a stored one. Maybe we can choose other colors? Let's ask for an opinion of our trajectory master @ChawinTan =)

src/components/negotiations-panel.tsx
ddengster: is it even worth it for to memoize a string that can be easily evaluated? my understanding of the side effects is that it increases memory usage because it keeps track of dependencies. i understand not re-rendering components, but this seems not worth it. correct me if I'm wrong.

also there's an incoming PR #105 that memoizes the whole panel, let the effort go there

e2e/tests/negotiations.test.ts
ddengster: removed 36d816d

src/components/schedule-visualizer/index.tsx
ddengster: i removed it so the checkbox doesn't unknowingly appear upon clicking a negotiation status TreeItem, causing some UI inconsistency. See 199c136a9289ac3a0eeb52f08dd1dec13571c0f4

replaced with curLevelName 36d816d

src/components/tests/negotiations-panel.test.tsx
```diff
@@ -66,5 +66,7 @@ it('renders negotiations', () => {
   const treeItem = root.find(TreeItem);
   expect(treeItem).toBeDefined();
 
+  expect(root.html()).toMatchSnapshot();
```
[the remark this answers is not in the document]
ddengster: changed to `toJson` 36d816d

e2e/tests/negotiations.test.ts
ddengster: good point. i've set the retained history to 50 with server.xml, not sure why this e2e test still passed on my setup. https://github.com/osrf/rmf_schedule_visualizer/pull/77

ddengster:
> Regarding this, now the trajectory appears under the robots, which is good but now we can notice the difference between the current trajectory and a stored one. Maybe we can choose other colors? Let's ask for an opinion of our trajectory master @ChawinTan =)

or we could get this in and let chawin make the decision :)

src/components/negotiations-panel.tsx
matiasbavera: > is it even worth it for to memoize a string that can be easily evaluated? my understanding of the side effects is that it increases memory usage because it keeps track of dependencies.

I totally agree with you in normal situations, but we should check how are the array `mapFloorLayerSorted` changes affecting to the components re-render. We can do that using React-Profiler. https://reactjs.org/blog/2018/09/10/introducing-the-react-profiler.html.

Sure we can take it a look at that in the FUP PR.

matiasbavera:
> > Regarding this, now the trajectory appears under the robots, which is good but now we can notice the difference between the current trajectory and a stored one. Maybe we can choose other colors? Let's ask for an opinion of our trajectory master @ChawinTan =)
> 
> or we could get this in and let chawin make the decision :)

Yeah, I think you can use the orange color for now (for the conflict trajectories), to represent the stored trajectories. And we can find a better solution in another PR with @ChawinTan

e2e/tests/negotiations.test.ts
matiasbavera: Awesome. The `test:e2e` uses your local backend (maybe you were run it with `keephistory`?). And the `test:e2e:docker` uses the docker setup. But I'm seeing now that the docker CI on GH passed too, so that's weird =/.

src/components/tests/negotiations-panel.test.tsx
matiasbavera: Since KP told us to reduce the number of the codebot warnings and augment the number of test, we should add tests to verify that the colors are being assigned correctly depending on the status of the negotiation, that the status labels are assigned correctly, etc.

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,268 @@
...
+    }
+
+    let nodeIdBase = 'conflict' + version;
+    return (
+      <TreeItem
+        data-component="TreeItem"
+        nodeId={nodeIdBase}
+        key={nodeIdBase}
+        classes={{ label: conflictStyle, selected: conflictStyle }}
+        label={conflictLabel}
```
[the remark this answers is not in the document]
ddengster: add typography to treeview 9a6cb79

ChawinTan:
> or we could get this in and let chawin make the decision :)

> Regarding this, now the trajectory appears under the robots, which is good but now we can notice the difference between the current trajectory and a stored one. Maybe we can choose other colors? Let's ask for an opinion of our trajectory master @ChawinTan =)

Noted. Will bear that in mind. Care to explain why we is it not good to notice the difference between the current trajectory and the stored one? Are you referring to the trajectories on the map or negotiation panel? I can't really understand the problem from the video itself :sweat_smile:

src/components/tests/negotiations-panel.test.tsx
ddengster: 37460d2 
let me know of a way to get TreeItems at the next depth, else that's all i can do for now

src/components/negotiations-panel.tsx
```diff
@@ -0,0 +1,271 @@
...
+      const conflict = conflicts[version];
+      let contents = renderNegotiations(version, conflict);
+      negotiationContents.push(contents);
+    });
+  } else {
+    console.log('prop negotationstatus empty');
+  }
+
+  // action callbacks
+  const handleSelect = (event: React.ChangeEvent<{}>, nodeIds: string): void => {
```
matiasbavera, line 231: I think you will get rid of a lot of these codebot notifications by mocking the `handleSelect` on treeView. You can also simulate  a click, you can check for an example here: 
https://github.com/osrf/romi-dashboard/blob/master/src/components/schedule-visualizer/door/tests/door.test.tsx


```
test('Trigger click event', async () => {
  let clicked = false;
  const handleClick = () => {
    clicked = true;
  };
  const wrapper = mount(
    <svg>
      <DefaultDoor
        v1={[mainDoor.v1_x, mainDoor.v1_y]}
        v2={[mainDoor.v2_x, mainDoor.v2_y]}
        onClick={handleClick}
      />
    </svg>,
  );
  wrapper
    .find(DefaultDoor)
    .at(0)
    .find('g')
    .simulate('click');
  expect(clicked).toBe(true);

  wrapper.unmount();
});
```

e2e/tests/negotiations.test.ts
```diff
@@ -0,0 +1,84 @@
...
+        timeoutMsg: 'expected TreeItem to be not null!'
+      }
+    );
+    console.log('done');
+    
+    var treeItem = $('[data-component=TreeItem]');
+    expect(treeItem).toBeVisible();
+    treeItem.click();
+    
+    var subTreeItem = treeItem.$('[data-component=TreeItem]');
```
matiasbavera, line 70: Const?

matiasbavera (commented):
Left a couple of comments.

After a few more tests and the trajectory differentiation I think we should be ready to go.

matiasbavera:
> Noted. Will bear that in mind. Care to explain why we is it not good to notice the difference between the current trajectory and the stored one? Are you referring to the trajectories on the map or negotiation panel? I can't really understand the problem from the video itself

The idea of ​​being able to observe the trajectories generated by the negotiations (correct me if I'm wrong, Edwin) is to carry out a better analysis of the history and result of the negotiations. For this Edwin is storing in memory a certain amount of negotiations with its corresponding trajectories. In order to make that analysis, we should be able to load onto the map the stored trajectories (genrated by the negotiations).

We currently have two cases:
1. All the robots on the map are still and there is no active trajectory, so we would have no problems to load onto the map the stored trajectories being the same as they are now (both in shape and color).
2. When the robots are moving, the (1) statement changes because when loading the stored trajectories on the map, they are confused with the active trajectories (which is what you see in the video, or actually, is what you don't see because trajectories (active-stored) are not differentiated), so that could be more confusing than helpful. For the reason mentioned above, it is proposed to differentiate these trajectories: the saved ones generated by the trajectories, from the active ones.

@ddengster any feedback or correction on this, it'll be appreciated

ddengster:
> @ddengster any feedback or correction on this, it'll be appreciated

Yea basically if you click on a green negotiation in the negotiation status panel while the robot is moving, it will show the negotiation trajectories in addition to the robot's trajectory (Note the greener trajectory in the video, while the robot isn't moving there is an overlap similar to the scenario i described). 

The idea is to differentiate them, say maybe through the differentiation of color or form from the robot trajectory.

e2e/tests/negotiations.test.ts
ddengster: de9ae9e

src/components/negotiations-panel.tsx
ddengster: the e2e is testing for this. catering for this test would include moving a function and its dependencies to the dashboard/index so that i can test 3 loc so I'd say skip it.

matiasbavera: Hmm ... yeah, you're right, we could let the e2e test that part like you're saying. Mmm .. It seems that all things are encapsulated in NegotiationsPanel, so it will be difficult to test internal methods of the component, so we could leave it as is and do black-box testing (as we have it now).

src/components/tests/negotiations-panel.test.tsx
```diff
@@ -110,6 +110,8 @@ it('renders negotiations correctly', () => {
     expect(classes?.label?.includes("forfeited"));
   }
 
+  treeItem.at(1).simulate('click');
```
matiasbavera, line 113: You can remove this, since we are not expecting anything.

src/components/tests/negotiations-panel.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+  let negotiationStatusManager : NegotiationStatusManager;
+  negotiationStatusManager = new NegotiationStatusManager("");
+
+  let conflicts = negotiationStatusManager.allConflicts();
+  expect(conflicts).toBeDefined();
+
+  negotiationStatusManager.removeOldConflicts();
+  let conflicts2 = negotiationStatusManager.allConflicts();
+  expect(conflicts2).toBeDefined();
+});
```
matiasbavera, line 141: Can you move this to another file? 
Since you testing another file, and another class that is not a React component, you can create a `tests/ `directory in `src/tests/ ` and place it there with the following name:` negotiation-status-manager.test.ts`.

matiasbavera (commented):
Thanks for the changes and your effort @ddengster . Left a couple of comments. Once the trajectory differentiation is addressed, it should be ready to merge.

ddengster:
> Once the trajectory differentiation is addressed, it should be ready to merge.
e176f51 override path color

src/components/tests/negotiations-panel.test.tsx
ddengster: removed 454e67a

src/components/tests/negotiations-panel.test.tsx
ddengster: so i wrote this hoping to see the effects of the codecov bot. it only tests a couple of lines to little benefit; the rest needs to be tested e2e.

while there is that urge to keep everything modular, moving it to another file would bloat the file count which we already have a large number of. i don't see a high count of functions with `negotiation-status-manager` that could warrant such a move so I'd say keep it here.

matiasbavera: > so i wrote this hoping to see the effects of the codecov bot. it only tests a couple of lines to little benefit; the rest needs to be tested e2e.

Oh ok!



> while there is that urge to keep everything modular, moving it to another file would bloat the file count which we already have a large number of. i don't see a high count of functions with `negotiation-status-manager` that could warrant such a move so I'd say keep it here.

Yeah, I guess it's not worth it to move it. You can leave it as it is.



> file count which we already have a large number of.

Haha Yeah! I feel your pain, I wouldn't say I liked it in the beginning too, but the idea of having a file per component/class and a test class, it helps in the long run.

src/components/schedule-visualizer/index.tsx
```diff
@@ -304,11 +310,32 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
                     conflicts={curMapConflicts}
                     colorManager={colorManager}
                     conflictRobotNames={conflictRobotNames}
+                    overridePathColor={undefined}
```
matiasbavera, line 313: You can remove this `overridePathColor={undefined}`. If you don't send the parameter the child will receive undefined. 

Ditto for all the components below

src/components/schedule-visualizer/robot-trajectory.tsx
```diff
@@ -71,7 +74,7 @@ export const RobotTrajectory = React.memo(
         case TrajectoryColor.Shades:
```
matiasbavera, line 74: It seems that the orange color will only appear when the shades option is selected. Can we implement it for others options too?

image

matiasbavera (commented):
Looking good. Adding an example 


It is correct that the trajectory is only marked when the status of the negotiation is [ongoing]?

Left a few comments

src/components/tests/negotiations-panel.test.tsx
ddengster: > the idea of having a file per component/class and a test class, it helps in the long run.

I generally disagree with that given all the tools we have for searching code nowadays, but let's move on.

ddengster:
> It is correct that the trajectory is only marked when the status of the negotiation is [ongoing]?

That's correct.

src/components/schedule-visualizer/index.tsx
ddengster: removed unneeded parameter a895dae

src/components/schedule-visualizer/robot-trajectory.tsx
ddengster: apply to all modes a895dae

matiasbavera (approved):
Thanks for the changes @ddengster. Looks good to me!

PR #82, opened 2020-08-14
Added waypoints on map

matiasbavera:
## What's new
The end-user can now see the waypoints of the map.

image

image

image


matiasbavera:
@ChawinTan  and @vallq What do you think about the color and the word style?

ChawinTan:
I think the color of the points is good. I like the word style in that it emphasizes itself really well and is easily distinguishable but is slightly hard to read when I am using my laptop (a smaller screen).

matiasbavera:
> I think the color of the points is good. I like the word style in that it emphasizes itself really well and is easily distinguishable but is slightly hard to read when I am using my laptop (a smaller screen).

Thanks for the feedback!

matiasbavera:
Rebased.

koonpeng:
how about different shapes for the waypoints? To make it more distinct from the robots.

src/components/schedule-visualizer/tests/waypoints-overlay.test.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { mount } from 'enzyme';
+import L from 'leaflet';
+import { Map as LMap } from 'react-leaflet';
+import Waypoint from '../waypoint';
+import WaypointsOverlay from '../waypoints-overlay';
+import getBuildingMap from '../../../mock/data/building-map';
+
+test('Render robots correctly', async () => {
```
koonpeng, line 9: typo here?

matiasbavera: You're right! Thank you.

matiasbavera:
> how about different shapes for the waypoints? To make it more distinct from the robots.

How about squares? 
image

koonpeng (commented):
image
I think we should make the waypoint layers on top of the door / lifts.

image
Do you think the waypoints should be on top of the robots?

The waypoints don't have any actions on click so let's remove the click cursor.

matiasbavera:
> I think we should make the waypoint layers on top of the door / lifts.

Yeah, sounds good.

> The waypoints don't have any actions on click so let's remove the click cursor.

You're right, I'll remove that.

> Do you think the waypoints should be on top of the robots?

Yeah, it looks better.

koonpeng (approved):
LGTM, I'm not sure if it's better for the waypoints to on on top or below the robots, but let's just go with on top for now, we can change it easily if more people prefer below.

PR #89, opened 2020-08-20
Added places from resource file

matiasbavera:
## What's new

* Added dispensers to Romi Dashboard (DispenserStates + external resource data combined)
* Dispensers position and icon information are consumed from an external resource (JSON file). 
* Created a Robot and Dispenser resources managers.
* Refactor resources TS types.
* Added tests for dispensers
* Added test for robots and dispenser managers

image


src/mock/data/resources.ts
```diff
@@ -0,0 +1,39 @@
+import { ResourceConfigurationsType } from '../../resource-manager';
```
matiasbavera, line 1: Added this file for testing purposes.

matiasbavera:
Hi @koonpeng  if you have some time, can you take a quick look at the dispensers of this PR. Would you like me to replace for a SVG icon? or that image it's ok?.

QQ about this https://github.com/osrf/rmf_demos/pull/122#issuecomment-686219434.  I will look for a human models and see how it looks.

Related PR https://github.com/osrf/rmf_demos/pull/122

koonpeng:
that looks fine to me, but if you can get a svg icon it will be better.

src/components/schedule-visualizer/dispenser-default-icon.tsx
```diff
@@ -0,0 +1,33 @@
...
+      y={-footprint}
+      width={footprint * 2}
+      height={footprint * 2}
+      viewBox="0 0 24 24"
+    >
+      <path
+        fill={theme.palette.success.main}
+        d="M19 3H4.99c-1.11 0-1.98.9-1.98 2L3 19c0 1.1.88 2 1.99 2H19c1.1 0 2-.9 2-2V5c0-1.1-.9-2-2-2zm0 12h-4c0 1.66-1.35 3-3 3s-3-1.34-3-3H4.99V5H19v10zm-3-5h-2V7h-4v3H8l4 4 4-4z"
+        className={classes.clickable}
+      />
```
matiasbavera, line 21: Do we need different colors for the dispensers? If we need it I can create a color manager for the dispensers too.

matiasbavera:
Mmm.. this is weird it's passing on my local env 

image

matiasbavera:
> Mmm.. this is weird it's passing on my local env

Oh, I get it now, since we are using real values now obtained from the resource file to load the places and dispensers, this is failing. @koonpeng should I add the setup step to the CI process. The BIG downside of this is that we will be depending on another project (`rmf-demos`) to get our configuration. The other option it's to have another test file (mainTest.JSON) and depending on a flag we can read the real file or not.

koonpeng:
We already depend on `rmf_demos` for the e2e, it actually makes sense to set up the resource pack, this allows us to test the robot image icons as well. The challenge would be to install the resource pack when building the e2e tests, we have to somehow make sure the same commit is used by the demo world and resource pack.

Here's some idea:
* Have `rmf_demos` "install" the resource pack. I think it currently only exists as "source", installing it would copy it on the output directory which we can find using ament.
* While building the e2e test, find and import the resource pack, this makes sure that the resources are always in sync with the current build of rmf_demos.
* For docker backend, obtain the resource pack from the rmf image. This ensures the resource pack used is always in sync with the docker image of rmf_demos. It's possible but there doesn't seem to be a nice way to do this though https://stackoverflow.com/questions/25292198/docker-how-can-i-copy-a-file-from-an-image-to-a-host#:~:text=8%20Answers&text=Unfortunately%20there%20doesn't%20seem,the%20file%20from%20the%20container.&text=If%20your%20image%20doesn't,as%20suggested%20in%20Igor's%20answer.

Or maybe we should change how we run CI, instead of launching rmf in docker, we run all the CI in docker. Meaning we create an image for e2e, that image would be based on the rmf:nightly image, with the e2e dependencies added (nodejs, chrome/chromium, docker etc). We can then run the CI using `test:e2e` instead of `test:e2e:docker`.

That would also mean `test:e2e:docker` is no longer useful and can be removed. Or we change it to do the same behavior as CI,  i.e. mount the current directory onto the e2e image and run the test there.

koonpeng:
> Or maybe we should change how we run CI, instead of launching rmf in docker, we run all the CI in docker. Meaning we create an image for e2e, that image would be based on the rmf:nightly image, with the e2e dependencies added (nodejs, chrome/chromium, docker etc). We can then run the CI using test:e2e instead of test:e2e:docker.

I think I will go ahead and try this out, it does sounds like it would make the CI more stable

matiasbavera:
> > Or maybe we should change how we run CI, instead of launching rmf in docker, we run all the CI in docker. Meaning we create an image for e2e, that image would be based on the rmf:nightly image, with the e2e dependencies added (nodejs, chrome/chromium, docker etc). We can then run the CI using test:e2e instead of test:e2e:docker.
> 
> I think I will go ahead and try this out, it does sounds like it would make the CI more stable

I like this idea better. 

> I think I will go ahead and try this out, it does sounds like it would make the CI more stable

Cool! Yeah, totally.

matiasbavera:
I added this in the meantime https://github.com/osrf/romi-dashboard/pull/89/commits/ed18f83fc30eb9aec0c7c5eac488c03df392dcc0. To check if that was the real problem.

matiasbavera:
Hey @koonpeng should we put on hold this until https://github.com/osrf/rmf_demos/pull/130 is ready? because `rmf_demos`'  master affects directly to the docker image.

koonpeng:
> Hey @koonpeng should we put on hold this until osrf/rmf_demos#130 is ready? because rmf_demos' master affects directly to the docker image.

Yeah, we should also try and make use of ament to find the dashboard resources, we could try to recreate ament's functionality in nodejs, or write a simple python script to output the path of the resource and run it as a child process. Theres another small problem in that this will overwrite the current `.resources.json`, maybe `get-icons.js` could receive an arg pointing to the resource file to use.

matiasbavera:
> Theres another small problem in that this will overwrite the current `.resources.json`, maybe `get-icons.js` could receive an arg pointing to the resource file to use.

I think there is no problem with overwriting this file, it is made to be overwritten. But added an arg either way

matiasbavera:
Ready for review @koonpeng

src/resource-manager.ts
```diff
@@ -1,44 +1,51 @@
...
-
-export interface RobotResource {
-  icons: Record<string, string>; // Record<ModelName|FleetName, IconPath>
-}
+import { RobotResourceManager, RobotResource } from './resource-manager-robots';
+import { DispenserResourceManager, DispenserResource } from './resource-manager-dispensers';
 
 export interface ResourceConfigurationsType {
-  robots?: Record<string, RobotResource>; // Record<FleetName, RobotResource>
+  robots: Record<string, RobotResource>; // Record<FleetName, RobotResource>
```
koonpeng, line 6: Instead of using `Partial` at most places, maybe we can have robots as optional and use `Required` in the few places (if there is any) that needs to assume that the properties are there.

src/tests/resource-manager-dispensers.test.ts
```diff
@@ -0,0 +1,64 @@
...
+describe('Correct functionally of getIconPath', () => {
+  let manager: DispenserResourceManager;
+  let resourceData: Record<string, DispenserResource>;
+
+  beforeAll(() => {
+    resourceData = fakeResources().dispensers as Record<string, DispenserResource>;
+    manager = new DispenserResourceManager(resourceData);
+  });
+
+  test('dispenserName does not exists', () => {
```
koonpeng, line 14: Can we have more descriptive test names? e.g. "returns falsy when dispenserName does not exist"

src/tests/resource-manager-robots.test.ts
```diff
@@ -0,0 +1,106 @@
...
+describe('Correct functionally of getAvailablePlacesPerFleet', () => {
+  let manager: RobotResourceManager;
+  let resourceData: Record<string, RobotResource>;
+
+  beforeAll(() => {
+    resourceData = fakeResources().robots;
+    manager = new RobotResourceManager(resourceData);
+  });
+
+  test('FleetName does not exists', () => {
```
koonpeng, line 13: See above comment on descriptive names.

src/mock/data/resources.ts
koonpeng, line 1: We can have `app-config` use this when running in mock mode, this will allow us to see the dispensers in `npm run start:mock`.

docker/docker-compose.yml
```diff
@@ -22,7 +22,7 @@ services:
     ipc: host
     network_mode: host
     privileged: true
-    command: npm run test:e2e
+    command: bash -c 'npm run setup:e2e && npm run test:e2e'
```
koonpeng, line 25: We can run setup as part of `test:e2e`, this ensures that it works when ran locally as well.

koonpeng (changes requested):
Thanks for these changes, left a couple of comments.

src/components/commands-panel.tsx
```diff
@@ -142,31 +145,48 @@ export const CommandsPanel = React.memo((props: CommandsPanelProps) => {
...
+                requestDelivery={handleDeliveryRequest}
+                fleetNames={allFleets}
+                robotHandler={resourcesContext.robots}
+              />
+            </AccordionDetails>
+          </Accordion>
+        </>
+      ) : (
+        <Typography className={classes.errorText}>
+          You do not have a configuration file loaded, therefore the commands cannot be executed. To
```
koonpeng, line 186: This error message may be confusing for normal users. I suggest something simpler like `There is an error while loading the commands panel (unable to load places metadata)`.

src/components/schedule-visualizer/dispensers-overlay.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+export const DispensersOverlay = React.memo(
+  (props: DispensersOverlayProps): React.ReactElement => {
+    const { currentFloorName, onDispenserClick, ...otherProps } = props;
+    const viewBox = viewBoxFromLeafletBounds(props.bounds);
+    const footprint = 0.4;
+    const dispenserResourcesContext = React.useContext(ResourcesContext).dispensers;
+    const dispenserStates = React.useContext(DispenserStateContext);
+    /**
+     * We choose to iterate the dispensers inside resources because we get the positions from the resources file and not from the dispenser's state. In case the dispenser doesn't have an entry in the resources file it will not appear in the map, but still will appear in the Omnipanel.
```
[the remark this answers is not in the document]
matiasbavera: Done! Thanks!

src/components/app-contexts.tsx
```diff
@@ -1,24 +1,30 @@
...
+import ResourceManager, {
+  ResourceConfigurationsType,
+  resourceManagerFactory,
+} from '../resource-manager';
 import { defaultSettings, Settings } from '../settings';
 import { NotificationBarContext, NotificationBarProps } from './notification-bar';
 
 /* Declares the ResourcesContext which contains the resources used on the app*/
-export const ResourcesContext = React.createContext<ResourceConfigurationsType>({});
+export const ResourcesContext = React.createContext<ResourceManager>({} as ResourceManager);
```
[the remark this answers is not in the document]
matiasbavera: Yeah, it was a leftover! Thanks!

src/components/schedule-visualizer/image-icon.tsx
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
```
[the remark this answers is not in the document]
matiasbavera: Sure thing!

src/components/schedule-visualizer/tests/dispenser.test.tsx
```diff
@@ -0,0 +1,77 @@
...
+      </svg>,
+    );
+
+    root.find(Dispenser).simulate('click');
+    expect(clicked).toBe(true);
+
+    root.unmount();
+  });
+
+  test(`Don't Fire click event when it has no state`, () => {
```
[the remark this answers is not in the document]
matiasbavera: Sure, added this comment

 We may want to add dispensers to the resource file, but we may not have a way to know the status of that dispenser or the status may not be sent by the backend because it is a new dispenser not yet registered. This means that if the dispenser icon will be shown on the map but it will not have a status in the omnipanel. Therefore, if a click is pressed on the dispenser, it will set an undefined state in the omnipanel.

src/components/tests/commands-panel.test.tsx
```diff
@@ -13,9 +18,22 @@ beforeEach(() => {
...
+  expect(root.find(RobotDeliveryForm).length).toBe(1);
+
+  root.unmount();
+});
+
+it('Renders error on render without context', () => {
   const root = mount(<CommandsPanel allFleets={fleets.map(fleet => fleet.name)} />);
-  const formElements = root.find(LoopForm);
-  expect(formElements.length).toBe(1);
+  expect(root.find(Typography).exists()).toBeTruthy();
```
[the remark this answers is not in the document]
matiasbavera: Yeah, you're totally right! Added and id.

src/components/commands-panel.tsx
matiasbavera: Sounds good!

src/mock/data/resources.ts
matiasbavera: Done!

src/resource-manager-dispensers.ts
```diff
@@ -0,0 +1,54 @@
...
+  x: number;
+  y: number;
+  yaw: number;
+  level_name: string;
+}
+
+export interface DispenserResource {
+  icons: Record<string, string>;
+  location: Location;
+  guid?: string;
```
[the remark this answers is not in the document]
matiasbavera: Thanks for the feedback! Yeah, actually the way I handled it was too convoluted, so I decided to change it.

PR #90, opened 2020-08-21
feat: docker backend

koonpeng:
Do help test this out if you are interested. This builds and runs all of rmf in a container, the dockerfiles could also be useful for official docker support in the corresponding rmf components.


ChawinTan:
Mine is failing at the `npm run build:docker` command

docker_errror

koonpeng:
@matiasbavera most of the errors are compiler warnings so they should be fine. I fixed the undefined symbol error so that should work now.

---

@ChawinTan 
you need to install [docker-compose](https://docs.docker.com/compose/install/)

ChawinTan:
Hmmm it is now giving me

`.FileNotFoundError: [Errno 2] No such file or directory: './docker/rmf/docker-compose.yml`

ChawinTan:
Just ran the command with `sudo` in front and it is working well now :)

matiasbavera:
I promise that I'll review this one, once https://github.com/osrf/romi-dashboard/pull/79 is merged =) (It will drastically decrease the number of files to verify).

matiasbavera:
I think something weird it's happening with the CI. 
image

It has been stuck there for hours.

koonpeng:
> I think something weird it's happening with the CI.
> image
> 
> It has been stuck there for hours.

I think it's because I changed the CI workflows and it's waiting for the non-existent checks. The `Test` workflow is now moved to `Pull Requests / Unit Test (pull_request)`.

docker/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+version: "3"
```
matiasbavera, line 1: Well, I have mixed feelings about this, because we are instantiating 4 containers with the same images to run a different service each.

In general, an image should have only one service and lift that service, but I understand that this is the case because in our particular case we have to build all the rmf projects together with Colcon for it to work (AFAIK).

I'm thinking of this as a hybrid solution, not following the principle of separation of services in the image. But if we are following the theory of a service per container.

I think it worth leaving a comment on the reason of this architecture.

https://docs.docker.com/config/containers/multi-service_container/#:~:text=It's%20ok%20to%20have%20multiple,defined%20networks%20and%20shared%20volumes.

e2e/rmf-launcher.ts
```diff
@@ -1,65 +1,132 @@
...
+
+      process.once('beforeExit', cleanUp);
+      process.once('exit', cleanUp);
+      process.once('SIGINT', cleanUp);
+      process.once('SIGTERM', cleanUp);
+    }
+    return this._instance;
+  }
+
+  private static _instance?: LocalLauncher;
```
matiasbavera, line 86: Should we move private methods at the end of the class?

PR #91, opened 2020-08-25
Optimise trajectory animations with animejs library usage

vallq:
Hi @koonpeng @matiasbavera, 

I experimented and compared with using GSAP and animejs for this PR. So for [GSAP](https://greensock.com/gsap/), while it would be a more robust platform to use, the functionality that we would require (DrawSVG plugin) is only available for paid members of Club GreenSock. I looked for an alternative and [animejs](https://animejs.com/) provides just that.

**Fill**
traj-animations-fill

**Follow**
traj-animations-follow

**Outline**
traj-animations-outline

*For outline there is one path that does not seem to complete itself, I'm not sure why or if it is only my machine that experiences this:
traj animations incomplete (outline)

Would love to hear if there is anything else to be improved!
@ChawinTan feel free to give your input as well c:


koonpeng:
https://greensock.com/js/speed.html interesting benchmark. But note that the page is sponsored by GSAP so I wouldn't be surprised if they a test case that their library excels at, but anyway since it is a paid software and I don't think our use case justify paying for it.

disregarding GSAP, animejs seems to be the 2nd best, only behind web animations. However, in the 3000 dots test animejs slows my browser down to a crawl. It would be interesting to compare animejs vs plain webanimations when we have the benchmark storybook @ChawinTan.

vallq:
Ooooh that's a cool demonstration, I didn't see it. But yea it might not be the most fair test as well hah

matiasbavera:
Awesome, thanks for this @vallq 

> Disregarding GSAP, animejs seems to be the 2nd best, only behind web animations. However, in the 3000 dots test animejs slows my browser down to a crawl. It would be interesting to compare animejs vs plain webanimations when we have the benchmark storybook @ChawinTan.

This would be great. You can read more about this idea here https://github.com/osrf/romi-dashboard/pull/78#issuecomment-679461726 @vallq. We can compare both in the storybook.

koonpeng:
Based on the benchmark, the main bottleneck is in rasterization and not in scripting, animejs makes it easy to create complex animations but the current animations in trajectories are simple enough that we don't really benefit much from it.

animejs would be really interesting when we need to implement more complex animations though.

PR #93, opened 2020-08-31
fix: robot icon subelements selectable

koonpeng:
This is built on top of #92.

For some reason the subelements of a robot icon like text, direction line etc are select-able in storybook, but it's not select-able in the dashboard. This do some refactoring to handle all events in the generic "robot icon" container so the storybook icons works the same way as the dashboard.


matiasbavera:
Rebased

matiasbavera:
Peek 2020-08-31 11-44

PR #94, opened 2020-08-31
feat: ellipsies long robot name

koonpeng:
This is built on top of #93.

The robot name currently is fixed to 8 chars, this uses some tricks with `getComputedTextLength()` to determine the optimal length and append `...` ellipsies at the end.

This also adds a new "Gallery" robot icons story to easily visualize how different robot icon looks.


koonpeng:
The test is failing because `getComputedTextLength()` is not supported in enzyme/jsdom. I tried mocking `React.useRef` but it doesn't work because enzyme overwrites the mock value after it renders the component. What do you think is a better approach to fix this?

  * move `robot` and `robot-overlay` tests to e2e, it will run in a real dom.
  * refactor the text svg out of robot and add a optional prop to customize the text, in unit tests, we will use a dummy component.

koonpeng:
I went with refactoring for now, it adds some extra props to avoid rendering the text for unit tests.

src/components/schedule-visualizer/robot.tsx
```diff
@@ -34,7 +42,17 @@ const Robot = React.forwardRef(function(
...
+  const {
+    robot,
+    footprint,
+    colorManager,
+    fleetName,
+    inConflict,
+    NameLabelComponent,
+    onClick,
+  } = props;
+  const Text = NameLabelComponent || SvgText;
```
matiasbavera, line 54: Can we use something like useMemo here? 

I ran some performance test with React Profiler and I got these numbers

Current:
image

With useMemo
image

HTH

src/components/schedule-visualizer/svg-text.tsx
```diff
@@ -0,0 +1,29 @@
...
+      return;
+    }
+    // svg text does not support auto ellipses, this workaround by testing the text length and
+    // truncate it bit by bit until it fits the icon. Its abit hacky but it shouldn't be too bad
+    // unless the robot name is mega long.
+    for (textElem.textContent = text; textElem.getComputedTextLength() > targetWidth; ) {
+      textElem.textContent =
+        textElem.textContent!.slice(0, textElem.textContent!.length - 6) + '...';
+    }
+  };
```
matiasbavera, line 26: > The test is failing because getComputedTextLength() is not supported in enzyme/jsdom. I tried mocking React.useRef but it doesn't work because enzyme overwrites the mock value after it renders the component. What do you think is a better approach to fix this?
> * move robot and robot-overlay tests to e2e, it will run in a real dom.
> * refactor the text svg out of robot and add a optional prop to customize the text, in unit tests, we will use a dummy component.

What do you think about adding a `try` and `catch` here avoid the extra props in `robot.tsx` and `robots-overlay`?

koonpeng: Thanks for the suggestion, I now fallback to the previous implementation if `getComputedTextLength` is not a function.

src/components/schedule-visualizer/robot.tsx
koonpeng: I removed NameLabelComponent prop now that `SvgText` can work in jsdom. Changed `RobotComponent` in `RobotsOverlay` to use memo as suggested, I'm surprised to see so much impact in a || statement.

src/stories/baseComponents/robot-gallery.tsx
```diff
@@ -0,0 +1,116 @@
...
+// const useStyles = makeStyles(() => ({
+//   spin: {
+//     animation: '$spin 2s linear infinite',
+//   },
+//   '@keyframes spin': {
+//     '100%': {
+//       transform: 'rotate(360deg);',
+//     },
+//   },
+// }));
```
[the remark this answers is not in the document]
koonpeng: removed

src/components/schedule-visualizer/svg-text.tsx
```diff
@@ -0,0 +1,29 @@
...
+    if (!textElem) {
+      return;
+    }
+
+    if (!targetWidth) {
+      textElem.textContent = text;
+      return;
+    }
+    // svg text does not support auto ellipses, this workaround by testing the text length and
+    // truncate it bit by bit until it fits the icon. Its abit hacky but it shouldn't be too bad
```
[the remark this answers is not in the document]
koonpeng: fixed

matiasbavera:
Rebased to remove conflicts.

PR #95, opened 2020-08-31
fix: missing shadows for robot names with spaces

koonpeng:
This is built on top of #94.

This fixes a bug where shadows are missing when the robot name has spaces, this is caused by unescaped filter ids in the elements.


matiasbavera:
Reseted and cherry-picked to solve conflicts 
image

PR #96, opened 2020-09-02
ignore files in test coverage and add new coverage command

ChawinTan:
Excluded unnecessary files in the test coverage so that it is easier to read through the report. Also included a command to run the report to make it easy to generate the report.

Added this fix so that in future, it is easier to see which conditions have yet to be tested when we are writing unit tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ## RoMi-dashboard
 
-test
+tes
 
 [![codecov](https://codecov.io/gh/osrf/romi-dashboard/branch/wip/graph/badge.svg)](https://codecov.io/gh/osrf/romi-dashboard)
 
--- a/README.md
+++ b/README.md
@@ -1,8 +1,10 @@
 ## RoMi-dashboard
 
-tes
+<h1 align="center">Badges</h1>
 
+<div align="center">
 [![codecov](https://codecov.io/gh/osrf/romi-dashboard/branch/wip/graph/badge.svg)](https://codecov.io/gh/osrf/romi-dashboard)
+</div>
 
 Hello! `romi-dashboard` is a web application that provides overall visualization and control over the RoMi-H system.
 
--- a/README.md
+++ b/README.md
@@ -1,12 +1,6 @@
-## RoMi-dashboard
-
-<h1>Badges</h1>
-
-<div>
-
 [![codecov](https://codecov.io/gh/osrf/romi-dashboard/branch/wip/graph/badge.svg)](https://codecov.io/gh/osrf/romi-dashboard)
 
-</div>
+## RoMi-dashboard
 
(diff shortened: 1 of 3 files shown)


koonpeng (changes requested):
can you update the actions to use coverage? Can also add a badge to show the current coverage.

ChawinTan:
This we need to merge this with the master branch first before we can get a coverage number as it is missing a base commit.

ChawinTan:
And while we are at this, are there any other file/folders that don't need to be included in the test?

README.md
```diff
@@ -1,5 +1,13 @@
 ## RoMi-dashboard
 
+<h1>Badges</h1>
```
koonpeng, line 3: no need for a badges header, just put them at the top before `## RoMi-dashboard`

package.json
```diff
@@ -98,5 +99,15 @@
     "**/*.{js,jsx,ts,tsx,json,css,scss,md}": [
       "prettier --write"
     ]
+  },
+  "jest": {
+    "collectCoverageFrom": [
+      "src/**/*.{js,jsx,ts,tsx}",
+      "!<rootDir>/node_modules/",
+      "!<rootDir>/path/to/dir/",
```
koonpeng, line 107: is this a leftover from examples?

koonpeng (commented):
We can exclude the mock modules as well.

package.json
```diff
@@ -98,5 +99,15 @@
     "**/*.{js,jsx,ts,tsx,json,css,scss,md}": [
       "prettier --write"
     ]
+  },
+  "jest": {
+    "collectCoverageFrom": [
```
matiasbavera, line 104: Can we exclude the build folder too? We don't have that one (/build) on the repo, but do have it on our local dev env.

ChawinTan: Added the line to exclude builder folder, but I think if you run the command, the test does not include the build folder by itself unless you are talking about something else?

koonpeng: correct me if im wrong, I think only files in the `src` directory is included based on the first line. So I don't think excluding the build and node_modules directories is needed.

ChawinTan: I think you are right. I usually leave it there just in case and because the documentation had it in the example :sweat_smile:

README.md
```diff
@@ -1,3 +1,5 @@
+[![codecov](https://codecov.io/gh/osrf/romi-dashboard/branch/wip/graph/badge.svg)](https://codecov.io/gh/osrf/romi-dashboard)
```
koonpeng, line 1: is this right? shouldn't it be master branch?

koonpeng (changes requested):

README.md
ChawinTan: Updated.

koonpeng (approved):
Thanks for this, these metrics will be really useful!

PR #97, opened 2020-09-02
Dashboard walkthrough with Reactour.js

vallq:
Changes: 
- Added a walkthrough with Reactourjs. (Tried out React Joyride but that is not as simple to use)

Discussion:
- Which specific features need a step in the walkthrough?
- Having a 'fake' map for the walkthrough guide seems to be the best because specific elements would need to be targeted for the steps to transition smoothly

To Do:
- Add steps for other features
- Finetune each step (Reactour follows a sequential order of the steps, so jumping randomly between different steps causes the tour to not reflect the steps's content properly)

dashboard-walkthrough


matiasbavera:
> * Having a 'fake' map for the walkthrough guide seems to be the best because specific elements would need to be targeted for the steps to transition smoothly

Agreed. We need to think on how to access that tutorial page (or demo) in an easy way and go back to the main project.

vallq:
> Cool! Thanks Val, looking good. We can also play with the colors and styles to make it more enjoyable =).

Yes! We're on the same page for that (: Definitely gonna make it look better.

I have a question though, would you have any suggestion on how the tour steps can be abstracted out into another component?  Asking because the app.ts is getting really long but the steps also require functions from app.tsx

ChawinTan:
Awesome work @vallq !

> Cool! Thanks Val, looking good. We can also play with the colors and styles to make it more enjoyable =).


Perhaps still a little early you may have known it already, but we can use `<typography>` tag to ensure consistent fonts across.

> I have a question though, would you have any suggestion on how the tour steps can be abstracted out into another component? Asking because the app.ts is getting really long but the steps also require functions from app.tsx

There is probably a better way to do it, but from what I can see right now, you can probably factorize out common functionalities 
like

```
        clearSpotlights();
        setShowOmniPanel(false);
        setShowSettings(false);
```
into a separate function to save some code. The content strings can be place in an object in a new file, for example, `tour-helper.ts` in the `util` folder and the content itself can be a separate function.

The core `Tour` component,
```
          <Tour
            steps={tourSteps}
            isOpen={tourState}
            onRequestClose={() => {
              setTourState(false);
              setShowOmniPanel(true);
              setShowSettings(false);
            }}
            lastStepNextButton={<Button>Start using RoMi!</Button>}
          />
```

can be on a different component.

vallq:
> Perhaps still a little early you may have known it already, but we can use <typography> tag to ensure consistent fonts across.
Yep right on that hahah, I was focused on getting things up first before changing the styling!

> There is probably a better way to do it, but from what I can see right now, you can probably factorize out common functionalities

And yep I did that alr in another commit, just not yet pushed oops

> The content strings can be place in an object in a new file, for example, tour-helper.ts in the util folder and the content itself can be a separate function.

Let me try this out and see how it goes! Thanks for the help @ChawinTan 
If you have any design / color scheme in mind too, let me know! We can also discuss on slack, to make things consistent

matiasbavera:
> I have a question though, would you have any suggestion on how the tour steps can be abstracted out into another component? Asking because the app.ts is getting really long but the steps also require functions from app.tsx

Yeah, actually you can also create a file for the `tourSteps` with the necessary information and another for a manager in case you need one. You can use both, with the component Chawin mentioned. 

But with this, I'm not implying that we do not need to refactor the App component  =). Actually a refactor is partially done in the Auth PR.

vallq:
Changes:
- Moved tour component out from app.tsx
- Tour manager handles the creation of the steps
- Changed the styling of the step

To Do:
- Handle testing and code coverage for the tour

dashboard-walkthrough-v2

koonpeng:
Some comments:

* It seems the tour is triggered whenever the page is loaded, it should remember the state and only show on the first visit. It would also help to add a help button at the appbar to re-trigger the tour.
* Clicking outside the popup exits the tour, not sure if this is intended.
* At step 7, 9 and 11, is it possible to highlight the relevant buttons?

vallq:
> It seems the tour is triggered whenever the page is loaded, it should remember the state and only show on the first visit.

Oh okay will look into this! So a refreshed page should not open up the tour again.

> It would also help to add a help button at the appbar to re-trigger the tour.

The button that opens up the tour again is the last button on the right, beside settings. One thing about this is that it opens the last opened step, been trying to figure out how to restart it at step 1 instead.

> Clicking outside the popup exits the tour, not sure if this is intended.

This is the default Tour config, I think it can be disabled so that users have to go through the entire guide first. Let me check!

> At step 7, 9 and 11, is it possible to highlight the relevant buttons?

This is the tricky part, I have not been able to do it without console warnings because how Reactour works is that
1. it finds an element in the DOM
2. it shows the content in the div
3. functions can be executed for that particular step under actions

so to give an example using step 7:
if step 7 highlights the door panel, that means the currentView is at `MainMenu`
moving to step 8 requires
1. changing the currentView to `Doors`,
2. _selecting_ main door `[data-name="main_door"]`
3. expanding main door tab

but for step 8 to select the main door, it has to already be in the view. so this will log an error that says the Tour is unable to find the `[data-name="main_door"]` if we keep step 7's view at the MainMenu. likewise moving back from step 8 to 7, it will say that `[data-item="Doors"]` is not in the DOM
step 7 to 8 console errors

This is why I've not added any selectors for those steps yet, although it would definitely be a lot nicer! It's also the reason why there is a 'lag' in step 8 (in above gif) when the main door tab gets expanded because the current view at that point in time is not showing all the door items yet

vallq:
I have some questions about testing. In this case for the Tour, other than testing the normal route of clicking through the tour guide sequentially, I should also test for non-sequential clicking of the navigation of the tour (ie. user clicks on different navigation dots) right?

And I should only test that the right step is rendered at a particular step, am I right to say that?

koonpeng:
> but for step 8 to select the main door, it has to already be in the view. so this will log an error that says the Tour is unable to find the [data-name="main_door"] if we keep step 7's view at the MainMenu. likewise moving back from step 8 to 7, it will say that [data-item="Doors"] is not in the DOM

maybe this will help, https://github.com/elrumordelaluz/reactour/issues/170, the suggested workaround is to create a phantom step that you go through programatically.

> I have some questions about testing. In this case for the Tour, other than testing the normal route of clicking through the tour guide sequentially, I should also test for non-sequential clicking of the navigation of the tour (ie. user clicks on different navigation dots) right?
>
> And I should only test that the right step is rendered at a particular step, am I right to say that?

I would suggest to try and do unit test if possible, that would allow you to do snapshot testing of each step. You should also be able to use that to test that non-sequential clicking results in the same snapshot.

vallq:
> maybe this will help, elrumordelaluz/reactour#170, the suggested workaround is to create a phantom step that you go through programatically.

okay will try this!

thanks @koonpeng , and noted on the unit testing

matiasbavera:
Hi @vallq, this looks waaay better! =)

> I would suggest to try and do unit test if possible, that would allow you to do snapshot testing of each step. You should also be able to use that to test that non-sequential clicking results in the same snapshot.

Totally agree with KP.

vallq:
Changes:
- custom navigation buttons 
> maybe this will help, [elrumordelaluz/reactour#170](https://github.com/elrumordelaluz/reactour/issues/170), the suggested workaround is to create a phantom step that you go through programatically.

//looked into this and realised that it's for changing the package's implementation instead ):
- removed the navigation dots because those will lead to the console warnings of not being able to find the element 
//will look into a custom navigation dot component here for a better ui
- disable clicking outside of step during the tour
- clicking tour button starts the tour from the first step
- added snapshot testing for each step*

To Do:
- adding a test for clicking of the last step button to close the tour (for some reason enzyme does not find the node even though `root.html()` shows it)
- setting a state to prevent the tour from reloading when the page is refreshed

matiasbavera:
> * adding a test for clicking of the last step button to close the tour

Well, that's weird, just an idea, can you find its parent and check for the button inside? or can you add a custom attribute?

vallq:
@matiasbavera I tried both methods you've recommended and it doesn't work too ): for some reason the parent group is identified as a singular node instead of a parent node with 2 children nodes

src/components/tests/appbar.test.tsx
```diff
@@ -9,39 +9,69 @@ const mount = createMount();
...
   test('renders correctly', () => {
     const root = mount(
-      <AppBar toggleShowOmniPanel={toggleShowOmniPanel} showSettings={showSettings} />,
+      <AppBar
+        toggleShowOmniPanel={toggleShowOmniPanel}
+        showSettings={showSettings}
+        showTour={showTour}
+      />,
     );
     expect(root).toMatchSnapshot();
```
matiasbavera, line 28: Can you try to mount the component with `shallow` because it's generating a kind of unreadable snapshots with the props of material UI. The idea of the snapshots is to generate a readable file where besides of the comparison that the test tooling does, the programmer can look at the code that the component it's generating.

HTH

src/components/tests/appbar.test.tsx
```diff
@@ -9,39 +9,69 @@ const mount = createMount();
...
   });
 
   test('toggles show omnipanel when dashboard button is clicked', () => {
     const root = mount(
-      <AppBar toggleShowOmniPanel={toggleShowOmniPanel} showSettings={showSettings} />,
+      <AppBar
```
matiasbavera, line 33: What do you think about creating a component generator function, since we are mounting the component with the same variables in every test?

```
 <AppBar
        toggleShowOmniPanel={toggleShowOmniPanel}
        showSettings={showSettings}
        showTour={showTour}
      />,
```

Something like `buildDoorWrapper` on `door.test.tsx`

src/components/tests/tour.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+const mount = createMount();
+
+describe('Dashboard Tour', () => {
+  let tourState: boolean = true;
+  let setTourState: jest.Mock = jest.fn();
+  let setTourShowOmniPanel: jest.Mock = jest.fn();
+  let setTourSettingsAndOmniPanel: jest.Mock = jest.fn();
+  let doorSpotlight: SpotlightValue<string> = { value: 'main_door' };
+  let setDoorSpotlight: jest.Mock = jest.fn();
```
matiasbavera, line 20: Can those be `const`? Since you're not assign values to those after its definition?

src/components/tests/tour.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+      OmniPanelViewIndex,
+      doorSpotlight,
+      setDoorSpotlight,
+    };
+    root = mount(<DashboardTour tourProps={tourProps} />);
+  });
+
+  afterEach(() => {
+    root?.unmount();
+  });
```
matiasbavera, line 41: Nice one =)

src/components/tests/tour.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+  const wait = (delay = 0) => new Promise((resolve) => setTimeout(resolve, delay));
+
+  it('renders without crashing', () => {
+    expect(root.html()).toMatchSnapshot();
```
matiasbavera, line 46: Same as the comments of the snapshots before +

root.html() it's converting the `HTML` to string. You can remove the `.html()` and it'll generate a Json format. Give it a try.

Ditto for all snapshots in this file

src/components/tests/tour.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+    expectedText = 'Welcome to RoMi dashboard';
+    stepText = root.find('.reactour__helper').hostNodes().text();
+
+    expect(stepText.includes(expectedText));
+    expect(root.html()).toMatchSnapshot();
+  });
+
+  it('renders zoom button step (2/12)', async () => {
+    expectedText =
+      'Click on the zoom buttons to change the view of the floor plan.Alternatively, the scroll button on your mouse would work too!';
```
matiasbavera, line 59: Space
plan.Alternatively, -> plan. Alternatively,

src/components/tests/tour.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+
+    setTimeout(() => (stepText = root.find('.reactour__helper').hostNodes().text()), 100);
+
+    expect(stepText.includes(expectedText));
+    expect(root.html()).toMatchSnapshot();
+  });
+
+  it('renders floor plan button step (3/12)', async () => {
+    expectedText =
+      'Use the floor plan button to switch between available levels and enabling / disabling the view of different components.';
```
matiasbavera, line 71: I'm thinking that maybe we can store the text in a Dictionary (in the tour file) and import it here.

export tourText {
step1: text,
step2: text
}

matiasbavera (commented):
Thanks @vallq ! Left a couple of comments

matiasbavera:
> * adding a test for clicking of the last step button to close the tour (for some reason enzyme does not find the node even though `root.html()` shows it)
>   cannot find last step btn

```
it('renders traj anims step (12/12)', async () => {
    expectedText =
      'Finally, Trajectory Animations can be changed using the options available. Look out for new features ahead!';

    const nextBtn = root.find('#tour-next-btn').first();
    for (let i = 1; i < 13; i++) {
      nextBtn.simulate('click');
      await wait(5);
    }

    stepText = root.find('.reactour__helper').hostNodes().text();
    console.log(root.find('#tour-last-step-btn').first().debug());

    root.find('#tour-last-step-btn').first().simulate('click');
    expect(stepText.includes(expectedText));
    expect(root.html()).toMatchSnapshot();
  });
```

This works for me, it's finding the button too
```

      <WithStyles(ForwardRef(Button)) variant="contained" color="primary" onClick={[Function: onClick]} id="tour-last-step-btn">
        <ForwardRef(Button) classes={{...}} variant="contained" color="primary" onClick={[Function: onClick]} id="tour-last-step-btn">
          <WithStyles(ForwardRef(ButtonBase)) className="MuiButton-root MuiButton-contained MuiButton-containedPrimary" component="button" disabled={false} focusRipple={true} focusVisibleClassName="Mui-focusVisible" type="button" onClick={[Function: onClick]} id="tour-last-step-btn">
            <ForwardRef(ButtonBase) classes={{...}} className="MuiButton-root MuiButton-contained MuiButton-containedPrimary" component="button" disabled={false} focusRipple={true} focusVisibleClassName="Mui-focusVisible" type="button" onClick={[Function: onClick]} id="tour-last-step-btn">
              <button className="MuiButtonBase-root MuiButton-root MuiButton-contained MuiButton-containedPrimary" onBlur={[Function (anonymous)]} onClick={[Function: onClick]} onFocus={[Function (anonymous)]} onKeyDown={[Function (anonymous)]} onKeyUp={[Function (anonymous)]} onMouseDown={[Function (anonymous)]} onMouseLeave={[Function (anonymous)]} onMouseUp={[Function (anonymous)]} onDragLeave={[Function (anonymous)]} onTouchEnd={[Function (anonymous)]} onTouchMove={[Function (anonymous)]} onTouchStart={[Function (anonymous)]} tabIndex={0} type="button" disabled={false} id="tour-last-step-btn">
                <span className="MuiButton-label">
                  Start Using Romi
                </span>
                <WithStyles(memo) center={false}>
                  <ForwardRef(TouchRipple) classes={{...}} center={false}>
                    <span className="MuiTouchRipple-root">
                      <TransitionGroup component={{...}} exit={true} childFactory={[Function: childFactory]} />
                    </span>
                  </ForwardRef(TouchRipple)>
                </WithStyles(memo)>
              </button>
            </ForwardRef(ButtonBase)>
          </WithStyles(ForwardRef(ButtonBase))>
        </ForwardRef(Button)>
      </WithStyles(ForwardRef(Button))>
```

HTH

src/components/appbar.tsx
```diff
@@ -49,6 +52,9 @@ export default function AppBar(props: AppBarProps): React.ReactElement {
         <IconButton id="show-settings-btn" color="inherit" onClick={() => showSettings(true)}>
           <SettingsIcon />
         </IconButton>
+        <IconButton id="show-tour-btn" color="inherit" onClick={() => showTour(true)}>
+          <FontAwesomeIcon icon={faMapSigns} />
+        </IconButton>
```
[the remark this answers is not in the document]
vallq: Oh that's what the tutorial there is for? I wasn't sure if it was something else! But sure, that would be better I think, rather than a stand-alone icon on the appbar.

src/components/tests/appbar.test.tsx
vallq: Oh okay let me test this out

src/components/tests/tour.test.tsx
vallq: Oh okay I'll change that

src/components/tests/tour.test.tsx
vallq: Ohh yes ok, that would shorten the file significantly

vallq:
Strange! That was my initial test as well but it did not work on my machine subsequently. Let me try this again *fingers crossed*
Thanks for the comments @matiasbavera !

src/components/tests/tour.test.tsx
vallq: Using the JSON format causes the snapshot file to be really long though, because I can't use a shallow render and perform the steps of finding the elements to click etc. there will be a lot of material ui html generated too
const root = shallow(<DashboardTour.../>) will return the Tour component only without any child elements in the render tree
this is what a partial snapshot of the shallow render looks like
shallow render tour

matiasbavera: Mmm.. one last try, can you check if there is a way to make something like this?  `expect(root.find('.reactour__helper')).toMatchSnapshot();`

To snap just a smaller part

matiasbavera:
> Strange! That was my initial test as well but it did not work on my machine subsequently. It's not even finding a node with `find(#tour-last-step-btn)` Let me try this again _fingers crossed_

Did it work?

vallq:
> > Strange! That was my initial test as well but it did not work on my machine subsequently. It's not even finding a node with `find(#tour-last-step-btn)` Let me try this again _fingers crossed_
> 
> Did it work?

It worked but I'm not sure it's supposed to be that way
because the last button should appear in the same step as the traj anim which is the 12th step and takes 11 clicks to reach so I'm not sure why it has to be `i < 13` instead of `i < 12`
if you refer to commit [ee43004](https://github.com/osrf/romi-dashboard/pull/97/commits/ee43004895494af5a6775ac5b9011f61fb59a095) and scroll all the way down to the final test, this was initially passing.

Regarding updating the snapshots to `root` instead of `root.html`, I tried to cut this shorter by finding the div that contains the step. It still gives a really long snapshot file though, because of the button components and the forward refs involved. Another way is to only find the Typography text component, that will give a super short snapshot of just the <h6></h6> mark up.

matiasbavera:
> It worked but I'm not sure it's supposed to be that way
> because the last button should appear in the same step as the traj anim which is the 12th step and takes 11 clicks to reach so I'm not sure why it has to be `i < 13` instead of `i < 12`

Yeah,  when I print the HTML generated by 'I<12'  it was the message from the 11/12 step. That's why I increment it to 13

vallq:
Got it, I've reverted the tests to check `root` for the snapshots! I don't like that it's so long as well haha. but alright, this is ready to merge then? @matiasbavera @koonpeng

src/components/tour/tour-navigation-control.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+export const NavButtons = React.memo((props: NavButtonProps) => {
+  const { goTo, step, handleNextClick, handleBackClick, lastStep } = props;
+  const classes = useStyles();
+  return (
+    <Box>
+      {step > 1 && (
+        <IconButton
+          onClick={() => {
+            if (handleBackClick) {
```
matiasbavera, line 32: Maybe this, to save a line =)
```
handleBackClick && handleBackClick()
goTo(step - 2);
```

matiasbavera: Ditto for the others below.

src/components/tour/tour-manager.tsx
```diff
@@ -0,0 +1,224 @@
...
+            lastStep={true}
+          />
+        </Box>
+      ),
+    },
+  };
+
+  const tourSteps: ReactourStep[] = [];
+
+  for (var key in tourContent) {
```
matiasbavera, line 214: maybe `let`?

matiasbavera (commented):
Thanks for the changes Val, I left a couple of comments!

matiasbavera:
Hi @koonpeng checking with you in this situation, we have a huge snapshot file ([22k lines](https://github.com/osrf/romi-dashboard/blob/3c97ccf560369da0b4bcd80ee55cdace54d902d3/src/components/tests/__snapshots__/tour.test.tsx.snap) )  that contains the snapshots of each step of the tutorial. Val tried to take a smaller snapshot finding inside elements, shallow mounting with the dive method, etc. and the only thing that she can refer (or find) was the tour step Typography, but if we are just snapping the typography we shouldn't be using snapshots at all, so we leave the entire snapshot. 

So the question here is the following, Do we want to keep and use these snapshots only as verification that the component did not change? Of course with the downside of having added 22K (+ all the other steps we will add) lines to the repository.

I think that since the snapshots are printing a lot of code and components related to the react-tour library, it may be possible that an update (of the react-tour library) could require us to update our snapshots.

koonpeng:
I had some problems with enzyme's selectors before as well, they seem to follow different rules then css selectors.

Some ideas
* Somehow skip enzyme and use other selectors?
* Take a snapshot of the props and states only, we can further reduce the footprint by targeting only the components which we are interested in for each step. I recall enzyme can obtain the props and states of a component.
* hash the html and take a snapshot of it, downside is that it would be an opaque snapshot.

src/components/tour/tour-data.tsx
```diff
@@ -0,0 +1,34 @@
...
+export const tourText: { [step: string]: string } = {
+  step1Welcome: 'Welcome to RoMi dashboard',
+  step2Zoom:
+    'Click on the zoom buttons to change the view of the floor plan. Alternatively, the scroll button on your mouse would work too!',
+  step3Floorplan:
+    'Use the floor plan button to switch between available levels and enabling / disabling the view of different components.',
+  step4Leaflet:
+    'Clicking individual components like doors, robots, lifts on the map will open up its corresponding information tab in the omnipanel.',
+  step5Omnipanel:
+    'The Omnipanel Button shows the different panel options available in the dashboard. Clicking each item would list different information about it!',
```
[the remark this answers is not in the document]
vallq: Ok let me think of a better way to paraphrase! And yep my intention was for the following step to demonstrate that feature for clarity

src/components/tour/tour-data.tsx
```diff
@@ -0,0 +1,34 @@
...
+  step6MainMenu:
+    'Each Panel contains a list of the available items and their corresponding states.',
+  step7DoorPanel: 'Let us take a look into the 🚪 Doors Panel.',
+  step8DoorTab: 'Here is an example of what you will see when a door tab is expanded',
+  step9CommandsPanel:
+    'The Commands Panel allows you to send different types of requests that will be handled by RoMi.',
+  step10LoopRequest:
+    'An example is the Loop Request which can be iterated multiple times. RoMi will assign the most suitable robot to perform the task at the point of request.',
```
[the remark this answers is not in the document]
vallq: Ah okay.. maybe we could merge to reduce a step then, since there are more coming up

src/components/tour/tour-manager.tsx
vallq: oh wow I missed this haha

vallq:
Alright, let me check out extraction of props from enzyme and I can try with the snapshot tests with React testing library and see if it works better too

vallq:
ignoring e2e for now because of rmf bugs, but e2e tests have to be modified to accommodate Tour being rendered on start when merged to master (clicking the tour close button on start should allow the rest of the test to continue)

src/components/tests/tour.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+
+    stepText = root.find('[data-testid="stepBox"]').first().text();
+
+    expect(stepText.includes(expectedText));
+  });
+
+  it('renders floor plan button step', async () => {
+    expectedText = tourText.step3Floorplan;
+
+    for (let i = 1; i < 3; i++) {
```
matiasbavera, line 77: Sorry Val, this occurred to me now and I think it will be extremely useful for anyone who has to add more cards to the tutorial. Could it be that we could find a way not to deal with fixed numbers within the tests, because if, for example, I want to add a card in position 3 I will have to re-enumerate the tests> 4. E.g .:
10 11 12 13 14
10 (new) 11 12 13 14 15.

So can we somehow get the positions from `tourSteps` or` tourContent`. And we do something like this

for (let i = 1; i < tourStepts.Floorplan; i++)

where tourStepts.Floorplan === 3.

The same thing would go with the variable names I think


What do you think about it?

matiasbavera (commented):
Thanks for the changes Val, left a couple of comments. Once those are addressed, it should be ready to go.

src/components/tests/tour.test.tsx
vallq: hm it's a bit tricky because how the tour does it is that the step number can only be accessed through the 'content' variable... let me think about it, I know where you're coming from and I think that would be mighty helpful in the future as well, so thanks for bringing that up too!!

src/resource-manager.ts
```diff
@@ -3,7 +3,7 @@ import { RobotResourceManager, RobotResource } from './resource-manager-robots';
 import { DispenserResourceManager, RawDispenserResource } from './resource-manager-dispensers';
 
 export interface ResourceConfigurationsType {
-  robots?: Record<string, RobotResource>; // Record<FleetName, RobotResource>
+  robots: Record<string, RobotResource>; // Record<FleetName, RobotResource>
```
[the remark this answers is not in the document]
vallq: oh but with the `?`, then the `commands-panel.test.tsx` will have a warning in line 21
resource manager

```
(alias) fakeResources(): ResourceConfigurationsType
import fakeResources
Argument of type 'ResourceConfigurationsType' is not assignable to parameter of type 'ResourceManagersProps'.
  Types of property 'robots' are incompatible.
    Type 'Record<string, RobotResource> | undefined' is not assignable to type 'Record<string, RobotResource>'.
      Type 'undefined' is not assignable to type 'Record<string, RobotResource>
```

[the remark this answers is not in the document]
matiasbavera: Mmm.. I'll look at that, maybe that's another rebase problem =)

matiasbavera (approved):
Thanks for the changes @vallq LGTM!

PR #98, opened 2020-09-03
pref: render optimizations

koonpeng:
aka the great memoization, the panel and schedule visualizer components are memoized to avoid rendering when nothing changed.

performance metrics:

before:
image

after:
image


src/util/react-helpers.ts
```diff
@@ -0,0 +1,60 @@
...
+ * @param callback
+ */
+export function makeCallbackArrayCallback<
+  U extends (...args: any[]) => any = (...args: any[]) => any,
+  T = any,
+  Args extends any[] = U extends (...args: infer Args) => any ? Args : never,
+  R = U extends (...args: any[]) => infer R ? R : never
+>(arr: T[], callback: (item: T, ...args: Args) => R): () => ((...args: Args) => R)[] {
+  return () => arr.map(item => (...args: Args) => callback(item, ...args));
+}
```
matiasbavera, line 60: Cool workaround. Thanks for the comments. Were you able to measure how this feature alone impacts in the Dashboard's performance?

src/components/schedule-visualizer/tests/__snapshots__/waypoint.test.tsx.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`Renders correctly 1`] = `
 <svg>
-  <Waypoint
+  <Memo()
```
matiasbavera, line 5: This looks weird.

Here those guys where getting `<Component/>` instead of their component. https://github.com/facebook/jest/issues/8122

matiasbavera (approved):
LGTM! se siente mucho más suave ahora

src/util/react-helpers.ts
koonpeng: I don't have the performance metrics, but this only affects the door and lifts panels as only those panel have non-changing list of items, it doesn't work for the other panels because the list is derived directly from the states, so changing the states will regenerate all the callbacks.

For those panels a different way (which might be better) is used, for example the robots panel, `data-fleet` and `data-name` is added to the elements so that we can retrieve the target robot based on html attributes.

src/components/schedule-visualizer/tests/__snapshots__/waypoint.test.tsx.snap
koonpeng: Yeah, seems like some bug either in enzyme or jest, using memo also sometimes causes enzyme to find a component twice so I had to switch to the css selector.

PR #101, opened 2020-09-08
perf: robot shadows

koonpeng:
This is an experimentation implementation of robot shadows using radial gradient. I tried to make the shadows look as close to the filter version as possible, in theory this should give better performance. Appreciate some inputs if this is preferred over the existing implementation.

### How to test

There is a new story to benchmark robot shadows, it can be found under Robot -> Benchmark. There is an input to set the number of robots and the shadow implementation. In my case, I got the follow results

* max robots while panning around
  * gradient: ~250
  * filter: ~100
* max robots while zoomed in and panning around
  * gradient: ~100
  * filter: ~20

The numbers are when the fps consistently drops below 60, there are still stutters that drops to ~30 fps occasionally for both gradient and filter perhaps from react or leaflet.

### Downsides

One downside of the gradient approach is it looks quite different for robot images, because the image does not follow a perfect circle, it is not possible to hack it to fit the shape of the image. This can be seen in the image icon in the robot gallery story.

The other small downside is that we can't use theme colors for conflicting highlight (technically we can but is not probably worth it), because it sets the alpha at each gradient "stops" using the hex notation, the theme color might be in rgb, named, or other notations. In order to use it we would have to convert all the possible color notations to hex and calculate the alpha to set for each stops.


ChawinTan:
> One downside of the gradient approach is it looks quite different for robot images, because the image does not follow a perfect circle, it is not possible to hack it to fit the shape of the image. This can be seen in the image icon in the robot gallery story.

I think the images look okay with a circle around it. At first I thought there would be white blotches that the shadow can't cover.

> The other small downside is that we can't use theme colors for conflicting highlight (technically we can but is not probably worth it), because it sets the alpha at each gradient "stops" using the hex notation, the theme color might be in rgb, named, or other notations. In order to use it we would have to convert all the possible color notations to hex and calculate the alpha to set for each stops.

Material ui theme uses hex colors too. The one that I am currently using to represent the trajectories `(secondary.main)` is #f50057, although the value is different when I check their website. The color values are also denoted in hex values on their website so I guess using theme colors should be fine.

matiasbavera:
> ### How to test
> There is a new story to benchmark robot shadows, it can be found under Robot -> Benchmark. There is an input to set the number of robots and the shadow implementation. In my case, I got the follow results
> 
> * max robots while panning around
>   
>   * gradient: ~250
>   * filter: ~100
> * max robots while zoomed in and panning around
>   
>   * gradient: ~100
>   * filter: ~20
> 
> The numbers are when the fps consistently drops below 60, there are still stutters that drops to ~30 fps occasionally for both gradient and filter perhaps from react or leaflet.

Yeah, I feel a huge performance improvement with the gradient on my PC.

> One downside of the gradient approach is it looks quite different for robot images, because the image does not follow a perfect circle, it is not possible to hack it to fit the shape of the image. This can be seen in the image icon in the robot gallery story.

Actually, I think that the images look okay with a circle around it too. Also, many games use this circular shadow to mark or highlight characters. So it doesn't seem like a bad approach to me at all. The other thing that can do, is not to put gradient shadows on the custom figures in their natural state, but put them and highlight them when there are conflicts
image



> Material ui theme uses hex colors too. The one that I am currently using to represent the trajectories `(secondary.main)` is #f50057, although the value is different when I check their website. The color values are also denoted in hex values on their website so I guess using theme colors should be fine.

Adding this https://material-ui.com/customization/palette/

But if I'm missing something here and we can't do that. We could define our hacky palette, for at least to have it all in one place.

```
export const hackyPalette = {
red = '# dc004e',
...
}
```

koonpeng:
I am not sure if material guarantees the colors to be in hex form and without alpha channel, that is the case for the default palette but I don't know of you can use other notations for custom palettes. I guess most of the custom themes would be using hex notations as well so it should work most of the time.

I think this is a very minor though, for now I just added the conflict color to `ColorManager`, it's using a fixed value now. I think it should be fine, we don't really need to follow the theme palette for things like this. Let me know if you disagree, it's just a small change to use the theme colors.

src/components/schedule-visualizer/tests/robot-image-icon.test.tsx
```diff
@@ -0,0 +1,45 @@
...
+        <RobotImageIcon
+          robot={makeRobot()}
+          footprint={1}
+          colorManager={colorManager}
+          dispatchIconError={jest.fn()}
+          iconPath={iconPath}
+        />
+      </svg>,
+    );
+    expect(root).toMatchSnapshot();
```
matiasbavera, line 27: -> `expect(toJson(root)).toMatchSnapshot();` you need to convert the container to Json with `enzyme-to-json` library first, an then compare with the snapshot. If you don't make that conversion the snapshot will be an empty object. You can check that here `src/components/schedule-visualizer/tests/__snapshots__/robot-default-icon.test.tsx.snap`

exports[`RobotDefaultIcon renders correctly 1`] = `ReactWrapper {}`;

src/components/schedule-visualizer/tests/robot-default-icon.test.tsx
```diff
@@ -0,0 +1,35 @@
...
+  colorManager.robotColor = jest.fn(async () => 'black');
+  colorManager.robotColorFromCache = jest.fn(() => 'black');
+
+  test('renders correctly', async () => {
+    const root = mount(
+      <svg>
+        <RobotDefaultIcon robot={makeRobot()} footprint={1} colorManager={colorManager} />
+      </svg>,
+    );
+    expect(root).toMatchSnapshot();
```
matiasbavera, line 19: -> `expect(toJson(root)).toMatchSnapshot();` you need to convert the container to Json with `enzyme-to-json` library first, an then compare with the snapshot. If you don't make that conversion the snapshot will be an empty object. You can check that here `src/components/schedule-visualizer/tests/__snapshots__/robot-default-icon.test.tsx.snap`

exports[`RobotDefaultIcon renders correctly 1`] = `ReactWrapper {}`;

src/components/schedule-visualizer/tests/__snapshots__/robot-image-icon.test.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`RobotImageIcon renders correctly 1`] = `ReactWrapper {}`;
```
matiasbavera, line 3: Explained in  comments below

src/components/schedule-visualizer/tests/__snapshots__/robot-default-icon.test.tsx.snap
```diff
@@ -0,0 +1,3 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`RobotDefaultIcon renders correctly 1`] = `ReactWrapper {}`;
```
matiasbavera, line 3: Explained in  comments below

src/components/schedule-visualizer/robot-image-icon.tsx
```diff
@@ -12,44 +12,74 @@ type RobotImageIconProps = Omit<RobotProps, 'colorManager' | 'fleetName'> & {
...
-            }}
+        <g>
+          <defs>
+            <Shadow id={shadowId} />
+            <ShadowConflict id={conflictShadowId} />
+          </defs>
+          <circle
+            id="shadow"
+            r={footprint * 1.3}
+            fill={inConflict ? `url(#${conflictShadowId})` : `url(#${shadowId})`}
```
matiasbavera, line 65: Just to open a discussion, do you think this could be a separate component that handles the unique id logic and receives a `makeGradientShadow` function, `colorManager`,  `inConflict`, and `footprint` because the same logic it's been used in `robot-image-icon` and `robot-default-icon`. 

Since this is a solution that improves the performance, we will tend to replicate this solution for components that may have shadows and icons (just to give an example: dispensers), and in my opinion, it would be nice to have this logic in a separate component.

matiasbavera (changes requested):
Awesome performance boost! Thanks @koonpeng ! Left a couple of comments.

src/components/schedule-visualizer/tests/robot-image-icon.test.tsx
koonpeng: You are right, totally missed that. There is also `root.html()` to get the rendered output, should I use that or convert it to json format?

src/components/schedule-visualizer/robot-image-icon.tsx
koonpeng: The main problem is that the shadow is different for each of them. `makeGradientShadow` itself is a higher order component, I thought it would be more "react-ful" to let the parent render the component instead of passing in a HOC as a prop.

A problem with wrapping this logic in the component is that it can't be reuse for waypoints/dispenser anyway as they do not have `colorManager`, `inConflict`, `footprint` etc.

matiasbavera: > The main problem is that the shadow is different for each of them. `makeGradientShadow` itself is a higher order component, I thought it would be more "react-ful" to let the parent render the component instead of passing in a HOC as a prop.

Yeah, you're right, it's not the best idea to pass the HOC as a prop, we can keep it that way. 

> A problem with wrapping this logic in the component is that it can't be reuse for waypoints/dispenser anyway as they do not have `colorManager`, `inConflict`, `footprint` etc.

Yeah, sorry I think I expressed myself wrong, I was aware of this. But don't worry this stops making sense after your first comment.

src/components/schedule-visualizer/tests/robot-image-icon.test.tsx
matiasbavera: Yeah, this is what each command generates. My personal opinion is that it's easier to read what the component is rendering by converting it to JSON.

### root.html():

```
exports[`RobotImageIcon renders correctly 1`] = `"<svg><g><defs><radialGradient id=\\"RobotImageIcon-0-shadow\\"><stop offset=\\"0%\\" stop-color=\\"#00000080\\"></stop><stop offset=\\"70%\\" stop-color=\\"#00000040\\"></stop><stop offset=\\"90%\\" stop-color=\\"#00000010\\"></stop><stop offset=\\"100%\\" stop-color=\\"#00000000\\"></stop></radialGradient><radialGradient id=\\"RobotImageIcon-0-shadow-conflict\\"><stop offset=\\"0%\\" stop-color=\\"#f4433680\\"></stop><stop offset=\\"70%\\" stop-color=\\"#f4433640\\"></stop><stop offset=\\"90%\\" stop-color=\\"#f4433610\\"></stop><stop offset=\\"100%\\" stop-color=\\"#f4433600\\"></stop></radialGradient></defs><circle id=\\"shadow\\" r=\\"1.3\\" fill=\\"url(#RobotImageIcon-0-shadow)\\"></circle><g transform=\\"translate(-1 -1)\\"><image href=\\"testIcon\\" height=\\"2\\" width=\\"2\\"></image></g></g></svg>"`;
```


### toJson(root):

```
exports[`RobotImageIcon renders correctly 1`] = `
<svg>
  <ForwardRef
    colorManager={
      ColorManager {
        "_robotColorCache": Object {},
        "conflictHighlight": "#f44336",
        "robotColor": [MockFunction],
        "robotColorFromCache": [MockFunction],
      }
    }
    dispatchIconError={[MockFunction]}
    footprint={1}
    iconPath="testIcon"
    robot={
      Object {
        "battery_percent": 100,
        "location": Object {
          "level_name": "",
          "t": Object {
            "nanosec": 0,
            "sec": 0,
          },
          "x": 0,
          "y": 0,
          "yaw": 0,
        },
        "mode": Object {
          "mode": 0,
        },
        "model": "BaseModel",
        "name": "",
        "path": Array [],
        "task_id": "",
      }
    }
  >
    <g>
      <defs>
        <Component
          id="RobotImageIcon-0-shadow"
        >
          <radialGradient
            id="RobotImageIcon-0-shadow"
          >
            <stop
              offset="0%"
              stopColor="#00000080"
            />
            <stop
              offset="70%"
              stopColor="#00000040"
            />
            <stop
              offset="90%"
              stopColor="#00000010"
            />
            <stop
              offset="100%"
              stopColor="#00000000"
            />
          </radialGradient>
        </Component>
        <Component
          id="RobotImageIcon-0-shadow-conflict"
        >
          <radialGradient
            id="RobotImageIcon-0-shadow-conflict"
          >
            <stop
              offset="0%"
              stopColor="#f4433680"
            />
            <stop
              offset="70%"
              stopColor="#f4433640"
            />
            <stop
              offset="90%"
              stopColor="#f4433610"
            />
            <stop
              offset="100%"
              stopColor="#f4433600"
            />
          </radialGradient>
        </Component>
      </defs>
      <circle
        fill="url(#RobotImageIcon-0-shadow)"
        id="shadow"
        r={1.3}
      />
      <g
        transform="translate(-1 -1)"
      >
        <image
          height={2}
          href="testIcon"
          onError={[Function]}
          width={2}
        />
      </g>
    </g>
  </ForwardRef>
</svg>
`;


```

matiasbavera (approved):
Thanks for this feature and the changes! LGTM!

PR #107, opened 2020-09-10
build: update npm scripts

koonpeng:
source the e2e env for start:react and start:docker as well, the env script now uses the existing env values if already defined and uses the default if not.

@ddengster does this work for your use case? You can launch only the react dev server with
`npm run start:react`
it should now build the app based on the default e2e env.

to launch the auth server only, use
`npm run start:auth`

to launch both at once, use
`npx concurrently npm:start:react npm:start:auth`
or you can just run the above 2 commands in different terminals.

this is an alternative to the proposed `start:noBackend` script in the negotiation PR.


ddengster:
i'd still prefer to have a shorter script to run considering its `npm run start:react-auth` now compared to `npx concurrently npm:start:react npm:start:auth`

PR #108, opened 2020-09-11
ci: run e2e natively

koonpeng:
Currently, the CI is ran by launch rmf services in a docker container and wdio/chrome on the host. The creates a problem that makes it hard for the CI to depend on rmf_demos for resource pack that is mandatory for dispensers to work. It may also makes it less stable as the why the tests are ran is different from what developers would use.

This change runs the whole CI in docker "natively", an e2e image is built nightly, this repo is then mounted onto that image and the tests are ran as if all the services are on the local host, hopefully this will make it easier to import resource pack when building the CI.

Relevant: #89

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,35 +1,35 @@
 name: Nightly
-on: push
-  # schedule:
-  #   # 2am SGT
-  #   - cron: '0 18 * * *'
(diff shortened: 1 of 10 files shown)


matiasbavera (approved):
Thanks for this, I'll be really helpful!

PR #113, opened 2020-09-16
Feature/settings drawer back button

ChawinTan:
<b>Button on the left</b>
left_button

<b>Button on the right</b>
right_button

Which layout do you prefer? Button on the left or right of the drawer?


matiasbavera:
> Which layout do you prefer? Button on the left or right of the drawer?

I would pick the right one. But still, this design does not seem harmonious to me. I think that if we are going to put the X there, we should center our content.

How many more options like these would we have in the future? I am asking this because if it is going to increase the number it might be interesting to move to accordions. Like this one:
image



Another option could be to have two buttons `exit` and `save changes`.  And we can actually save the preference of the user on the local storage since we don't have a database.

koonpeng:
> Button on the left
> left_button
> 
> Button on the right
> right_button
> 
> Which layout do you prefer? Button on the left or right of the drawer?

ahh the good old mac vs windows debate, as a windows user, i prefer it on the right.

ChawinTan:
> Another option could be to have two buttons `exit` and `save changes`. And we can actually save the preference of the user on the local storage since we don't have a database.

Do we need the save changes button?  I think the settings are automatically saved  in `onSettingsChange`, where `saveSettings` will save it in local storage.

koonpeng:
> > Another option could be to have two buttons `exit` and `save changes`. And we can actually save the preference of the user on the local storage since we don't have a database.
> 
> Do we need the save changes button? I think the settings are automatically saved in `onSettingsChange`, where `saveSettings` will save it in local storage.

Ya, i agree with auto save. It seems to be the trend for modern apps.

ChawinTan:
centered_items

Updated changes

matiasbavera:
> centered_items
> 
> Updated changes

What do you think about adding a centered title, and left the content to the left as it were (sorry for this =/ )?  Like:

```
          Settings             [X]
----------------------------------
Trajectory Animations
----------------------------------
  op1  op2
  op3  op4

```

With this config, we will not have an empty white space looking like it's missing something
image

ChawinTan:
with_heading

Updated with headings. Also included some unit tests for trajectory animations file so as to appease codecov complaining.

src/components/settings-drawer.tsx
```diff
@@ -9,18 +10,22 @@ import {
...
+  Grid,
 } from '@material-ui/core';
+import CloseIcon from '@material-ui/icons/Close';
 import React from 'react';
 import { Settings, TrajectoryAnimation, TrajectoryDiameter, TrajectoryColor } from '../settings';
 
 export interface SettingsDrawerProps extends DrawerProps {
   settings: Readonly<Settings>;
   onSettingsChange(settings: Settings): void;
+  handleCloseButton: Function;
```
matiasbavera, line 23: Maybe `React.Dispatch<React.SetStateAction<boolean>>`?

src/components/settings-drawer.tsx
```diff
@@ -136,17 +155,19 @@ const useStyles = makeStyles((theme) => ({
...
     },
   },
   trajGroup: {
     '@media (min-aspect-ratio: 8/10)': {
       flexDirection: 'row',
-      paddingLeft: theme.spacing(4),
+      paddingLeft: theme.spacing(8),
     },
     '@media (max-aspect-ratio: 8/10)': {
       flexDirection: 'row',
```
matiasbavera, line 173: Can you add more vertical space between the trajectories group? I think they are too close to each other

image

matiasbavera (commented):
Hi @ChawinTan it looks better, I know this is WIP but I left a few comments!

ChawinTan:
more_spacing

added more spacing between traj groups.

matiasbavera:
> more_spacing
> 
> added more spacing between traj groups.

Thanks for the changes, it looks better!

src/components/schedule-visualizer/tests/robot-image-icon.test.tsx
```diff
@@ -45,4 +43,31 @@ describe('RobotImageIcon', () => {
...
+    );
+    await act(
+      () =>
+        new Promise<void>((resolve) => {
+          setImmediate(() => {
+            root.update();
+            resolve();
+          });
+        }),
+    );
```
matiasbavera, line 70: Just in case, did you tried with the "act" inside the mount?

https://stackoverflow.com/questions/58127167/why-do-i-need-to-wrap-react-enzyme-state-changes-in-act

src/components/schedule-visualizer/tests/trajectory-animations.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+      <OutlineAnim
+        trajectory={trajectoryValue}
+        conflicts={trajectoryConflict}
+        footprint={footprint}
+        colorManager={colorManager}
+      />,
+    );
+    wrapper.unmount();
+  });
+});
```
matiasbavera, line 76: Maybe I'm missing something, but you're no expecting anything here

matiasbavera (commented):
Hi @ChawinTan , I really appreciate your proactivity and these tests, but a suggestion for next time, the tests that are not related to your changes or to the PR could you do it in another branch? What happens is the following: unfortunately, I will not be able to merge your changes from the drawer to the master because I left you some comments about the dashboard and trajectories tests.

HTH,

ChawinTan:
> Hi @ChawinTan , I really appreciate your proactivity and these tests, but a suggestion for next time, the tests that are not related to your changes or to the PR could you do it in another branch? What happens is the following: unfortunately, I will not be able to merge your changes from the drawer to the master because I left you some comments about the dashboard and trajectories tests.
> 
> HTH,

Got it. So if codecov is failing due to files that are not related to my branch, it's okay to merge the branch with codecov failing and then resolving the errors in a separate branch?

Will move the tests to a separate branch in the meantime

matiasbavera:
> Got it. So if codecov is failing due to files that are not related to my branch, it's okay to merge the branch with codecov failing and then resolving the errors in a separate branch?
> 
> Will move the tests to a separate branch in the meantime since codecov is failing again

It seems to me that we can merge it if the new features that are in the PR has their corresponding tests. The reviewer is going to realize that. We will continue to have codebot errors due to the technical  debt that we are having (with test), little by little we will fix that.

matiasbavera (approved):
LGTM!

PR #114, opened 2020-09-16
Add hotkeys to dashboard

matiasbavera:
## What's new

Added hotkeys to access components of the Dasboard.

```
  OPEN_COMMANDS: 'shift+c',
  OPEN_ROBOTS: 'shift+r',
  OPEN_DOORS: 'shift+d',
  OPEN_DISPENSERS: 'shift+i',
  OPEN_LIFTS: 'shift+l',
  OPEN_SETTINGS: 'shift+s',
  OPEN_OMNIPANEL: 'shift+o',
```

https://www.npmjs.com/package/react-hotkeys


koonpeng (changes requested):
These are great! Couple of comments

open omnipanel doesn't work, it closes the panel when it's already opened, pressing it again doesn't open the panel.

the commands to open doors/lift etc panels doesn't work when omnipanel is closed, might be a good idea to automatically open it and switch to the relevant view.

A help button at the app bar to show the list of shortcuts would be great, along with a shortcut to open the shortcuts reference itself. In the future this help button could also be used to trigger the dashboard tour, so I recommend a button that shows a menu list.

matiasbavera:
> These are great! Couple of comments
> 
> open omnipanel doesn't work, it closes the panel when it's already opened, pressing it again doesn't open the panel.
> 
> the commands to open doors/lift etc panels doesn't work when omnipanel is closed, might be a good idea to automatically open it and switch to the relevant view.
> 
> A help button at the app bar to show the list of shortcuts would be great, along with a shortcut to open the shortcuts reference itself. In the future this help button could also be used to trigger the dashboard tour, so I recommend a button that shows a menu list.

Thanks for the feedback, changes in process!

matiasbavera:
> open omnipanel doesn't work, it closes the panel when it's already opened, pressing it again doesn't open the panel.

Good catch, it was due to the async call of the setState function.

matiasbavera:
Hi @koonpeng can I have feedback on this interface? 
feed

Still not for review though, I'll like to add tests for the new components

koonpeng (approved):
:+1: Ready to merge once the failing test is addressed.

PR #116, opened 2020-09-18
[ready] fixes/support for multilevel clinic demo

ddengster:
~~depends on https://github.com/osrf/rmf_demos/pull/136~~ merged!

~~waiting for #80~~ merged!

- crash fix
- negotiation trajectories now show up on multilevel

run `ros2 launch demos clinic.launch.xml` instead of the office demo

### DIFF
--- a/src/components/negotiations-panel.tsx
+++ b/src/components/negotiations-panel.tsx
@@ -270,7 +270,6 @@ export default function NegotiationsPanel(props: NegotiationsPanelProps): JSX.El
             response: 'negotiation_trajectory',
             values: []
            };
-        console.log(traj.map_name);
         negotiationTrajStore[traj.map_name].values.push(traj);
       });
     }
--- a/src/components/negotiations-panel.tsx
+++ b/src/components/negotiations-panel.tsx
@@ -58,8 +58,7 @@ export interface NegotiationsPanelProps {
   negotiationStatusUpdateTS: number; // used to trigger rerenders
 }
 
-
-export const NegotiationsPanel = React.memo((props: NegotiationsPanelProps) => {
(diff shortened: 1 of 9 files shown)


src/robot-trajectory-manager.ts
```diff
@@ -38,6 +38,7 @@ export interface Trajectory {
   segments: RawKnot[];
   robot_name: string;
   fleet_name: string;
+  map_name: string;
```
matiasbavera, line 41: Just to check is the `map_name` === Level name, right? 
Also if the `map_name` property is required you may want to add it to `trajectories.json` too. For the 
`npm run start:mock` to work

ddengster: add map_name to json 6910e26

src/components/negotiations-panel.tsx
```diff
@@ -58,8 +58,7 @@ export interface NegotiationsPanelProps {
   negotiationStatusUpdateTS: number; // used to trigger rerenders
 }
 
-
-export const NegotiationsPanel = React.memo((props: NegotiationsPanelProps) => {
+export default function NegotiationsPanel(props: NegotiationsPanelProps): JSX.Element {
```
matiasbavera, line 61: Maybe I'm missing something or maybe is a rebase mistake. But you're removing the memo here (which improves performance). 

You can check the difference on `src/components/tests/__snapshots__/negotiations-panel.test.tsx.snap` too

Related PR https://github.com/osrf/romi-dashboard/pull/112

matiasbavera (approved):
After that issue it's resolved, you should be able to merge it.

src/components/negotiations-panel.tsx
ddengster: nice catch cd53210

PR #118, opened 2020-09-22
test/Added tests for drawers and hotkeys

matiasbavera:
### what's new
* Added test for drawers and drawers-header
* Added test for hotkeys-dialog


koonpeng (approved):
:+1:

PR #119, opened 2020-09-23
ci: foxy

koonpeng:
This is based on `feat/api-gateway` as soss doesn't work in foxy.


koonpeng:
merged into #121 as they need each other.

PR #120, opened 2020-09-23
Fix/airpot demo

ChawinTan:
https://github.com/osrf/romi-dashboard/issues/115

Removed the auto zoom in feature that I was doing earlier since setting the `minZoom`  properly automatically brings the map to a fitting size.
`setInitialBounds` is required to be placed after setting min zoom values otherwise map won't automatically be resized.

<b>To do</b>
Find a better way to determine value for min zoom. 
Find correct event type instead of setting it to (any).

Currently, the width and respective default zoom for both maps are:
- office: 25.xxxx -> 5
- airpot: 282.xxxx -> 2.5

Given that the magnification is x2 for each level up, I think we can deduce that minzoom for 50 ~ 4, min zoom for 100 ~3 min zoom for 200-400 ~ 2 and so on and we can come up with a formula to calculate from here.

Will there be a situation where the height is > than the width?


ChawinTan:
It seems like there isn't really a need for a major refactor. The bounds for each level have already been calculated and stored in `mapFloorLayers` in the initial render. Subsequent changes to bounds can be handled in `handleBaseLayerChange`, through the event object. The event provides the name and we can use it to access the correct bounds from there.

The minZoom value is dynamically calculated from the width at the moment in `determineMinZoom`, not taking height into account. It seems to be giving pretty reasonable values after testing it with a range of width values from 10 - 1000.

Right now, the algorithm is dependent on a `baseWidth` and `baseZoom` as constants. I remember there is a discussion on giving a vendors a range of size to upload their assets? (icons, maps etc ...) so we might need to change it in future if the range veers too far from what I have tested.

matiasbavera:
> It seems like there isn't really a need for a major refactor. The bounds for each level have already been calculated and stored in `mapFloorLayers` in the initial render. Subsequent changes to bounds can be handled in `handleBaseLayerChange`, through the event object. The event provides the name and we can use it to access the correct bounds from there.
> 
> The minZoom value is dynamically calculated from the width at the moment in `determineMinZoom`, not taking height into account. It seems to be giving pretty reasonable values after testing it with a range of width values from 10 - 1000.
> 
> Right now, the algorithm is dependent on a `baseWidth` and `baseZoom` as constants. I remember there is a discussion on giving a vendors a range of size to upload their assets? (icons, maps etc ...) so we might need to change it in future if the range veers too far from what I have tested.

Thanks for this @ChawinTan  looks like a good solution for now. I tested a bit as well and it seems to be working as we want. 

> Will there be a situation where the map (or floor) height is > than the width?
 
It'll be good to know that. Let's ask @mxgrey

mxgrey:
> Will there be a situation where the map (or floor) height is > than the width?

This is definitely something that we should be able to support in the long-term, because we don't want to put a height vs width constraint on how floor layouts are defined. If it adds considerable effort, then we can open a ticket to deal with it later, but if it can be fixed now, then that would be preferable.

ChawinTan:
I just added the calculations for the situation when height > width. I have done some testing with my own maps as well and they seem to be working fine too, at least when the monitor is displayed horizontally. 

Nonetheless, I will still open up an issue for us to keep a lookout for a situation where the minzoom is not calculated properly should such a situation happen.

matiasbavera:
Can you test this branch with the clinic world too?
`run ros2 launch demos clinic.launch.xml` instead of the office demo

src/components/schedule-visualizer/index.tsx
```diff
@@ -214,9 +216,35 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
     return () => clearInterval(interval);
   }, [props.trajManager, curMapFloorLayer, trajAnimDuration]);
 
+  function determineMinZoom(width: number, height: number): number {
+    if (width >= height * -1) {
```
matiasbavera, line 220: It must be a reason for you to multiply it for` -1`, but is there a way where you can compare the absolute values?
Something like:
Math.abs(width)>=Math.abs(height)

matiasbavera (commented):
Thanks @ChawinTan left some comments.

ChawinTan:
Clinic world for all levels seem to be working fine. It should look better if my monitor is oriented vertically.
clinic_world

src/components/schedule-visualizer/index.tsx
```diff
@@ -214,9 +216,35 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
     return () => clearInterval(interval);
   }, [props.trajManager, curMapFloorLayer, trajAnimDuration]);
 
+  function determineMinZoom(width: number, height: number): number {
+    if (Math.abs(width) >= Math.abs(height)) {
+      // Right now baseWidth and baseZoom are set to 50 and 4
+      // We might need to change this in future if a range of
+      // values is given
+      const baseWidth = 50;
+      const baseZoom = 4;
```
koonpeng, line 225: Based on my research, zoom level 0 in leaflet means 1:1 scale of the image, i.e. a image of 1024x1024 would be shown as 1024x1024. Each zoom level means a x2 focus, so at zoom level 1 the image would be 2048x2048, and at -1 it would be 512x512 (but there appears to be some bugs in negative zooms and its not officially supported by leaflet).

Take the office demo for example, its width is roughly 25.8. On my desktop, the zoom leaflet calculates that will fit the map onto the viewport is 6 (which is 64x zoom resulting in ~1650 pixels). If I reduce the viewport (by resizing the browser), leaflet calculates a zoom level of 5, because zoom 6 would be too big and wouldnt fit onto the viewport.

As such, the "amount" of map that can be shown on screen depends on the viewport of the browser. So if the goal is to find a min zoom level that can fit the map, we need to take into account of the viewport size. I guess one way to do it is set a min zoom of 0 initially. When the map is loaded, leaflet will calculate a suitable zoom that will fit the map into the viewport, we then set that as the min zoom to prevent the user from zooming out too much.

@mxgrey What do you think of the proposal? I assume the initial problem is that some maps can't zoom out far enough to fit the viewport.

mxgrey: That sounds good to me :+1: 

The only question it raises is whether there's ever a reason a user might want some extra zoom out, but as long as we're adjusting the minimum zoom based on the viewport size, I can't think of any reason a user would want that.

koonpeng: Or maybe we could just set the min zoom to 0.

ChawinTan: > I guess one way to do it is set a min zoom of 0 initially.

Setting a minZoom of 0 sounds good. 

> When the map is loaded, leaflet will calculate a suitable zoom that will fit the map into the viewport, we then set that as the min zoom to prevent the user from zooming out too much.

I initially tried to use the `getZoom` method from leaflet. Unfortunately, it does not show the calculated zoom value for the new map when you are toggling between maps and gives the zoom value of the current map. This is why I resorted to the rather tedious calculations to approximate the minZoom value from width or height.

If I understand correctly, you want to somehow get the zoom value for the new map once the map changes and set it as the minZoom value without having to resort to the approximations. I think we can access it somewhere in the event. The drawback is we need to set the event type to `any`.

> To be more robust, we would also have to handle window resize events to recalculate the new min zoom.

Good Idea!

ChawinTan: > Or maybe we could just set the min zoom to 0.

:tada:

koonpeng: Yeah, it can be a ~bug~ feature. leaflet should default to a sensible zoom level, if the user choose to zoom more too much then it is the user's choice.

ChawinTan:
Came up with my first implementation on the window resize feature. Tested it in my environment and it seems to be working as I expect it to be and would be nice is other people can test it out in their environment as well.

Tried using [this](https://leafletjs.com/reference-1.7.1.html#map-invalidatesize) as it seems to be the most straightforward way but it does not work for manual resizing of the window. You would need to dynamically resize the map for it to work. 

Removed the calculations to determine the minzoom and set the default to 0.

matiasbavera:
> Tested it in my environment and it seems to be working as I expect it to be and would be nice is other people can test it out in their environment as well.

Hi @ChawinTan it's working correctly for me too!

src/components/schedule-visualizer/index.tsx
```diff
@@ -217,6 +229,50 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
...
+    const mapHeight = document.getElementsByClassName('leaflet-image-layer')[0].clientHeight;
+    setMapDimensions({ width: mapWidth, height: mapHeight });
+    setBound(mapFloorLayers[e.name].bounds);
+  }
+
+  function mapZoomOut({ currZoom, minZoom, maxZoom, mapWidth, mapHeight }: MapResizeProps) {
+    mapRef.current?.leafletElement.setZoom(currZoom ? currZoom - 0.5 : 0);
+    mapRef.current?.leafletElement.setMinZoom(minZoom ? minZoom - 0.5 : 0);
+    mapRef.current?.leafletElement.setMaxZoom(maxZoom ? maxZoom - 0.5 : 8);
+    setMapDimensions({ width: mapWidth / 2 ** 0.5, height: mapHeight / 2 ** 0.5 });
```
koonpeng, line 242: Can you add a comment pointing to the formula used in leaflet?

src/components/schedule-visualizer/index.tsx
```diff
@@ -217,6 +229,50 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
   function handleBaseLayerChange(e: L.LayersControlEvent): void {
     debug('set current level name');
     setCurLevelName(e.name);
+    const mapWidth = document.getElementsByClassName('leaflet-image-layer')[0].clientWidth;
+    const mapHeight = document.getElementsByClassName('leaflet-image-layer')[0].clientHeight;
+    setMapDimensions({ width: mapWidth, height: mapHeight });
```
koonpeng, line 234: I think this will return the viewport of the image layer, and not the map dimensions.

src/components/schedule-visualizer/index.tsx
```diff
@@ -217,6 +229,50 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
...
+
+  function mapZoomIn({ currZoom, minZoom, maxZoom, mapWidth, mapHeight }: MapResizeProps) {
+    mapRef.current?.leafletElement.setZoom(currZoom ? currZoom + 0.5 : 0);
+    mapRef.current?.leafletElement.setMinZoom(minZoom ? minZoom + 0.5 : 0);
+    mapRef.current?.leafletElement.setMaxZoom(maxZoom ? maxZoom + 0.5 : 8);
+    setMapDimensions({ width: mapWidth * 2 ** 0.5, height: mapHeight * 2 ** 0.5 });
+  }
+
+  function handleResize() {
+    const documentWidth = document.documentElement.clientWidth;
```
koonpeng, line 253: This returns the viewport of the document, not the leaflet image layer.

src/components/schedule-visualizer/index.tsx
```diff
@@ -217,6 +229,50 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
...
+    const documentHeight = document.documentElement.clientHeight;
+    const mapWidth = mapDimensions.width;
+    const mapHeight = mapDimensions.height;
+    const currZoom = mapRef.current?.leafletElement.getZoom();
+    const minZoom = mapRef.current?.leafletElement.getMinZoom();
+    const maxZoom = mapRef.current?.leafletElement.getMaxZoom();
+    if (documentHeight < mapHeight && documentWidth === documentDimensions.width) {
```
koonpeng, line 260: Can you explain the logic here, why do we need to check `documentWidth`?

src/components/schedule-visualizer/index.tsx
```diff
@@ -217,6 +229,50 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
...
+      mapZoomOut({ currZoom, minZoom, maxZoom, mapWidth, mapHeight });
+    } else if (
+      documentHeight > mapHeight * 2 ** 0.5 &&
```
koonpeng, line 263: Why `mapHeight * 2 ** 0.5`?

src/components/schedule-visualizer/index.tsx
```diff
@@ -217,6 +229,50 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
   function handleBaseLayerChange(e: L.LayersControlEvent): void {
     debug('set current level name');
     setCurLevelName(e.name);
+    const mapWidth = document.getElementsByClassName('leaflet-image-layer')[0].clientWidth;
+    const mapHeight = document.getElementsByClassName('leaflet-image-layer')[0].clientHeight;
+    setMapDimensions({ width: mapWidth, height: mapHeight });
+    setBound(mapFloorLayers[e.name].bounds);
```
koonpeng, line 235: A major downside of this is that the zoom in reset each time the floor is changed.

koonpeng (changes requested):
[2020-10-06 10-40-03.zip](https://github.com/osrf/romi-dashboard/files/5331109/2020-10-06.10-40-03.zip)

Sometimes it behaves weirdly when resizing.

Maybe I'm missing something here, if we are using min zoom of 0, why do we still need to change the zoom on resize? Handling resize events should only be needed when we are calculating the min zoom ourselves, because the min zoom will depends on the viewport size, so the min zoom we calculate initially may not match after the user resize the viewport. If we are using a fixed min zoom of 0 then it shouldn't matter if the user do any resizing.

ChawinTan:
> Maybe I'm missing something here, if we are using min zoom of 0, why do we still need to change the zoom on resize? Handling resize events should only be needed when we are calculating the min zoom ourselves, because the min zoom will depends on the viewport size, so the min zoom we calculate initially may not match after the user resize the viewport. If we are using a fixed min zoom of 0 then it shouldn't matter if the user do any resizing.

I see where you are coming from now. I initially thought it was intended for it to be a feature such that when the user resize the browser, map will fit accordingly.

ChawinTan:
Anyway, here is the explanation to what the formulas were doing:

Every step up/down in our map increases/decreases the map size by a factor of 0.5, so this means the map will increase by (square root of 2) or decrease by (1/square root 2) every time the user toggles the button or move the scroll on his mouse.

To deal with size changes, I needed to compare if the height or width of the viewport is smaller than or greater than the image displayed.

Determining when to zoom out is quite straight forward. If the documentWidth for documentHeight is smaller than the image width or height, then set the zoom -0.5 smaller.

When trying to determine when to zoom out, I set it to be at the point where the documentHeight or documentWidth is bigger then the existing image width or height by a factor of (square root of 2) so that the image fills the viewport nicely when expanded.

We need to check with the documentWith and documentHeight to make sure that the conditions don't conflict with each other  when checking documentWidth against imageWidth and documentHeight against imageHeight.

koonpeng (approved):
thanks

PR #127, opened 2020-09-29
refactor: monorepo

koonpeng:
### What's New
This forms the basic skeleton of the monorepo, it currently just uses some simple pre-install scripts to link things together. I'm conflicted whether we should use lerna, I tried it out abit, there are many nice features but some disadvantages as well.

- lerna is a separate tool from npm, meaning we have to run `lerna bootstrap` on the root package to initialize the repo correctly. Failing to do so may cause the repo to be in unsupported state. E.g., the root package installs husky, if one just perform the usual `npm install` on the dashboard package, husky wont be installed and pre-commit hooks will not be ran.
- lerna doesn't perform symlink of `file://` packages. So we will have to "release" each of the packages, again, failing to do so will cause a bare `npm install` to fail as it can't find the dependencies. We don't actually have to publish the packages if we are using lerna to install the dependencies, as it will resolve them and symlink them locally.

In short, using lerna means we will have change the way we install dependencies. lerna is probably going to be very useful if we want to publish the packages but atm I'm not really sure if that is our goal. This repo is not really a library that can be published and reuse by other packages (although it could be with some work, so maybe using lerna could save us some potential tech debt?). Most likely the api-server will be "released" as a docker image, the dashboard would probably just be a tarball for nginx or some web servers.

### Further Work
Now that the api-server and dashboard is in the same repo, there a couple of things we can do to improve the workflow.
- Remove dependency on rclnodejs (and rmf) for the dashboard. This can be done several ways, one way is to
  - move e2e into a sub package under dashboard, the e2e devDependencies will be moved over there (this includes the api-server)
  - the dashboard still needs typing information from api-server, so a solution is to move the ros2 plugin "out-of-tree", so the dashboard will depend on `api-server` and dashboard/e2e will depend on `api-server-plugin-ros2`. This should decouple rclnodejs from the dashboard.
  - The api server is designed with plugins in mind, but some work still needs to be done for out-of-tree plugins to work nicely.


packages/dashboard/e2e/package.json
```diff
@@ -0,0 +1,29 @@
+{
```
matiasbavera, line 1: +1 to this since this will contain tests from both parties

packages/dashboard/e2e/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "author": "",
+  "license": "Apache-2.0",
+  "dependencies": {
+    "@types/mocha": "^8.0.3",
+    "@wdio/cli": "^6.5.2",
+    "@wdio/local-runner": "^6.5.2",
+    "@wdio/mocha-framework": "^6.5.0",
+    "@wdio/spec-reporter": "^6.4.7",
+    "@wdio/sync": "^6.5.0",
+    "chromedriver": "^85.0.1",
```
matiasbavera, line 25: I think we can remove the `chromedriver`  from `dashboard/package.json`

packages/api-server/src/websocket-connect.ts
```diff
@@ -0,0 +1,35 @@
...
+  socket: WebSocket,
+  data: WebSocket.Data,
+  next: () => void,
+) => void;
+
+export default class WebSocketConnect {
+  stack: WebSocketMiddleware[] = [];
+
+  constructor(public server: WebSocket.Server) {
+    server.on('connection', (socket, req) => {
```
matiasbavera, line 14: Should we set explicitly `any` here?

packages/api-server/src/websocket-connect.ts
```diff
@@ -0,0 +1,35 @@
...
+      const logger = baseLogger.child({ tag: req.socket.remoteAddress });
+      logger.info('connection established');
+
+      socket.on('message', async (data) => {
+        try {
+          for (let mdw of this.stack) {
```
matiasbavera, line 20: same here comment about any

packages/api-server/src/tests/websocket-connect.test.ts
```diff
@@ -0,0 +1,46 @@
+import WebSocket from 'ws';
```
matiasbavera, line 1: Should we add a connection failure test?

packages/api-server/src/authenticator.ts
```diff
@@ -0,0 +1,34 @@
...
+import * as jwt from 'jsonwebtoken';
+import WebSocket from 'ws';
+import { WebSocketMiddleware } from './websocket-connect';
+
+export interface Options {
+  type: 'secret' | 'publicKey';
+  secretOrPublicKey: string;
+}
+
+export default function auhthenticator(options: Options): WebSocketMiddleware {
```
matiasbavera, line 11: Typo auhthenticator

packages/api-server/src/tests/authenticator.test.ts
```diff
@@ -0,0 +1,34 @@
+import * as jwt from 'jsonwebtoken';
+import WebSocket from 'ws';
+import auhthenticator from '../authenticator';
```
matiasbavera, line 3: auhthenticator

packages/api-server/src/tests/authenticator.test.ts
```diff
@@ -0,0 +1,34 @@
...
+import WebSocketConnect, { WebSocketMiddleware } from '../websocket-connect';
+
```
matiasbavera, line 5: Should we add tests expecting errors on the jwt verification and on the token type checking?

packages/api-server/src/api-gateway.ts
```diff
@@ -0,0 +1,118 @@
...
+        req.id !== null &&
+        socket.send(msgpack.encode(buildResponse({ result: data, more: true }))),
+      end: (data) =>
+        req.id !== undefined &&
+        req.id !== null &&
+        socket.send(msgpack.encode(buildResponse({ result: data }))),
+      error: (error) =>
+        req.id !== undefined &&
+        req.id !== null &&
+        socket.send(msgpack.encode(buildResponse({ error }))),
```
matiasbavera, line 87: What do you think about creating a method for these restrictions and encodings?

packages/api-server/src/api-gateway.ts
```diff
@@ -0,0 +1,118 @@
...
+      const handlerRet = await handler(req.params, sender);
+      if (handlerRet === undefined) {
+        if (this._rpcHandlers[req.method].length === 1) {
+          sender.end(null);
+        }
+      } else {
+        if (this._rpcHandlers[req.method].length === 1) {
+          sender.end(handlerRet);
+        } else {
+          sender.send(handlerRet);
```
matiasbavera, line 110: Is it ok for `handlerRet === null ` to enter here and then go to the catch?

matiasbavera (commented):
Thanks for this, left a few comments, HTH

packages/api-server/src/index.ts
```diff
@@ -0,0 +1,123 @@
...
+    } else {
+      logger.info('serving through http');
+      return http.createServer();
+    }
+  })();
+  const wsServer = new WebSocket.Server({ server });
+
+  const app = new WebSocketConnect(wsServer);
+  if (config.secret) {
+    app.use(authenticator({ type: 'secret', secretOrPublicKey: config.secret }));
```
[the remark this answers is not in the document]
koonpeng: It makes sense when we have more middlewares, but atm I think it should be ok to leave it as is since we only have one middleware.

We could improve it more by having middlewares as "plugins" but there isn't a need for that currently and I would prefer to not overengineer things.

scripts/install.js
```diff
@@ -0,0 +1,26 @@
+const { execSync } = require('child_process');
+
+const installCmd = process.env.npm_config_refer === 'ci' ? 'npm ci' : 'npm install';
```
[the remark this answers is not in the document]
koonpeng: Yeah, pretty much. Currently the flow is like

dashboard-e2e -> dashboard -> api-server

They all depend on the "root" package as well, which installs the git hooks and prettier etc. The requirement for ros2/rmf is kind of implied by the requirement on `rmf_core` and `traffic_editor`. When we move the ros2 plugin out of tree, it would be like

dashboard-e2e -> dashboard
dashboard-e2e -> api-server-plugin-ros2
dashboard -> api-server
api-server
api-server-plugin-ros2 -> api-server

So installing dashboard no longer needs ros2/rmf.

packages/api-server/src/authenticator.ts
koonpeng: thanks!

packages/api-server/src/tests/authenticator.test.ts
koonpeng: thanks for the find!

packages/api-server/src/tests/authenticator.test.ts
koonpeng: good point! I will add these.

packages/api-server/src/api-gateway.ts
koonpeng: That will be nice but probably not crucial yet. These undefined and null should always be there for a rpc request and never there for a rpc notify, I think a stronger type checking is probably better in the long term.

packages/api-server/src/api-gateway.ts
koonpeng: I think it should be ok, I added a comment to explain the expected behavior.

```
      /**
       * handler is "req -> resp" if it only has 1 argument, else it is a "stream" with messages
       * sent in chunks.
       *
       * "req -> resp" handler:
       *   * if it returns void (undefined), we send back "null" as the resp, because "result" is
       *     required on success.
       *   * if it returns any other value, send that as the result.
       * "stream" handler:
       *   * if it returns void (undefined), don't send anything back. The handler will take care
       *     of sending the result in chunks.
       *   * if it returns a value, send that as the result, but always set the "more" flag, the
       *     handler should take care of sending the rest of the chunks.
       */
```

matiasbavera: Awesome! Thanks!

packages/api-server/src/plugins/trajectory/tests/README.md
```diff
@@ -0,0 +1,5 @@
+Tests are indirectly done in https://github.com/osrf/romi-dashboard.
+
+It's hard to write tests for this, the most feasible approach is to mock the trajectory server but that is flaky at best and is not worth the effort.
+
+Instead, we should focus on integration/e2e tests, but doing so requires depdency on many components of rmf, the main challenge is that it is not easy to build and launch all those components, instead of replicating the work done by romi-dashboard, it is easier to delegate the tests there.
```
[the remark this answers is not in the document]
matiasbavera: I think you missed those comments that were collapsed =)

packages/api-server/src/plugins/trajectory/index.ts
```diff
@@ -0,0 +1,145 @@
...
+export interface Trajectory {
+  id: number;
+  shape: string;
+  dimensions: number;
+  segments: RawKnot[];
+  robot_name: string;
+  fleet_name: string;
+}
+
+export type Conflict = number[];
```
matiasbavera, line 52: Not for this PR but, should we use just these definitions in the `dashboard/robot-trajectory-manager.ts` too? to keep one source of truth

matiasbavera (approved):
Thanks for this! Left a couple of comments. I think we can add something like this (this is just a quick draft) when we'll add the readme file for the api-server 
Apigateway

matiasbavera:
I'm having errors building on my local env. I cannot load the map. I'm not getting errors on the backend though.
image

packages/api-server/src/tests/api-gateway.test.ts
```diff
@@ -0,0 +1,174 @@
...
+  app = new WebSocketConnect(server);
+  api = new ApiGateway();
+  app.use(api.middleware);
+});
+
+afterEach(() => {
+  server.close();
+});
+
+test('primitive request params', (done) => {
```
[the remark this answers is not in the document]
koonpeng: If you are using vscode, it appears it is not using the correct tsconfig, you might have to set up a multi-root workspace or open the packages/api-server folder directly.

packages/api-server/src/plugins/trajectory/index.ts
koonpeng: yea, these are copied from that file. When we change the trajectory client to use the api server we should have it reference the typings here.

koonpeng:
> I'm having errors building on my local env. I cannot load the map. I'm not getting errors on the backend though.
> image

maybe rclnodejs is not built with the rmf messages? I think these error messages are propagated from the backend.

matiasbavera:
Hi @koonpeng, I built the docker images locally with those commands

```bash
# build rmf image	
docker-compose -f <path-to-romi-dashboard>/packages/dashboard/docker/rmf/docker-compose.yml build --no-cache
# build e2e image
docker-compose -f <path-to-romi-dashboard>/packages/dashboard/docker/docker-compose.yml build --no-cache e2e
```
And I got these errors, sounds like it didn't copy the whole folder
image


```
dev_1        | [start:api] sh: 1: api-server: not found
dev_1        | [start:api] npm ERR! code ELIFECYCLE
dev_1        | [start:api] npm ERR! syscall spawn
dev_1        | [start:api] npm ERR! file sh
dev_1        | [start:api] npm ERR! errno ENOENT
dev_1        | [start:react] 
dev_1        | [start:react] > romi-dashboard@0.0.1-alpha.0 start:react /root/romi-dashboard
dev_1        | [start:react] > . e2e/env.sh && react-scripts start
dev_1        | [start:react] 
dev_1        | [start:api] npm ERR! romi-dashboard@0.0.1-alpha.0 start:api: `api-server --port 50002 --trajectoryServerUrl ws://localhost:8006 --publicKey e2e/certs/keycloak.pub`
dev_1        | [start:api] npm ERR! spawn ENOENT
dev_1        | [start:api] npm ERR! 
dev_1        | [start:api] npm ERR! Failed at the romi-dashboard@0.0.1-alpha.0 start:api script.
dev_1        | [start:api] npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
dev_1        | [start:api] 

```


```

dev_1        | [start:react] Failed to compile.
dev_1        | [start:react] 
dev_1        | [start:react] /root/romi-dashboard/src/api-client.ts
dev_1        | [start:react] TypeScript error in /root/romi-dashboard/src/api-client.ts(2,41):
dev_1        | [start:react] Cannot find module 'api-server/src/api-gateway' or its corresponding type declarations.  TS2307
dev_1        | [start:react] 
dev_1        | [start:react]     1 | import * as msgpack from '@msgpack/msgpack';
dev_1        | [start:react]   > 2 | import { RpcRequest, RpcResponse } from 'api-server/src/api-gateway';
dev_1        | [start:react]       |                                         ^
dev_1        | [start:react]     3 | import Debug from 'debug';
dev_1        | [start:react]     4 | 
dev_1        | [start:react]     5 | const debug = Debug('ApiClient');
```

matiasbavera:
> maybe rclnodejs is not built with the rmf messages? I think these error messages are propagated from the backend.

Yeah, that was it, I got an old version.

There is another problem, the setup script was affected somehow by the changes. Without the setup file we can't have the robots icons, the commands, or the dispensers.

image

I can give it a look tomorrow

HTH

koonpeng:
> And I got these errors, sounds like it didn't copy the whole folder
> image

Did you do npm install on the packages?

matiasbavera:
> > And I got these errors, sounds like it didn't copy the whole folder
> > image
> 
> Did you do npm install on the packages?

I found the problem, we just were using the dashboard directory and not the whole romi-dashboard directory so we should use this configuration in the `docker/docker-compose.yml` file when we try to run 
`npm run start:docker`

```

  dev:
    image: docker.pkg.github.com/osrf/romi-dashboard/e2e
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - ../../..:/root/romi-dashboard
    working_dir: /root/romi-dashboard/packages/dashboard
    ipc: host
    network_mode: host
    privileged: true
    command: bash -c 'BROWSER=none npm start'
```

The difference with the current one is the following:
```
- ../../..:/root/romi-dashboard
    working_dir: /root/romi-dashboard/packages/dashboard
```

matiasbavera:
> Yeah, that was it, I got an old version.
> 
> There is another problem, the setup script was affected somehow by the changes. Without the setup file we can't have the robots icons, the commands, or the dispensers.
> 
> I can give it a look tomorrow
> 
> HTH

Well, the setup.js script it's working it was an annoying typo. But in the docker env the dashboard now it's getting the resources from the repo directly instead of a folder inside the docker. Those changes went missing. I'll fix that when this is merged

matiasbavera (approved):
Thanks for the changes @koonpeng . Should be ready to merge when this https://github.com/osrf/romi-dashboard/pull/127#issuecomment-704320096 is addressed. 

It's a good time to merge it since we only have two PR open besides this.

CC: @ChawinTan this will affect your PR

koonpeng:
>I found the problem, we just were using the dashboard directory and not the whole romi-dashboard directory so we should use this configuration in the docker/docker-compose.yml file when we try to run
npm run start:docker

thanks for the investigation!

PR #129, opened 2020-09-30
Add an emergency alarm component to RomiDashaboard

matiasbavera:
## Pre-requisite

https://github.com/osrf/romi-js-core-interfaces/pull/18 (Done)

## What's new
* Added an emergency alarm component to the Dashboard (not implemented).
* Refactored alerts
* Added tests for alerts
* Added Logic subscribe to the emergency topic and update the state of the alarm.

### Context
while a std::msgs::Bool  message with value true is published over topic /fire_alarm_trigger, the emergency situation is active. During this situation, the robots will head to their nearest holding points on the map.


matiasbavera:
Hi @koonpeng I'm trying to publish to a specific topic `fire_alarm_trigger` that it does not have a custom message like `RomiCore.loopRequests` in the `romi-js-core-interfaces`. So i'm getting the type of the message from `rclnodejs/std_msgs/Bool`. I'm trying to use the transport (that we define in app-config) to publish the message. I did the following:

```
class Emergency {
  static readonly typeName = "std_msgs/Bool";
  static fromObject(obj: any): Emergency {
    return new Emergency(true);
  };
  constructor(public data: boolean){ };
}



export default function AppBar(props: AppBarProps): React.ReactElement {
  const { toggleShowOmniPanel, showSettings, showHelp, transport } = props;
  const [anchorEl, setAnchorEl] = React.useState<HTMLElement | null>(null);
  const classes = useStyles();
  const authenticator = React.useContext(AuthenticatorContext);
  const user = React.useContext(UserContext);
  const [alarm, setAlarm] = React.useState(false);

  const emergency : RomiCore.RomiTopic<Emergency> = {
    validate: msg => Emergency.fromObject(msg),
    type: Emergency.typeName,
    topic: 'fire_alarm_trigger',
  };

  const EmergencyAlarmRequestPub = transport && transport.createPublisher( emergency);
  

  const handleAlarm = (): void => {
    EmergencyAlarmRequestPub?.publish({
      data:true
    })

  };

```
I'm having this error right now
image


I would really appreciate any inside/guide that you can give me on this.

koonpeng:
rclnodejs requires the "full" type name, i.e. `std_msgs/msg/Bool`

packages/dashboard/src/components/appbar.tsx
```diff
@@ -7,25 +7,46 @@ import {
...
+  transport?: Readonly<RomiCore.Transport>;
+}
+
+class Emergency {
+  static readonly typeName = 'std_msgs/Bool';
+  static fromObject(obj: any): Emergency {
+    return new Emergency(true);
+  }
+  constructor(public data: boolean) {}
 }
```
matiasbavera, line 33: Hi @koonpeng  should we move this to `romi-js-core-interfaces`

packages/dashboard/src/components/appbar.tsx
```diff
@@ -7,25 +7,46 @@ import {
...
   const classes = useStyles();
   const authenticator = React.useContext(AuthenticatorContext);
   const user = React.useContext(UserContext);
+  const [alarm, setAlarm] = React.useState(false);
+
+  const emergency: RomiCore.RomiTopic<Emergency> = {
+    validate: (msg) => Emergency.fromObject(msg),
+    type: Emergency.typeName,
+    topic: 'fire_alarm_trigger',
+  };
```
matiasbavera, line 47: Should we move this too to romi-js-core-interfaces

matiasbavera: Related PR https://github.com/osrf/romi-js-core-interfaces/pull/18

packages/dashboard/src/components/appbar.tsx
matiasbavera: Something like this https://github.com/osrf/romi-js-core-interfaces/pull/18

matiasbavera: Yes we should, closing this

packages/dashboard/src/components/appbar.tsx
matiasbavera: Same here

matiasbavera:
If the browser crashes or is reloaded the publisher is destroyed but other active browsers keep the last state. If I enter to a new session when the publisher is already destroyedhue

matiasbavera:
E2e it's failing in the loop request, I'm taking a look into that
image

packages/dashboard/src/util/alerts.ts
```diff
@@ -1,20 +1,81 @@
...
+  cancelCallback?: () => void;
+}
+
+export class Alerts {
+  static success = (message?: string) => {
+    const swalParams = Alerts.getSuccessMsg(message);
+    return Swal.fire(swalParams);
+  };
+
+  static error = (message?: string) => {
```
koonpeng, line 22: This looks testable but is not tested according to codecov.

packages/dashboard/src/util/alerts.ts
```diff
@@ -1,20 +1,81 @@
...
+    const swalParams = Alerts.getErrorMsg(message);
+    return Swal.fire(swalParams);
+  };
+
+  /**
+   * Modal to warn a user about something related to an action.
+   */
+  static verification = (props: VerificationAlertProps) => {
```
koonpeng, line 30: This looks testable but is not tested according to codecov.

packages/dashboard/src/components/appbar.tsx
```diff
@@ -51,6 +60,12 @@ export default function AppBar(props: AppBarProps): React.ReactElement {
         </IconButton>
         {user && (
           <>
+            <EmergencyAlarm
+              isActive={alarmState}
+              onTurnOn={() => emergencyAlarmRequestPub?.publish({ data: true })}
```
koonpeng, line 65: To make this more testable and reduce coupling, use a callback instead of adding a dependency on `transport`.

packages/dashboard/src/components/emergency-alarm.tsx
```diff
@@ -0,0 +1,38 @@
...
+interface EmergencyAlarmProps {
+  onTurnOn?: () => void;
+  onTurnOff?: () => void;
+  isActive: boolean | null;
+}
+
+export const EmergencyAlarm = React.memo((props: EmergencyAlarmProps) => {
+  const { onTurnOn, onTurnOff, isActive } = props;
+
+  const handleAlarm = (): void => {
```
koonpeng, line 15: This looks testable but is not tested according to codecov.

packages/dashboard/src/components/emergency-alarm.tsx
```diff
@@ -0,0 +1,38 @@
...
+    if (isActive) {
+      Alerts.verification({
+        confirmCallback: () => {
+          onTurnOff && onTurnOff();
+        },
+        body: `You're about to turn off the alarm! The robots will resume their tasks.`,
+      });
+    } else {
+      Alerts.verification({
```
koonpeng, line 24: This looks testable but is not tested according to codecov.

koonpeng (changes requested):
Have you experimented with ros2 QoS settings to make alarms persist across sessions?

packages/dashboard/e2e/tests/emergency-alarm.test.ts
```diff
@@ -0,0 +1,71 @@
...
+    requestLoop('pantry', 'cubicle_1');
+    // wait for the robot to move
+    browser.pause(5000);
+
+    // activate Alarm
+    openAlarmModal();
+    confirmAction();
+
+    // time for the robot to go to a holding point
+    browser.pause(15000);
```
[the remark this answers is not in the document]
matiasbavera: Yeah, you're right, I'll add a TODO for this.

packages/dashboard/src/components/appbar.tsx
matiasbavera: Sounds good, thanks! https://github.com/osrf/romi-dashboard/pull/129/commits/771b4b4755692e3b0cad9b8ff1395f20a5afe7fa

packages/dashboard/src/util/alerts.ts
```diff
@@ -1,20 +1,81 @@
...
+    Swal.fire(verificationParams).then((modalState) => {
+      if (modalState.isConfirmed) {
+        !!confirmCallback && confirmCallback();
+      } else {
+        !!cancelCallback && cancelCallback();
+      }
+    });
+  };
+
+  static getSuccessMsg = (message?: string): SweetAlertOptions => {
```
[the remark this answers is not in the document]
matiasbavera: Makes sense! Thanks!

packages/dashboard/src/util/tests/alerts.test.ts
```diff
@@ -0,0 +1,83 @@
+import { createMuiTheme } from '@material-ui/core';
+import { Alerts } from '../alerts';
+
+const theme = createMuiTheme();
+
+describe('Verification Alert', () => {
+  test('Correct generation of default params values', () => {
+    const verification = Alerts.getVerificationParams({});
+    expect(verification).toEqual({
+      title: 'Are you sure you want to continue?',
```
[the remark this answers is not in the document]
matiasbavera: Done! Thanks!

packages/dashboard/src/util/alerts.ts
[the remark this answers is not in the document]
matiasbavera: > We should also convert this to a react component, but that is for another PR

Yeah, it's difficult to test the alerts from `sweetalert2`, since they are not rendered on the jsdom, so the functionality is been tested just on the e2e test. This can be a problem in the future so I'm thinking of replacing it with https://material-ui.com/components/dialogs/ or use this react wrapper and see if I can test it https://github.com/alex-shamshurin/sweetalert2-react.

matiasbavera:
> Have you experimented with ros2 QoS settings to make alarms persist across sessions?

Hi @koonpeng, yes but I had the same behavior, added again here https://github.com/osrf/romi-dashboard/pull/129/commits/530c36602e020b827d6c539d07cc045c4eae7346. Am I doing something wrong? 

```
startSubscription(transport: RomiCore.Transport) {
    this._subscriptions.push(
      transport.subscribe(
        RomiCore.emergency,
        (state) => {
          console.log(state);
          this._alarmState = state.data;
          this.emit('updated');
        },
        {
          qos: {
            historyPolicy: RomiCore.HistoryPolicy.KeepLast,
          },
        },
      ),
    );
  }
```

```
  const emergencyAlarmRequestPub = React.useMemo(
    () =>
      transport
        ? transport.createPublisher(RomiCore.emergency, {
            qos: {
              historyPolicy: RomiCore.HistoryPolicy.KeepLast,
            },
          })
        : null,
    [transport],
  );

```

koonpeng:
I'm not very familiar with ros2 qos as well, I will have to spend some time to experiment around. Hopefully we can fix the persistency issue before merging this.

koonpeng:
Looks like you need the equivalent of these settings

```
      qos: new rclnodejs.QoS(
        rclnodejs.QoS.HistoryPolicy.RMW_QOS_POLICY_HISTORY_KEEP_LAST,
        1,
        rclnodejs.QoS.ReliabilityPolicy.RMW_QOS_POLICY_RELIABILITY_RELIABLE,
        rclnodejs.QoS.DurabilityPolicy.RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL,
      ),
```
Using the above, I managed to publish first, then late subscriptions receives the last message.

We should also set that as the default qos for the emergency topic in `romi-js-core-interfaces`.

---

Test code with rclnodejs

```ts
import rclnodejs from 'rclnodejs';

async function main() {
  await rclnodejs.init();

  const nodeA = rclnodejs.createNode('nodeA');
  const pub = nodeA.createPublisher('std_msgs/msg/String', 'test', {
    qos: new rclnodejs.QoS(
      rclnodejs.QoS.HistoryPolicy.RMW_QOS_POLICY_HISTORY_KEEP_LAST,
      1,
      rclnodejs.QoS.ReliabilityPolicy.RMW_QOS_POLICY_RELIABILITY_RELIABLE,
      rclnodejs.QoS.DurabilityPolicy.RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL,
    ),
  });
  pub.publish({ data: 'hello' });
  rclnodejs.spin(nodeA);

  await new Promise((res) => setTimeout(res, 1000));
  const nodeB = rclnodejs.createNode('nodeB');
  nodeB.createSubscription(
    'std_msgs/msg/String',
    'test',
    {
      qos: new rclnodejs.QoS(
        rclnodejs.QoS.HistoryPolicy.RMW_QOS_POLICY_HISTORY_KEEP_LAST,
        1,
        rclnodejs.QoS.ReliabilityPolicy.RMW_QOS_POLICY_RELIABILITY_RELIABLE,
        rclnodejs.QoS.DurabilityPolicy.RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL,
      ),
    },
    console.log,
  );
  rclnodejs.spin(nodeB);
}

main();
```

matiasbavera:
> Looks like you need the equivalent of these settings
> 
> ```
>       qos: new rclnodejs.QoS(
>         rclnodejs.QoS.HistoryPolicy.RMW_QOS_POLICY_HISTORY_KEEP_LAST,
>         1,
>         rclnodejs.QoS.ReliabilityPolicy.RMW_QOS_POLICY_RELIABILITY_RELIABLE,
>         rclnodejs.QoS.DurabilityPolicy.RMW_QOS_POLICY_DURABILITY_TRANSIENT_LOCAL,
>       ),
> ```
> 
> Using the above, I managed to publish first, then late subscriptions receives the last message.


 
maybe I'm missing something but it's not getting the last message. 
not saving

But since the example that you post is working, I'll dig how the message reaches to the API gateway.

koonpeng:
Oh, I think it's probably because the api gateway shares the same internal subscription across multiple subscribers. This is because ros2 used to have scaling issues (maybe its been addressed now) with too many subscribers, if we spawn a new subscriber for each new connection we could be easily spawning hundreds or thousands of subscribers.

mxgrey:
I'm a bit confused on what the exact problem is, so I'll try to articulate my current understanding and you guys can let me know if I've misinterpreted things.

My understanding: It sounds to me like the goal is to subscribe to the alarm state topic on startup to see what the last alarm value was and then initialize the GUI to start with that value. This would allow the dashboard's alarm state to be persistent between dashboard sessions. But also, the dashboard is the only DDS participant that's publishing to the alarm topic.

As far as I can figure, there's no way for that setup to work. Once your dashboard session ends, its publisher will die with it. At that point, the messages which were stored by the transient local quality setting will effectively be erased because their publisher is gone. Any new subscribers who connect to that topic will be unable to discover the old publisher because it's dead.

From a Google search, I can see that Connext DDS provides a "persistence" service that stores messages to disk and will retransmit those messages after the publisher has died, but I think that might be a bonus feature of Connext and not something that is natively supported by ROS2.

I think a different strategy will be needed to get persistence for the dashboard.

koonpeng:
Sorry for the misunderstanding. The "dashboard" in this case consist of several components, namely the web app client and a backend ros2 bridge (among other backends like the auth server). The actual publishers and subscribers are maintained by the backend bridge, currently it only maintains one pub/sub for each topic and multiplexes the messages to each of the connected web app clients. We need to do this to avoid an explosion of pub/subs, each connection will need to create 5-10 pub/subs, using individual pub/subs could easily explode the ros2 discovery with hundreds or thousands of participants.

But because we are multiplexing the pub/subs, webapp clients will not receive previous messages from QoS "keep last" as there is only one internal ros2 subscription. We could emulate ros2 QoS settings in the bridge to let the clients behave as if they all have their own internal subscribers. Or maybe it is better to use a service to get the current state of emergency alarm some where in `rmf_core` so that the web app can display the correct emergency status at all times.

mxgrey:
Okay, thanks for the more detailed explanation.

I think the aspect of this that's most tricky is that the "source of truth" for the alarm is supposed to be a heartbeat from a building management system adapter. In an actual deployment, that message should be getting published periodically, and it should never emanate from the dashboard in the first place.

What if the alarm management is implemented in the web app backend, where its state could be `on`, `off`, or `null`, where `null` means that no message is being broadcast. Then instead of the web client publishing directly to a topic, it can switch the backend between those three states, where the default state is `null`?

koonpeng:
Thanks, afaik, there isn't a "building management system adapter" yet? If that is the case, I propose we remove this feature until it is ready, we can still merge in this PR, just remove the feature from the ui.

On another note, should the dashboard provide a way to clear the alarm? Or should write-level control only be available in the rviz ui?

mxgrey:
> there isn't a "building management system adapter" yet?

For the physical demos, there is a node responsible for publishing the alarm state. It's hooked up to a big red button. For simulation demos, we've just been using rviz. There was some confusion when setting up a demo one time where the big red button and the rviz panel were both publishing alarm states, and when those states disagreed with each other, the fleet adapters would end up rapidly flipping in and out of emergency mode.

It's not obvious to me what the "right" way to deal with this is. I'm not sure if real-life deployments are guaranteed to have a single source of truth or if we'll have to deal with multiple sources of truth for the alarm. If there will actually be multiple sources of truth, then it would be good to have a multiplexer node that publishes an on-state if any one source is on, and an off-state if all of the sources are off.

matiasbavera:
Alright, thank you both for the explanation. 

> For the physical demos, there is a node responsible for publishing the alarm state. It's hooked up to a big red button. For simulation demos, we've just been using rviz. There was some confusion when setting up a demo one time where the big red button and the rviz panel were both publishing alarm states, and when those states disagreed with each other, the fleet adapters would end up rapidly flipping in and out of emergency mode.

Ok, to be on the same page:
1. In the meanwhile, we can subscribe to the `/emergency_alarm` topic from the dashboard; the alarm icon will be displayed just when the alarm is turned on. 
2. For the publisher, we have a couple of options for now:
   a. If we want to use the Dashboard for the demo instead of rviz, but not in the actual deployment, we can keep the alarm button (on/off) and the `publisher` behind a flag. If we want to publish the message from the Dashboard we should do something like Michael mentioned: add a backend service that saves, change and broadcast states. Maybe we can add them of o our API gateway, and use it as our backend
   b. We can just remove the publisher from the dashboard, in this case, we will be depending on an external actor (someone pressing the big red button or rviz) to publish the message on the `/emergency_alarm` topic.
 
I'll do the 1 and 2.b. on this PR. 

> It's not obvious to me what the "right" way to deal with this is. I'm not sure if real-life deployments are guaranteed to have a single source of truth or if we'll have to deal with multiple sources of truth for the alarm. If there will actually be multiple sources of truth, then it would be good to have a multiplexer node that publishes an on-state if any one source is on, and an off-state if all of the sources are off.

Yeah, it makes total sense, not having a true scenario complicates things. Ideally, we should cover both scenarios, but I don't know if we should do it until we have more specifications. I'll create a ticket to deal with this again in the future.

packages/dashboard/src/components/confirm-alert-dialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+import { makeStyles } from '@material-ui/core';
+
+const Transition = React.forwardRef(function Transition(
+  props: TransitionProps & { children?: React.ReactElement<any, any> },
+  ref: React.Ref<unknown>,
+) {
+  return <Slide direction="up" ref={ref} {...props} />;
+});
+
+export interface ConfirmAlertDialogProps {
```
koonpeng, line 20: extend material ui dialog props and forward props into it?

packages/dashboard/src/components/confirm-alert-dialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+  open: boolean; // controls when the modal should open
+  close: () => void; // close the modal
```
koonpeng, line 22: The material ui pattern is to let the parent component handle closing the dialog from the confirm/cancel callback.

packages/dashboard/src/components/confirm-alert-dialog.tsx
```diff
@@ -0,0 +1,109 @@
...
+    close();
+  };
+  const classes = useStyles();
+
+  return (
+    <div>
+      <Dialog
+        open={open}
+        TransitionComponent={Transition}
+        keepMounted
```
koonpeng, line 64: We should let the parent component decides if it wants to keep it mounted.

packages/dashboard/src/components/dashboard.tsx
```diff
@@ -37,6 +37,7 @@ import { buildHotKeys } from '../hotkeys';
 import HelpDrawer from './drawers/help-drawer';
 import HotKeysDialog from './drawers/hotkeys-dialog';
 import { GlobalHotKeys } from 'react-hotkeys';
+import EmergencyManager from '../emergency-manager';
```
koonpeng, line 40: remove this while the backend isn't available.

packages/dashboard/src/components/emergency-alarm.tsx
```diff
@@ -0,0 +1,20 @@
+import { IconButton } from '@material-ui/core';
+import React from 'react';
+import NotificationsActiveIcon from '@material-ui/icons/NotificationsActive';
+
+interface EmergencyAlarmProps {
```
koonpeng, line 5: good to export this

packages/dashboard/src/components/appbar.tsx
```diff
@@ -13,15 +13,17 @@ import SettingsIcon from '@material-ui/icons/Settings';
 import React from 'react';
 import { AuthenticatorContext, UserContext } from './auth/contexts';
 import HelpIcon from '@material-ui/icons/Help';
+import { EmergencyAlarm } from './emergency-alarm';
```
koonpeng, line 16: remove this while the backend support isn't available.

packages/dashboard/src/components/appbar.tsx
```diff
@@ -13,15 +13,17 @@ import SettingsIcon from '@material-ui/icons/Settings';
...
 
 export interface AppBarProps {
   toggleShowOmniPanel(): void;
   showSettings(show: boolean): void;
   showHelp(show: boolean): void;
+  alarmState: boolean | null;
```
koonpeng, line 22: make it optional, with a `TODO` tag explaining it's not currently used because of the lack of usable backend.

packages/dashboard/src/components/tests/appbar.test.tsx
```diff
@@ -1,8 +1,10 @@
 import { createMount } from '@material-ui/core/test-utils';
+import NotificationsActiveIcon from '@material-ui/icons/NotificationsActive';
 import React from 'react';
 import FakeAuthenticator from '../../mock/fake-authenticator';
 import AppBar from '../appbar';
 import { AuthenticatorContext, UserContext } from '../auth/contexts';
+import { EmergencyAlarm } from '../emergency-alarm';
```
koonpeng, line 7: remove this while the backend isn't available

packages/dashboard/src/components/tests/confirm-alert-dialog.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+        cancelCallback={cancelCallback}
+        confirmCallback={() => jest.mock}
+        close={() => jest.mock}
+      ></ConfirmAlertDialog>,
+    );
+    root.find('#alert-dialog-cancel-button').first().simulate('click');
+    expect(cancelCallback).toHaveBeenCalledTimes(1);
+    root.unmount();
+  });
+  test('Confirm callback is executed', () => {
```
koonpeng, line 53: nit: empty line between functions

packages/dashboard/src/components/tests/confirm-alert-dialog.test.tsx
```diff
@@ -0,0 +1,93 @@
...
+        cancelCallback={() => jest.mock}
+        confirmCallback={confirmCallback}
+        close={() => jest.mock}
+      ></ConfirmAlertDialog>,
+    );
+    root.find('#alert-dialog-confirm-button').first().simulate('click');
+    expect(confirmCallback).toHaveBeenCalledTimes(1);
+    root.unmount();
+  });
+  test('Modal is closed', () => {
```
koonpeng, line 72: nit: empty line between functions

koonpeng (changes requested):

packages/dashboard/src/components/confirm-alert-dialog.tsx
matiasbavera: Makes sense!

packages/dashboard/src/components/confirm-alert-dialog.tsx
matiasbavera: Ok, let's make the parent handle the closing from the confirm/cancel.

koonpeng (approved):

matiasbavera:
@vallq

PR #139, opened 2020-10-13
Run docker e2e test from local dev env

### LINKED ISSUE #138: Error while trying to run npm run test:e2e
Error while trying to run on local env
```
npm run test:e2e 
. e2e/env.sh && scripts/dockert docker-compose -f docker/docker-compose.yml up e2e
docker-compose -f docker/docker-compose.yml up e2e
```

```
e2e_1        | npm WARN prepare removing existing node_modules/ before installation
e2e_1        | npm ERR! Error while executing:
e2e_1        | npm ERR! /usr/bin/git ls-remote -h -t git://github.com/RobotWebTools/rclnodejs.git
e2e_1        | npm ERR! 
e2e_1        | npm ERR! fatal: failed to stat '/root/romi-dashboard/packages/dashboard': Permission denied
e2e_1        | npm ERR! 
e2e_1        | npm ERR! exited with error code: 128
```

`docker-compose -f docker/docker-compose.yml up dev
`it's working correctly

I think it has something to do with the permissions when executing npm ci.

matiasbavera:
## What's new

fixes #138

* Add a service on `docker-compose.yml` file to run e2e tests on a container locally:  
`docker-compose -f docker/docker-compose.yml up e2e-local`
* Added an apt-get update to the `e2e.dockerfile`, to install the last stable version of chrome.

## Self-checks
Does not apply.

### DIFF
--- a/packages/dashboard/docker/docker-compose.yml
+++ b/packages/dashboard/docker/docker-compose.yml
@@ -23,6 +23,22 @@ services:
     network_mode: host
     privileged: true
     command: bash -c 'npm ci --unsafe-perm && (npm test || npm test || npm test)'
+  e2e-local:
+    build:
+      context: ..
+      dockerfile: docker/e2e.Dockerfile
+    image: docker.pkg.github.com/osrf/romi-dashboard/e2e
+    volumes:
+      - /var/run/docker.sock:/var/run/docker.sock
+      - ../../..:/root/romi-dashboard
+    working_dir: /root/romi-dashboard/packages/dashboard
+    command: bash -c 'cd e2e/ && npm test'
+    environment:
+      CI: 1
+      CHROME_BIN: /chrome-no-sandbox
+    ipc: host
+    network_mode: host
+    privileged: true
   unit-test:
     image: docker.pkg.github.com/osrf/romi-dashboard/e2e
     volumes:
--- a/packages/dashboard/docker/e2e.Dockerfile
+++ b/packages/dashboard/docker/e2e.Dockerfile
@@ -2,6 +2,7 @@ FROM docker.pkg.github.com/osrf/romi-dashboard/rmf:nightly
 
 RUN wget -q -O - https://dl.google.com/linux/linux_signing_key.pub | apt-key add - \
   && sh -c 'echo "deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main" >> /etc/apt/sources.list.d/google-chrome.list' \
+  && apt-get update \
   && curl -sL https://deb.nodesource.com/setup_12.x | bash - \
   && apt-get install -y nodejs google-chrome-stable docker.io docker-compose \
   && rm -rf /var/lib/apt/lists/*


koonpeng:
how is this different from the "e2e" service?

matiasbavera:
> how is this different from the "e2e" service?

It skips the npm ci, which when trying to remove `node_packages` outside the docker it's failing.

koonpeng:
> > how is this different from the "e2e" service?
> 
> It skips the npm ci, which is having permission problems when trying to remove `node_packages` outside the docker it's failing.
> 
> I'll look for another solution. setting this as a draft again.

Yeah, I don't think this is a good solution. We need to do `npm ci` inside the container because it requires rclnodejs (aka rmf/ros2). If we can do it outside, it means the host has a rmf installation and you should be able to run the e2e directly on the host.

matiasbavera:
> > > how is this different from the "e2e" service?
> > 
> > 
> > It skips the npm ci, which is having permission problems when trying to remove `node_packages` outside the docker it's failing.
> > I'll look for another solution. setting this as a draft again.
> 
> Yeah, I don't think this is a good solution. We need to do `npm ci` inside the container because it requires rclnodejs (aka rmf/ros2). If we can do it outside, it means the host has a rmf installation and you should be able to run the e2e directly on the host.


Well, I did some research on this and test it. 
These are the situations:

- We currently can install (npm install) the project locally if we have the RMF project installed locally, generally, we install the `node_modules` with our host user, let's called `userX`. When we try to run the e2e test inside the container, the container points its volume to the project folder hosted in a directory outside the docker, and owned by the `userX`. The problem here is when we use `npm ci` inside the container (that is what the docker-compose up e2e does), the container uses its own `Nodejs` and the `root` user to try to reinstall the need packages. This causes problems with the npm installation. I also try to `npm ci` packages as the root user on my PC let's called `PCroot`, having the same error that when I tried to run the command with the container. I tried to change the ownership of the folder to `root`, but didn't worked.
-  If we do a fresh install with the PCroot and we later run the `npm ci` inside the container works perfectly. The node_packages is owned by root.
- If we do a fresh install with the `docker-compose -f docker/docker-compose.yml up e2e` works perfectly too. The node_packages is owned by root.

 These are a couple of possible solutions if we want to run `e2e:test` from the docker locally:

1.  We can ask the developers that if they going to work with containers install the packages from the container. With this, we will be using the `node_modules` installed by the container.

2. We can skip the `npm ci` inside the docker, as it is now `e2e-dev` but the developer will have to take care of the installations of packages.

3. We can check in the `install.js` script if the `node_modules` belongs to the `root` instead of 1000:1000 (userX). If it belongs to userX we can remove and reinstall it with the container's Nodejs and its user (root).

4. I think there is a way to pass the user's file to the container and run things as that user. Let's say the container access the folder outside the container as `userX`. This option is trickier than the others and we need to be cautious because the container will be losing privileges.

Off-topic: maybe it's a good idea to have a `docker-compose` service that just installs the `node_packages`, if the person it's only going to work with containers. E.g:

```
# This will install the node_modules on the volume outside the container
docker-compose -f docker/docker-compose.yml up install  
docker-compose -f docker/docker-compose.yml up dev

```

What do you think about it @koonpeng ?

koonpeng:
I'm not sure if supporting a mixed dev environment worth it, there are too many little differences between docker and host that will make things fail. I have been thinking of dropping `start:docker` because of that too.

Due to all the little differences, I think it's best if we just drop the docker npm scripts. So instead of a mixed environment where you run `start:docker` or `test:docker` on the host, just move everything to docker and run everything from within the container. 

The problem though is that devs will have to set up their own docker env, that depends on the IDE/tools you use so we can't really provide a ready-to-go solution. E.g. If you are using vscode, you can use the container remote extension, or if you are using vim, I suppose you can install vim and all your scripts inside the container and launch vim from within.

There are also many other problems with docker based dev which requires the dev to configure the container manually, for example GPU acceleration, running as the current user, user namespace remapping, apt mirror configuration etc.

We still need the docker-compose file for e2e ci (for now). Ideally we can run the full ci in a container but there is some problems with the network configuration and nested docker (because of auth) which I haven't gotten working yet. The "github actions" solution is to run auth as a service container but I don't want to do that because it creates a disconnect between local runs and ci runs. Ideally the ci would just run `npm run test:e2e` similar to the local run, without any special setups etc.

matiasbavera:
> I'm not sure if supporting a mixed dev environment worth it, there are too many little differences between docker and host that will make things fail. I have been thinking of dropping `start:docker` because of that too.
> 
> Due to all the little differences, I think it's best if we just drop the docker npm scripts. So instead of a mixed environment where you run `start:docker` or `test:docker` on the host, just move everything to docker and run everything from within the container.
> 
> The problem though is that devs will have to set up their own docker env, that depends on the IDE/tools you use so we can't really provide a ready-to-go solution. E.g. If you are using vscode, you can use the container remote extension, or if you are using vim, I suppose you can install vim and all your scripts inside the container and launch vim from within.
> 
> There are also many other problems with docker based dev which requires the dev to configure the container manually, for example GPU acceleration, running as the current user, user namespace remapping, apt mirror configuration etc.
> 
> We still need the docker-compose file for e2e ci (for now). Ideally we can run the full ci in a container but there is some problems with the network configuration and nested docker (because of auth) which I haven't gotten working yet. The "github actions" solution is to run auth as a service container but I don't want to do that because it creates a disconnect between local runs and ci runs. Ideally the ci would just run `npm run test:e2e` similar to the local run, without any special setups etc.

I understand and I agreed with your concerns. I'll close this PR. As you said we can remove the `start:docker` giving the sign that we are not supporting it anymore but I think we can leave the `dev` on the `docker-compose` file in case anyone wants to use it without warranties. For the `test:e2e:docker` we can create a script that copies all the files (except for the `node_modules` of course) and runs from the docker directly.

> The "github actions" solution is to run auth as a service container

Makes sense because they're different services, but I understand where you want to go. 

> Ideally we can run the full ci in a container but there is some problems with the network configuration and nested docker (because of auth) which I haven't gotten working yet

Hey, I'm no expert but let me give you a hand on this. In the worst-case scenario, we can also install the service manually (without docker) inside the container. The downside is that we will have to install, configure and serve the service and the other downside is that we will no be replicating the deployment setup

PR #142, opened 2020-10-14
Fix/negotiations ui improvements

ChawinTan:
## What's new

Add 3 more functionality to the negotiations panel, namely:
- button to clear all existing negotiations
- button to reset selected negotiations
- button to restore all cleared negotiations

## fixes

Improvement to the user interface of the negotiation panel, namely:
- make the tree structure more obvious
- have each negotiation and it's content separated for easier viewing
- improvements to the overall presentation

## [Describe your changes]
negotiation_panel1

Added 3 more buttons at the top of the negotiations panel, namely `clear`, `reset` and `restore`. The buttons are disabled until `conflicts` prop is not empty.

Upon clicking the `clear` button, localStorage will store the list of nodeIds of the conflicts that are currently presented on the panel and reset the `negotiationContents` state to an empty object. Subsequent renders with new nodeIds in the `conflicts` prop will be compared to the ids in localStorage and will not be updated in the `negotiationContents` if the ids exist in localStorage. Do also note that I refactored the `negotiationContents` to be an object in `useState` instead of a standalone array as I think it will be beneficial in future if we want to manipulate the negotiations.

Clicking the `reset` button will set `negotiationTrajStore` back to an empty object and clear the highlighted negotiation trajectories while automatically collapsing any expanded tree node and resetting the selected colors.

Clicking restore will clear the list of nodeIds stored in localStorage and will render all the negotiations on negotiations panel. I added this button so that users don't have to clear cache every time they want to reset the nodeIds.

Read the 2 sections above

## Self-checks

## Discussion

Should we include the buttons in the tour guide? I feel that it might be quite hard to integrate the buttons into the guide due to it's implementation so maybe we can have a separate FAQ page to tell users what the buttons are meant to do?


matiasbavera:
> Should we include the buttons in the tour guide? I feel that it might be quite hard to integrate the buttons into the guide due to it's implementation

Mmm.. maybe I'm missing something but I think we can explain what each button does and select them as well. Let's check with @val.

> we can have a separate FAQ page to tell users what the buttons are meant to do

I like this idea as well.

packages/dashboard/src/components/negotiations-panel.tsx
```diff
@@ -275,17 +318,74 @@ export const NegotiationsPanel = React.memo((props: NegotiationsPanelProps) => {
...
+  };
+
+  const handleResetNegotiations = () => {
+    setExpanded([]);
+    setSelected('');
+    setNegotiationTrajStore({});
+  };
+
+  const handleRestoreNegotiations = () => {
+    localStorage.setItem('conflictIds', '');
```
matiasbavera, line 336: A couple of comments on this:
I think this would great if the browser restarts suddenly but I don't know if we should be using localStorage here. In theory in the future, and please correct me if I'm the wrong @koonpeng , we should be consuming the history from the backend directly so we shouldn't be storing the conflicts. 

Some comments:
I saw that there are leftovers persisting after a refresh or a new session but it's not affecting the behavior though.
image

As an alternative, we can use a state (list or a stack) to store the conflicts in memory. But as you know this will disappear after a browser refresh.


For me, in the meanwhile, we can keep using the local storage to store that negotiation history, we can set the maximum quantity of negotiation to save too.

packages/dashboard/src/components/negotiations-panel.tsx
```diff
@@ -275,17 +318,74 @@ export const NegotiationsPanel = React.memo((props: NegotiationsPanelProps) => {
...
+    setNegotiationContents({});
+  };
+
+  const handleNodeToggle = (event: React.ChangeEvent<{}>, nodeIds: string[]) => {
+    setExpanded(nodeIds);
   };
 
   return (
     <Typography variant="body1" component={'span'}>
+      <div style={{ padding: '0.5rem 1rem' }}>
```
matiasbavera, line 346: What do you think about using className here and move it to `useStyles` to keep consistency?

matiasbavera (commented):
First of all thanks for this @ChawinTan, looks good! I left a couple of comments.

packages/dashboard/src/components/negotiations-panel.tsx
ChawinTan: I was also trying not to use local storage initially but `useState` is also extremely tricky to use for this case due to how it is designed.

> I saw that there are leftovers persisting after a refresh or a new session but it's not affecting the behavior though.

Yup, there are leftovers after a browser refresh but I guess since the existing conflicts are cleared, the stored values don't matter anymore.


> As an alternative, we can use a state (list or a stack) to store the conflicts in memory. But as you know this will disappear after a browser refresh.

Sounds like a good idea. Now that I think of it, since refreshing the browser will reset the existing conflicts, resetting this list also doesn't matter anymore

ChawinTan:
> Mmm.. maybe I'm missing something but I think we can explain what each button does and select them as well. Let's check with @Val.

Oh ... cause I was worried that users may not know they need to reset the local storage through the reset button. Maybe I am thinking too much about this.  I think a simple, temporary FAQ page should suffice if this is needed as I assume we would be changing data sources?

packages/dashboard/src/components/negotiations-panel.tsx
koonpeng: I agree that local storage is not a good idea. I don't know how it gets the negotiations, why do we not want to clear the array? Even if so, can't we use state/memo/ref to persist the array across renders?

ChawinTan: > I agree that local storage is not a good idea. I don't know how it gets the negotiations, why do we not want to clear the array? Even if so, can't we use state/memo/ref to persist the array across renders?

Oops :facepalm: 

Ok so initially the confusion was with this piece of code here:

```
let negotiationContents: JSX.Element[] = [];
  if (conflicts) {
    let reversedConflicts = Object.keys(conflicts).reverse();
    reversedConflicts.forEach((version) => {
      const conflict = conflicts[version];
      let contents = renderNegotiations(version, conflict);
      negotiationContents.push(contents);
    });
  } else {
    console.log('prop negotationstatus empty');
  }
```

it wasn't in a function and was needed on the initial render so I can't use setState here. I tried putting it in `useEffect` but couldn't work. I then tried out all the above options, as you mentioned (useMemo, refs, useState) but all did not work too. I then refactored the negotiations to be an object in useState instead of an array hanging by itself.

Just tested saving the array in useState and it works now. Complete oversight from me and totally forgot the initial intentions of the refactor :sweat_smile:

koonpeng: I think you should be able to use ref in this case

```tsx
const negotiationContentsRef = React.useRef<JSX.Element[]>([]);
const negotiationContents = negotiationContentsRef.current;
```

That should make it persist across renders.

ChawinTan: > I think you should be able to use ref in this case
> 
> ```tsx
> const negotiationContentsRef = React.useRef<JSX.Element[]>([]);
> const negotiationContents = negotiationContentsRef.current;
> ```
> 
> That should make it persist across renders.

Just tried it but I currently can't think of a way to re-render the component when I modify the value of ref, so I think I would still use useState for negotiationContents instead.

matiasbavera (approved):
LGTM!

PR #144, opened 2020-10-14
Check tasks (plans) on the Dashboard

matiasbavera:
## What's new
pre-requisite https://github.com/osrf/romi-dashboard/pull/153 [Merged]

fixes #

* Created a taskSummary windows on the omnipannel with a tree structure, 
* Subscribed to TaskSummary
* Created mock
* Added a pause button but it is not implemented yet.
* Add unit tests
* Add e2e tests

image

## Self-checks

- [X] I tested the behavior of the components that interact with the backend, with an e2e test

## Discussion

If we will keep the tree button group component I'll refactor the negotiation panel in another PR

### DIFF
--- a/packages/dashboard/src/components/task-summary-panel.tsx
+++ b/packages/dashboard/src/components/task-summary-panel.tsx
@@ -11,6 +11,7 @@ import PlayCircleFilledWhiteIcon from '@material-ui/icons/PlayCircleFilledWhite'
 import TaskManager from '../managers/task-manager';
 import { colorPalette } from '../util/css-utils';
 import { TreeButtonGroup } from './tree-button-group';
+import { Theme } from '@material-ui/core/styles/createMuiTheme';
 
 const debug = Debug('OmniPanel:TaskSummaryPanel');
 
@@ -37,7 +38,9 @@ export const TaskSummaryPanel = React.memo((props: TaskSummaryPanelProps) => {
   }>({});
   const [expanded, setExpanded] = React.useState<string[]>([]);
   const [selected, setSelected] = React.useState<string>('');
-  const [savedTasksContent, setSavedTasksContent] = React.useState<{
+
+  // We need to persist across the renders
+  const savedTasksContent = React.useRef<{
     [key: string]: JSX.Element;
   }>({});
 
@@ -55,21 +58,20 @@ export const TaskSummaryPanel = React.memo((props: TaskSummaryPanelProps) => {
   };
 
   const handleClearAllCurrTasks = () => {
-    setSavedTasksContent(taskContents);
+    savedTasksContent.current = taskContents;
     setTaskContents({});
   };
 
   const handleRestoreTasks = () => {
     // Assigning another reference
     let taskContentsTemp = Object.assign({}, taskContents);
-    Object.keys(savedTasksContent).forEach((element) => {
-      console.log(!(element in taskContents));
+    Object.keys(savedTasksContent.current).forEach((element) => {
       if (!(element in taskContents)) {
-        taskContentsTemp[element] = savedTasksContent[element];
+        taskContentsTemp[element] = savedTasksContent.current[element];
       }
     });
     setTaskContents(taskContentsTemp);
-    setSavedTasksContent({}); // setTaskContents({});
+    savedTasksContent.current = {};
   };
 
   const determineStyle = (state: number): string => {
@@ -87,19 +89,19 @@ export const TaskSummaryPanel = React.memo((props: TaskSummaryPanelProps) => {
     }
   };
 
-  const renderActor = (taskStatus: string): React.ReactElement | undefined => {
-    const actor = TaskManager.getActorFromStatus(taskStatus);
-    if (!actor) return;
-    return (
-      <Typography variant="body1" className={classes.taskActor}>
-        {actor}
-      </Typography>
-    );
-  };
+  // Update Task list content
+  React.useEffect(() => {
+    const renderActor = (taskStatus: string): React.ReactElement | undefined => {
+      const actor = TaskManager.getActorFromStatus(taskStatus);
+      if (!actor) return;
+      return (
+        <Typography variant="body1" className={classes.taskActor}>
+          {actor}
+        </Typography>
+      );
+    };
 
-  // Set it like useCallback because is used inside a hook
-  const renderTaskTreeItem = React.useCallback(
-    (task: RomiCore.TaskSummary) => {
+    const renderTaskTreeItem = (task: RomiCore.TaskSummary) => {
       return (
         <TreeItem
           data-component="TreeItem"
@@ -131,18 +133,16 @@ export const TaskSummaryPanel = React.memo((props: TaskSummaryPanelProps) => {
           <TaskSummaryPanelItem task={task} key={task.task_id} />
         </TreeItem>
       );
-    },
-    [classes],
-  );
+    };
 
-  // Update Task list content
-  React.useEffect(() => {
     if (allTasks.length === 0) return;
     let taskContentsTemp = Object.assign({}, taskContents);
     allTasks.forEach((task) => {
       taskContentsTemp[task.task_id] = renderTaskTreeItem(task);
     });
     setTaskContents(taskContentsTemp);
+    // Ignoring dependency taskContents, it'll enter in a infinite loop
+    // eslint-disable-next-line
   }, [allTasks]);
 
   return (
@@ -175,7 +175,7 @@ export const TaskSummaryPanel = React.memo((props: TaskSummaryPanelProps) => {
   );
 });
 
-const useStyles = makeStyles((theme) => ({
+const useStyles = makeStyles((theme: Theme) => ({
   root: {
     padding: '1rem',
   },
--- a/packages/dashboard/src/components/task-summary-panel.tsx
+++ b/packages/dashboard/src/components/task-summary-panel.tsx
@@ -51,7 +51,7 @@ export const TaskSummaryPanel = React.memo((props: TaskSummaryPanelProps) => {
       selected={selected}
     >
       {allTasks.map((task) => (
-        <TaskSummaryPanelItem task={task} />
+        <TaskSummaryPanelItem task={task} key={task.task_id} />
       ))}
     </TreeView>
   );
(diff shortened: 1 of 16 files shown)


matiasbavera:
Converting to draft again, I Refactored the code, test are missing

packages/dashboard/src/components/task-summary-panel.tsx
```diff
@@ -0,0 +1,221 @@
...
+  const handleSelect = (event: React.ChangeEvent<{}>, nodeIds: string) => {
+    setSelected(nodeIds);
+  };
+
+  const handleResetTasks = () => {
+    setExpanded([]);
+    setSelected('');
+  };
+
+  const handleClearAllCurrTasks = () => {
```
koonpeng, line 60: Why would the user want to clear current tasks? From what I understand, this only clears the task from the list, and does not actually cancel the tasks.

packages/dashboard/src/managers/task-manager.ts
```diff
@@ -0,0 +1,53 @@
...
+        this.emit('updated');
+      }),
+    );
+  }
+
+  stopAllSubscriptions(): void {
+    this._subscriptions.forEach((sub) => sub.unsubscribe());
+  }
+
+  static getActorFromStatus(status: string) {
```
koonpeng, line 26: This looks like a hack, can you add a `TODO` and a issue (here and at `rmf_core`) to provide a reliable way to retrieve the actor from the message?

packages/dashboard/src/components/task-summary-panel.tsx
```diff
@@ -0,0 +1,221 @@
...
+        </TreeItem>
+      );
+    };
+
+    if (allTasks.length === 0) return;
+    let taskContentsTemp = Object.assign({}, taskContents);
+    allTasks.forEach((task) => {
+      taskContentsTemp[task.task_id] = renderTaskTreeItem(task);
+    });
+    setTaskContents(taskContentsTemp);
```
koonpeng, line 143: You should be able to use the functional form to avoid recursive dependencies and use of a temp object.
```tsx
setTaskContents((currentContents) => {
  return doStuff(currentContents);
});
```

packages/dashboard/src/components/task-summary-panel.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+  const handleRestoreTasks = () => {
+    // Assigning another reference
+    let taskContentsTemp = Object.assign({}, taskContents);
+    Object.keys(savedTasksContent.current).forEach((element) => {
+      if (!(element in taskContents)) {
+        taskContentsTemp[element] = savedTasksContent.current[element];
+      }
+    });
+    setTaskContents(taskContentsTemp);
```
koonpeng, line 73: You should be able to use the functional form to avoid recursive dependencies and use of a temp object.
```tsx
setTaskContents((currentContents) => {
  return doStuff(currentContents);
});
```

packages/dashboard/src/managers/task-manager.ts
```diff
@@ -0,0 +1,53 @@
...
+    // Gets the name of the robot if it has any
+    // eslint-disable-next-line
+    return status.match(/\[[A-Za-z]([a-zA-Z0-9\/]){3,}\]+/gi);
+  }
+
+  static formatStatus(status: string) {
+    return status.split('|');
+  }
+
+  static getStateLabel(state: number): string {
```
koonpeng, line 36: Better to keep presentation logic out of here, move it to the component instead.

koonpeng (changes requested):

packages/dashboard/src/components/task-summary-panel.tsx
matiasbavera: Yeah, it only clears the task from the list and does not cancel the task.

I want to implement the task pause in a FUP PR. 

I just thought that they will want to clear it when they have a lot of old finished tasks on the list.

packages/dashboard/src/managers/task-manager.ts
matiasbavera: Sure, sounds good! I'll create and issue and add a todo

matiasbavera: https://github.com/osrf/rmf_core/issues/205

packages/dashboard/src/components/task-summary-panel.tsx
matiasbavera: I looked into it and we cannot assign directly the stored values to `currentContent` because it is updating the taskContents too, so when we set the taskContent new value with `setTaskContents` it'll no re-render because it'll not detect any changes.
We can also force the re-render of the component

packages/dashboard/src/components/task-summary-panel.tsx
koonpeng: Hmm, it's the lack of a backend database problem again. If we keep the previous tasks in memory it will be gone when we refresh the page, we can't persist the tasks without a db. I guess this is the best we can do now so that the user can see the result of the task. Can you add a TODO here?

packages/dashboard/src/components/task-summary-panel.tsx
```diff
@@ -0,0 +1,252 @@
...
+            root: classes.treeChildren,
+            expanded: classes.expanded,
+          }}
+          label={
+            <>
+              <Typography variant="body1" noWrap>
+                {task.state === RomiCore.TaskSummary.STATE_ACTIVE && (
+                  <IconButton
+                    // TODO: replace onClick with the pause plans logic.
+                    onClick={(e) => {
```
koonpeng, line 131: make it not clickable?

packages/dashboard/src/components/task-summary-panel.tsx
```diff
@@ -0,0 +1,252 @@
...
+    setTaskContents({});
+  };
+
+  const handleRestoreTasks = () => {
+    // Adding this because the test interpreter cannot find the reference to `savedTasksContent current`
+    // inside the setTaskContents
+    const storedTasks = savedTasksContent.current;
+    setTaskContents((currentContent) => {
+      // Assigning another reference.
+      let taskContentsTemp = Object.assign({}, currentContent);
```
koonpeng, line 72: use `const`. Also let's rename this to `newTaskContents` to better reflect its purpose.

packages/dashboard/src/components/task-summary-panel.tsx
```diff
@@ -0,0 +1,252 @@
...
+      // We cannot assign directly the stored values to `currentContent` because it is updating
+      // the taskContents too, so when we set the taskContent new value with `setTaskContents`
+      // it'll no re-render because it'll not detect any changes.
+      Object.keys(storedTasks).forEach((element) => {
+        if (!(element in currentContent)) {
+          taskContentsTemp[element] = storedTasks[element];
+        }
+      });
+      console.log(taskContentsTemp);
```
koonpeng, line 81: leftover?

packages/dashboard/src/components/task-summary-panel.tsx
```diff
@@ -0,0 +1,252 @@
...
+      // it'll no re-render because it'll not detect any changes.
+      allTasks.forEach((task) => {
+        taskContentsTemp[task.task_id] = renderTaskTreeItem(task);
+      });
+      return taskContentsTemp;
+    });
+
+    // Ignoring dependency taskContents, it'll enter in a infinite loop
+    // eslint-disable-next-line
+  }, [allTasks]);
```
koonpeng, line 177: eslint warning
```
React Hook React.useEffect has missing dependencies: 'classes.expanded', 'classes.labelContent', 'classes.taskActor', 'classes.treeChildren', and 'determineStyle'. Either include them or remove the dependency array.
```

packages/dashboard/src/components/task-summary-panel.tsx
```diff
@@ -0,0 +1,252 @@
...
+      // the taskContents too, so when we set the taskContent new value with `setTaskContents`
+      // it'll no re-render because it'll not detect any changes.
+      allTasks.forEach((task) => {
+        taskContentsTemp[task.task_id] = renderTaskTreeItem(task);
+      });
+      return taskContentsTemp;
+    });
+
+    // Ignoring dependency taskContents, it'll enter in a infinite loop
+    // eslint-disable-next-line
```
koonpeng, line 176: This no longer holds a dependency on `taskContents` we can remove this.

packages/dashboard/src/components/task-summary-panel.tsx
```diff
@@ -0,0 +1,252 @@
...
+          <TaskSummaryPanelItem task={task} key={task.task_id} />
+        </TreeItem>
+      );
+    };
+
+    if (allTasks.length === 0) return;
+
+    setTaskContents((currentContent) => {
+      // Assigning another reference.
+      let taskContentsTemp = Object.assign({}, currentContent);
```
koonpeng, line 165: use `const`. Also let's rename this to `newTaskContents` to better reflect its purpose.

koonpeng (changes requested):

koonpeng:
Peek 2020-11-06 09-37
I have a bug where the the color is "stucked" in the hover state after clicking the task.

packages/dashboard/src/components/task-summary-panel.tsx
matiasbavera: Sure thing!

matiasbavera:
> I have a bug where the the color is "stucked" in the hover state after clicking the task.

You're right, I'll take a look. Thanks

packages/dashboard/src/components/tree-button-group.tsx
```diff
@@ -0,0 +1,57 @@
...
+          id="restore-button"
+          disabled={disableRestore}
+          onClick={() => handleRestoreClick && handleRestoreClick()}
+        >
+          <RestoreFromTrashIcon />
+          Restore
+        </Button>
+      </ButtonGroup>
+    </div>
+  );
```
matiasbavera, line 56: Should I move this to `react-component` package on this PR?

koonpeng: yea, but we can merge this first and have that in another PR.

koonpeng (approved):
Thanks for this! Can you make another PR to migrate the components to `react-components`? Try to re-use some of the code also (`SimpleInfoData` for example).

matiasbavera:
> Thanks for this! Can you make another PR to migrate the components to `react-components`? Try to re-use some of the code also (`SimpleInfoData` for example).

Sure thing, leave it to me.

PR #145, opened 2020-10-19
remove checks

ChawinTan:
## fixes

Remove codecov in github checks..

[Describe your changes]

Added a codecov file with necessary configuration.

## Self-checks

## Discussion

What are some other files that can be ignored? Do we need the state-managers to run in the unit tests since I suppose they will be covered in the e2e tests?

[Questions for reviewers]

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,0 +1,2 @@
+github_checks:
+    annotations: false
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,2 +1,4 @@
-github_checks:
-    annotations: false
+coverage:
+  status:
+    project: off
+    patch: off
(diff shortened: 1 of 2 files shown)


matiasbavera:
> What are some other files that can be ignored? Do we need the state-managers to run in the unit tests since I suppose they will be covered in the e2e tests?


I think you can ignore them but it depends on the manager as well, not all the managers handle a backend connection and there are a lot of things that can test it with a unit test. So, I think if the manager is been test it on the e2e test you can ignore it.

matiasbavera (approved):

koonpeng:
> What are some other files that can be ignored? Do we need the state-managers to run in the unit tests since I suppose they will be covered in the e2e tests?

I would encourage any refactoring or extra tests be moved to the component library in `refactor/component-lib` branch.

PR #151, opened 2020-10-22
Fix/e2e container communications

ChawinTan:
## What's new

- Removed `e2e` from docker-compose
- e2e tests can now be ran from the same command, `npm run test:e2e`
- A bunch of github workflow related changes mainly in `auth-ready.js`, `pull-requesnt.yml` and `env.sh`.

## fixes # [related issue, if exist]

This pr is related to issue  #138 

The main goal of this pr is to unify the process of e2e testing using the `npm run test:e2e` process for both CI and local runs. The existing set up has 2 separate process:

- `npm run test:e2e` for local testing
- using docker-compose for github's workflow CI.

By standardizing the launch process, we don't have to maintain different ways to probe for the availability of servers/services during testing.

<b>Major fixes</b>
- Differentiating between local and remote CI runs
- Using the same network between auth container and dashboard container
- Connecting to the Auth server
- Authentication errors from script tag

<b>Minor fixes</b>
- Networking error from webdriver

## Describe your changes

Following the points above in `fixes`:

<b>Differentiating between local and remote CI runs</b>
In `auth-ready`, this is done by checking for `CI` env variable in an if statement.

In `env.sh`, `CI` is used to determine if `REACT_APP_AUTH_CONFIG` is required to be set.

<b>Using the same network between auth container and dashboard container</b>
The first part to this problem is to only switch the network of the auth container when the service is fully loaded. Switching networks while the service is still building will throw errors in the auth container. I changed the `retryTimer` interval from 1 second to 20 seconds to ensure ample time for the auth service to build before the next attempt to connect.

Secondly, due to the need to use ip addresses in the `172.16.xx.x` range, I decided to create a new network instead of using an existing one by github since we need a network with a user defined subnet to manually assign a container to an ip address when switching network.

Finally, I disconnected the containers from their old networks to prevent certain commands to return 2 different addresses concatenated together so as to prevent unnecessary operations.

<b>Connecting to the Auth server</b>
This is done by directly assigning the ip address of the auth container to url instead of using localhost.

`http://${authIpAddress ? authIpAddress : 'localhost'}:8080/auth/`

<b>Authentication errors from script tag</b>
This is resolved by setting` REACT_APP_AUTH_CONFIG` with the following value:

```
REACT_APP_AUTH_CONFIG='{"realm":"master", "clientId":"romi-dashboard", "url":"http://172.16.0.2:8080/auth"}'
```

<b>Networking error from webdriver</b>

Included `whitelisted-ips` flags in services in `wdio.conf.js`

```
services: [
    [
      'chromedriver',
      {
        args: ['--whitelisted-ips='],
      },
    ],
  ],
```

## Self-checks

## Discussion

[Questions for reviewers]


ChawinTan:
Finally managed to connect to the auth service although it is by directly assigning the ip address of the auth container and not through localhost.

`http://${authIpAddress ? authIpAddress : 'localhost'}:8080/auth/`

However, the tests are still failing because of the chrome driver probably from certain wdio config as it does allow connection to the dashboard on localhost. It seems like the github environment blocks access to the localhost ip address unless they are defined as services in the yml file? 

I will reading up on wdio to find out more on how it works.

Should we even be assigning ip address directly?

matiasbavera:
Thanks for the update. 

> It seems like the github environment blocks access to the localhost ip address unless they are defined as services in the yml file? I will reading up on wdio to find out more on how it works.

Oh, this can be an annoying problem. Let me know what you find =).

> Should we even be assigning ip address directly?

if you mean here `http://${authIpAddress ? authIpAddress : 'localhost'}:8080/auth/`
Since they are dynamically assigned looks like it fits our needs for now.

ChawinTan:
I'm still unable to get the `localhost:5000` to render the login page. 

As you can see, locally we can see the css and html of the login page being rendered:
local_run

On the workflow side, we don't see any of it:
github_workflow

I've also been trying to debug this error but to no avail. Base on what I found, it seems to suggest that the driver is trying to listen to ivp6 but docker doesn't have it. It will fall back to ivp4 automatically if it fails so the tests are still running fine. Another possibility is that `localhost` is preventing a proper connection so we should use `0.0.0.0` to "allow all possible ips".  Tried but did not work either. Also tried using github's network ip but also failed.

Not sure if this is a cause of login page not rendering but I suspect it might be due to a connection failure.
`bind() failed: Cannot assign requested address (99)`

Anyone has any ideas on or the direction I should be looking at?

matiasbavera:
> Anyone has any ideas on or the direction I should be looking at?

You may have already analyzed these points but I leave these comments just in case:

If the rmf and authentication containers have different IPs, one of the problems could also be that the browser (in this case the chrome of the chromedriver) is somehow blocking the redirection to the authentication screen.

They can also be blocked by some type of firewall that only allows you to interact with certain ports. But I don't know how much I can put my hand on fire for this theory. The error in the image, is it an error thrown by the RMF service or is it what it returns when the connection with the authentication service fails?

It may be helpful to make a simple drawing in draw.io to see graphically what is happening and what you want to achieve. This will also serve us for documentation.

ChawinTan:
Finally managed to get the login test working!!!!

TLDR, I used an ip in the 172.16.x.x range to replace localhost since it is also a private ip. This eliminated the need for ssl  in `http://localhost:8080/auth/realms/master/protocol/openid-connect/3p-cookies/step1.html` in the script tag while also avoiding localhost. 

Thanks @koonpeng  and @matiasbavera for all the help during the past 3 weeks :smiley: 

> It may be helpful to make a simple drawing in draw.io to see graphically what is happening and what you want to achieve. This will also serve us for documentation.

I think I can start doing this up after cleaning up.

I notice that the error for the test immediately after the login test is different from what we are seeing where it may or may not be related to this pr. Do you think we should wait for the new `rclnodejs` to be ready first or we can look into this issue in a separate pr? Just attaching this for visibility.

Different_error

koonpeng:
> I notice that the error for the test immediately after the login test is different from what we are seeing where it may or may not be related to this pr. Do you think we should wait for the new rclnodejs to be ready first or we can look into this issue in a separate pr? Just attaching this for visibility.

There is multiple bugs actually.

1. The latest node v12 has a bug which can cause segfault in rclnodejs.
2. rclnodejs has a bug which causes segfault in node >= v14.
3. rmf has perfornance issues, gh actions runners does not meet the system requirements and so will crash sporadically.

ChawinTan:
Hi @matiasbavera and @koonpeng 

Just finished tidying up. If you are free, please do try it out on your local machine and see if it works. Thanks!

packages/dashboard/e2e/auth-ready.js
```diff
@@ -15,13 +15,70 @@ async function authReady(timeout = 30000) {
     }, timeout);
     let retryTimer;
     const waitAuthReady = () => {
-      req = http.request('http://localhost:8080/auth/', () => {
+      // initialize network and container related variables.
+      let authIpAddress;
+      const commonNetwork = 'auth_dashboard_network';
+      const commonNetworkSubnet = '172.16.0.0/16';
+      const authContainerIp = '172.16.0.2';
```
matiasbavera, line 22: In case we need it to change this, shouldn't be good for this to be on an `env var`.  Maybe you can set them on this file `packages/dashboard/e2e/env.sh` too and replace

`export REACT_APP_AUTH_CONFIG='{"realm":"master", "clientId":"romi-dashboard", "url":"http://172.16.0.2:8080/auth"}'`

with 

`export REACT_APP_AUTH_CONFIG='{"realm":"master", "clientId":"romi-dashboard", "url":"http://$authContainerIp:8080/auth"}'`

So we will be defining the IP just in one place

packages/dashboard/e2e/auth-ready.js
```diff
@@ -15,13 +15,70 @@ async function authReady(timeout = 30000) {
...
+
+      // check if we are in github CI environment
+      if (process.env.CI) {
+        // check if auth container has already been initialized
+        let authContainer = execSync(
```
matiasbavera, line 27: It doesn't seem to be reassigning a value to this. Maybe const?

packages/dashboard/e2e/auth-ready.js
```diff
@@ -15,13 +15,70 @@ async function authReady(timeout = 30000) {
...
+          'docker ps -q --filter ancestor=romi-dashboard/auth',
+        ).toString();
+        if (authContainer) {
+          process.env.AUTH_CONTAINER = authContainer;
+
+          // check if auth container is already connected to custom network to prevent running the same operation if it has
```
matiasbavera, line 33: nit: if it has...

Maybe if it has is a leftover or I think something is missing here.

packages/dashboard/e2e/auth-ready.js
matiasbavera: Are you having an error if you use const for the `execSync().toString()`?

packages/dashboard/e2e/auth-ready.js
```diff
@@ -15,13 +15,70 @@ async function authReady(timeout = 30000) {
...
+          let isConnected = execSync(
+            `docker ps -q --filter network=${commonNetwork} --filter ancestor=romi-dashboard/auth`,
+          ).toString();
+
+          if (!isConnected) {
+            // find and set default auth network id
+            let defaultAuthNetwork = execSync(
```
matiasbavera, line 40: Same comment/question about the const above

packages/dashboard/e2e/auth-ready.js
```diff
@@ -15,13 +15,70 @@ async function authReady(timeout = 30000) {
...
+              "docker inspect $AUTH_CONTAINER -f '{{range.NetworkSettings.Networks }}{{.NetworkID}}{{end}}'",
+            ).toString();
+            process.env.DEFAULT_AUTH_NETWORK = defaultAuthNetwork;
+
+            // create common network with 172.16.xx.x subnet and connect the auth and dashboard container
+            // We need to specify a subnet to assign an ip address later
+            // use a 172.16 range ip as it is private, like localhost but is not blocked by github's environment
```
matiasbavera, line 47: nit 172.16.0.0/16

matiasbavera (commented):
Thanks for the changes! Left a couple of comments. 

Also, I tested on my local env, it seems to be working fine (accessing the login screen and login to the app). But I have some errors on the e2e tests as you and KP described before.

packages/dashboard/e2e/env.sh
```diff
@@ -5,3 +5,9 @@ export REACT_APP_TRAJECTORY_SERVER=${REACT_APP_TRAJECTORY_SERVER:-ws://localhost
 export REACT_APP_API_SERVER=${REACT_APP_API_SERVER:-ws://localhost:50002}
 export ROMI_DASHBOARD_PORT=${ROMI_DASHBOARD_PORT:-5000}
 export COMPOSE_PROJECT_NAME=${COMPOSE_PROJECT_NAME:-romidashboarde2e}
+
+if [ -z "${CI}" ]; then
```
koonpeng, line 9: One possible use of `CI` locally is to run the tests in headless mode.

If using proposal 1, we should be able to use the same logic on both gh and local runs. If using proposal 2, I think it's better to use `GITHUB_NETWORK` to check the environment.

koonpeng (changes requested):

packages/dashboard/e2e/wdio.conf.js
```diff
@@ -124,7 +123,14 @@ exports.config = {
   // Services take over a specific job you don't want to take care of. They enhance
   // your test setup with almost no effort. Unlike plugins, they don't add new
   // commands. Instead, they hook themselves up into the test process.
-  services: ['chromedriver'],
+  services: [
+    [
+      'chromedriver',
+      {
+        args: ['--whitelisted-ips='],
```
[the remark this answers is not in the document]
ChawinTan: I wanted to get rid of this error:

`bind() failed: Cannot assign requested address (99)`

but it seems like the webdriver is just looking out for ipv6 addresses and then defaulting to ip4 when it can't find it. It doesn't affect the tests.

packages/dashboard/e2e/env.sh
ChawinTan: Are we going to run the tests in headless mode locally?

ChawinTan:
> Also, thanks for the detailed documentation. Maybe you can create a README with this on the `e2e` folder. Also, you can add your graphics from draw.io to that README.

Will need to change the diagram with the recent changes

> We should not hardcode the IP address, it may not work on different systems.

> Instead I have 2 proposal,
> 
> Run the auth in a known network (can configure it in docker-compose.yml) and we find the ip from docker inspect <network-name>. We should be using the gateway ip instead of the container ip, docker does NAT to route that to the container.
> 
> Instead of switching the network, launch the auth container in the same network as github actions container from the start. Since we export GITHUB_NETWORK, we can use it to set the network docker-compose uses, see https://docs.docker.com/compose/environment-variables/. In this case, we should use the container ip address.
> 
> We might need to change the test script to launch the containrs or create the network before building.

Am I right to say that for method 1, the gateway ip of the known network sort of act as the router to interface between romi container and auth container and we don't have to manually connect them to the same network anymore?

So for proposal 2, because the 2 containers are started in the same network, there is no need for NAT?

Anyways, I went ahead with option 1 as that seems more straightorward and so far it's working well

matiasbavera:
> Am I right to say that for method 1, the gateway ip of the known network sort of act as the router to interface between romi container and auth container and we don't have to manually connect them to the same network anymore?

Maybe I'm wrong but I think it's more like the gateway is bridging the networks.

packages/dashboard/e2e/auth-ready.js
matiasbavera: Outdated

koonpeng:
> Am I right to say that for method 1, the gateway ip of the known network sort of act as the router to interface between romi container and auth container and we don't have to manually connect them to the same network anymore?

In general, yes. When you are in the container `172.16.0.2/16` and you try to connect to `172.17.0.1/16`, linux sees that it is on a different network so it looks at its routing table to decide where to send it to. It probably sends it to `172.16.0.1/16` which is the docker bridge. The `172.16.x.y` bridge then routes it to the `172.17.x.y` bridge. But I don't know why you can't connect to the container directly, maybe you need to configure docker to do so. In any case, the docker bridge performs NAT and we exposed port 8080 in docker-compose so it forwards it to the container.

> So for proposal 2, because the 2 containers are started in the same network, there is no need for NAT?
Yes, because they are in the same network, linux uses arp to get the mac address and connects to it directly.

packages/dashboard/e2e/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Container and network interactions
+
+Below is a diagram of how the containers and networks interact when up and running
+
+![Containers and network interaction diagram](documentation-utils/e2e-Container-networks.png)
+
+Broadly speaking, there are 3 major parts of the system:
+
+- Romidashboard container. It holds the API server and the React dashboard
```
matiasbavera, line 22: I think we should mention that it has the RMF backend as well

packages/dashboard/e2e/README.md
```diff
@@ -0,0 +1,36 @@
...
+- Auth container
+- auth_network, a known network where the Auth container is reachable 
+
+Requests from the romidashboard to the auth container would be routed to the gateway ip address of `auth_network`. Docker would then perform a NAT to route the requests to the auth container and vice versa.
+
+The dashboard would need to be compiled with the gateway ip address of the `auth_network`. This is done by retrieving the ip address the moment the network is created and then have it set in the `url` paramenter of the `REACT_APP_AUTH_CONFIG` env variable.
```
matiasbavera, line 28: nit: compiled -> built?

matiasbavera (approved):
LGTM, but let's wait for KP approval

packages/dashboard/e2e/README.md
```diff
@@ -0,0 +1,36 @@
...
+This is to document the flow and interaction of the e2e services in the github environment when:
+
+- starting up the services
+- running the tests
+
+## Starting up the services
+
+Below is a diagram representing the flow of commands when running `npm run test:e2e`
+
+![Flow of commands diagram](documentation-utils/e2e-E2e-start-process.png)
```
koonpeng, line 12: Might want to change to diagram to show that `npm run test:e2e` runs `source env.sh` -> `npm run setup`... sequentially. The diagram seems to suggest they are ran in parallel.

packages/dashboard/e2e/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Container and network interactions
+
+Below is a diagram of how the containers and networks interact when up and running
+
+![Containers and network interaction diagram](documentation-utils/e2e-Container-networks.png)
```
koonpeng, line 18: Can we use a different background for the container boxes? Don't know if its my screen but the dashboard container particularly is very hard to read.

packages/dashboard/e2e/scripts/test-e2e.js
```diff
@@ -1,10 +1,26 @@
 const concurrently = require('concurrently');
 const { execSync } = require('child_process');
 
+// create external network for auth container
+execSync('docker network create auth_network', { stdio: 'inherit' });
```
koonpeng, line 5: see above on network name

koonpeng (changes requested):
nit: move these to a more fitting path https://github.com/osrf/romi-dashboard/tree/fix/e2e-container-communications/packages/dashboard/e2e/documentation-utils

suggestion: `dashboard/e2e/docs/resources`

ChawinTan:
Changed the auth_container's network name, modified the diagrams in README and added a check to remove romi_dashboard_e2e_network prior to starting the subsequent 2 runs

koonpeng (changes requested):
looking good! Can you also update `nightly.yml` with the new changes and it should be good to merge.

koonpeng (approved):
:+1:

PR #152, opened 2020-10-23
refactor: component lib

koonpeng:
## What's new

This is a ongoing work of refactoring many of the existing components into a standalone library. The goal is to reduce coupling, improve code reuse and better code design in general. This initial work ported the following components

- OmniPanel
- DoorItem (renamed to DoorAccordion)
- LiftItem (renamed to LiftAccordion)
- DispenserItem (renamed to DispenserAccordion)
- RobotItem (renamed to RobotAccordion)
- LoopRequestForm
- DeliveryRequestForm
- Door (renamed to DoorMarker)

And also "sub components" of the above components (e.g. lift info). They are not used in the dashboard yet, I would like to get a review and only start refactoring the dashboard after these looks good.

## Self-checks

## Discussion

[Questions for reviewers]


packages/react-components/lib/doors/door-marker.tsx
```diff
@@ -0,0 +1,222 @@
...
+ *  - range of motion is entire distance from v2->v1. No need to specify.
+ */
+const SingleSlidingDoor = SingleSwingDoor;
+
+/*
+ * single/double telescoping doors:
+ *   * common in elevators; same parameters as sliding doors; they just
+ *     open/close faster and take up less space inside the wall.
+ */
+const SingleTelescopeDoor = SingleSwingDoor;
```
matiasbavera, line 140: I think we are pointing to the same reference here

packages/react-components/lib/lifts/tests/lift-accordion.test.tsx
```diff
@@ -0,0 +1,8 @@
+import { render } from '@testing-library/react';
+import React from 'react';
+import { LiftAccordion } from '..';
+import { makeLift } from './test-utils';
+
+test(`smoke test`, () => {
```
matiasbavera, line 6: Maybe we can add a snapshot too.

packages/react-components/lib/robots/tests/robot-accordion.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+  name: 'test',
+  battery_percent: 1,
+  location: { level_name: 'test_level', x: 0, y: 0, yaw: 0, t: { sec: 0, nanosec: 0 } },
+  mode: { mode: RomiCore.RobotMode.MODE_PAUSED },
+  model: 'test_model',
+  task_id: 'test_task_id',
+  path: [],
+};
+
+test('smoke test', () => {
```
matiasbavera, line 16: maybe we can add a snapshot test too

matiasbavera (commented):
Awesome, it looks cleaner. Left a couple of questions.

packages/react-components/package.json
```diff
@@ -0,0 +1,52 @@
...
+    "@testing-library/user-event": "^12.1.9",
+    "@types/jest": "^26.0.14",
+    "@typescript-eslint/eslint-plugin": "^4.4.0",
+    "@typescript-eslint/parser": "^4.4.0",
+    "babel-loader": "^8.1.0",
+    "eslint": "^7.11.0",
+    "eslint-plugin-react": "^7.21.4",
+    "eslint-plugin-react-hooks": "^4.1.2",
+    "jest": "^26.5.2",
+    "react": "^16.13.1",
```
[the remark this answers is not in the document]
koonpeng: We can't/shound't mix react versions so a component library should use the dependant's react. See material-ui's [package-json](https://github.com/mui-org/material-ui/blob/next/packages/material-ui/package.json) for example.

packages/react-components/examples/simple-omni-panel.tsx
```diff
@@ -0,0 +1,52 @@
...
+      <OmniPanelView viewId={0}>
+        <Button variant="outlined" onClick={() => (stack.push(1), setView(1))}>
+          Panel A
+        </Button>
+        <Button variant="outlined" onClick={() => (stack.push(2), setView(2))}>
+          Panel B
+        </Button>
+        <Button variant="outlined" onClick={() => (stack.push(3), setView(3))}>
+          Panel C
+        </Button>
```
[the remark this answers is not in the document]
koonpeng: The omnipanel does not control the view of each panel, it's like this simply because it's the default. This is just use as an example/test of the omnipanel so I don't think it's important to style it nicely.

packages/react-components/lib/commands/tests/loop-request-form.test.tsx
```diff
@@ -0,0 +1,73 @@
+import { render, screen, within } from '@testing-library/react';
```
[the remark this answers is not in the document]
koonpeng: I don't really have a preference actually, I just used testing-library to give it a try. Since this is in a different package it doesn't affect the dashboard, maybe we can look at the pros and cons of each libraries and standardize which to use.

packages/react-components/lib/doors/door-marker.tsx
koonpeng: Good catch, it should be referencing to sliding doors.

packages/react-components/lib/lifts/tests/lift-accordion.test.tsx
koonpeng: I did do snapshot testing at first, but material-ui is not designed with snapshot testing in mind and they do not recommend it. You would need to go through some hacks to make the components stable under snapshots, I think it's probably not worth the trouble.

packages/react-components/lib/robots/tests/robot-accordion.test.tsx
koonpeng: I did do snapshot testing at first, but material-ui is not designed with snapshot testing in mind and they do not recommend it. You would need to go through some hacks to make the components stable under snapshots, I think it's probably not worth the trouble.

packages/react-components/stories/lift-marker.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+        v1_y: 0,
+        v2_x: 0,
+        v2_y: -1.414,
+      },
+    ],
+  }),
+  makeLiftState(),
+);
+
+// FIXME: Broken
```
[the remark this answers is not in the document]
koonpeng: I think it's fine, this is an ongoing effort to isolate the components we are currently using, I intend to continue working on this. This PR is for a early review before we replace the current components with the ones here.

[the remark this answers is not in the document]
matiasbavera: Just an idea, since they are independent packages, and they are not imported yet, would not it be good to merge at least one base structure already?, Then new reusable components can be created there too.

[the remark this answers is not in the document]
koonpeng: Yeah, thats the goal of this PR. If it looks good then we can merge it and start replacing the components in dashboard. I'm sure there would be issues when we actually try to replace the components which might need some refactor on the components here, they are not meant to be final so I meant for this to be an early preliminary review to see if things make sense before merging.

packages/react-components/lib/robots/robot-marker.tsx
```diff
@@ -0,0 +1,77 @@
...
+  robot: RomiCore.RobotState;
+  footprint: number;
+  fleetName: string;
+  iconPath?: string;
+  variant?: 'normal' | 'inConflict';
+}
+
+/**
+ * Contexts: ColorContext
+ */
```
matiasbavera, line 39: Are you putting this to know that this component is within the scope of the ColorContext context to use it in the DefaultMarker? In other words, it is required id you want to use the `DefaultMarker` component, right?

matiasbavera (approved):
LGTM, left one comment. I think we can proceed as we discuss here https://github.com/osrf/romi-dashboard/pull/152#discussion_r512391119. As you mentioned to be able to find, solve and manage problems before the PR becomes massive

packages/react-components/lib/robots/robot-marker.tsx
koonpeng: Yeah, but it is not required, if the context is not given it will use the default `ColorManager`. The context is to allow consumers to override the default colors with their own implementation.

PR #154, opened 2020-10-27
Feature/tooltips

vallq:
## What's new
- Adding tooltips in place of tutorial

tooltips
- use of local storage to keep the state of the tooltips when romi is loaded

## Self-checks

## Discussion
I'm not able to add tooltips to the individual panels / leaflet components. using the tool tip from leaflet's library requires a map pane too, so I think only the robots/doors/lifts overlays can have them. 
would the current additions be enough? I'm uncertain of where else we could add them too
@matiasbavera @koonpeng


vallq:
> You mean you are not able to set a tooltip in an element on the OmniPanel? Maybe it's something related to the z component of the CSS.
yea it doesnt show up 

> what do you think about adding a question icon here and explain a bit, what does each command does here
Oh yea I think that would be good,
like a (?) beside each request type and hovering over that to get the tooltip
Then I think that doesn't need to be disabled

koonpeng:
> For discussion, what do you think about adding a question icon here and explain a bit, what does each command does here 
>
> image
>
> We can also add tooltips to describe what the clear, reset, and restore icons does on the negotiation Panel.

Be sure to check the the new `react-components` package. You might want to implement it there instead.

matiasbavera:
> what do you think about adding a question icon here and explain a bit, what does each command does here
> Oh yea I think that would be good,
> like a (?) beside each request type and hovering over that to get the tooltip
> Then I think that doesn't need to be disabled

Yeah, you're right

vallq:
final_5f9ab81cf7f36c0109d4dce3_983656

Oh I mean over the main menu panel but I realised I was doing it wrongly haha. here are the additional tooltips I added in!

packages/dashboard/src/components/tests/appbar.test.tsx
```diff
@@ -26,32 +28,49 @@ describe('AppBar', () => {
...
+    expect(root.find('#setting-tooltip').length > 0).toBeTruthy();
+    expect(root.find('#help-tooltip').length > 0).toBeTruthy();
+  });
+
+  test('does not render tooltips when disabled', () => {
+    tooltips = false;
+    const root = buildWrapper(toggleShowOmniPanel, showSettings, showHelp, tooltips);
+    expect(root.find('#omnipanel-tooltip').length <= 0).toBeTruthy();
+    expect(root.find('#setting-tooltip').length <= 0).toBeTruthy();
+    expect(root.find('#help-tooltip').length <= 0).toBeTruthy();
```
matiasbavera, line 57: Do you think it'll be possible to use `exists() `here?

packages/dashboard/src/components/appbar.tsx
```diff
@@ -18,13 +19,14 @@ export interface AppBarProps {
   toggleShowOmniPanel(): void;
   showSettings(show: boolean): void;
   showHelp(show: boolean): void;
+  tooltips?: boolean;
```
matiasbavera, line 22: What do you think about setting this tooltip as a context instead of a prop?

packages/dashboard/src/components/appbar.tsx
```diff
@@ -40,51 +42,129 @@ export default function AppBar(props: AppBarProps): React.ReactElement {
...
-        <IconButton id="toggle-omnipanel-btn" color="inherit" onClick={() => toggleShowOmniPanel()}>
-          <DashboardIcon />
-        </IconButton>
-
-        <IconButton id="show-settings-btn" color="inherit" onClick={() => showSettings(true)}>
-          <SettingsIcon />
-        </IconButton>
-        {user && (
-          <>
+          <Tooltip
```
matiasbavera, line 51: I think we can improve this solution, maybe we can create our own Tooltip component that receives the tooltip state and decides if It'll return just the child or the child with the tooltip wrapped. E.g:
```
<CustomTooltip tooltip=true>
    <ChildComponent>
</CustonTooltip>
```

returns:

```
<Tooltip>
    <ChildComponent>
</Tooltip>
```


```
<CustomTooltip tooltip=false>
    <ChildComponent>
</CustonTooltip>
```

returns:

```

    <ChildComponent>

```

packages/dashboard/src/components/dashboard.tsx
```diff
@@ -430,9 +427,10 @@ export default function Dashboard(_props: {}): React.ReactElement {
               toggleShowOmniPanel={() => setShowOmniPanel(!showOmniPanel)}
               showSettings={setShowSettings}
               showHelp={setShowHelp}
+              tooltips={tooltipState}
```
matiasbavera, line 430: What do you think about setting a React.context for this?

packages/dashboard/src/components/main-menu.tsx
```diff
@@ -11,47 +11,83 @@ export interface MainMenuProps {
...
         <Typography variant="h5">Dispensers</Typography>
       </ListItem>
-
       <Divider />
 
-      <ListItem data-item="Commands" button={true} onClick={props.onCommandsClick}>
-        <Typography variant="h5">Commands</Typography>
-      </ListItem>
-      <Divider />
+      {!tooltips && (
```
matiasbavera, line 44: Same comment about the `CustomTooltip` component

matiasbavera (changes requested):

packages/dashboard/src/components/appbar.tsx
vallq: Ok let me work on this!

packages/dashboard/src/components/tests/appbar.test.tsx
vallq: yep this would work too!

packages/dashboard/src/components/appbar.tsx
vallq: I could give this a shot, the implementation does bother me too ((':

vallq:
Changes:
- Custom tooltip
- Using context for tooltip
- Updating test files

packages/dashboard/src/components/ui-contexts.tsx
```diff
@@ -0,0 +1,6 @@
+import React from 'react';
+
+export const TooltipContext = React.createContext({
```
matiasbavera, line 3: Do you think that the tooltip is not in the same category as the `notificationBar` on the file `app-context.tsx`

packages/dashboard/src/components/tooltips.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
```
matiasbavera, line 1: Let's use the singular for this component to keep name consistency. `tooltips` -> `tooltip`

packages/dashboard/src/components/tests/tooltips.test.tsx
```diff
@@ -0,0 +1,18 @@
...
+  ReactDOM.render(
+    <DashboardTooltip title="test" id="test-tooltip" enabled={true}>
+      <IconButton>
+        <SettingsIcon />
+      </IconButton>
+    </DashboardTooltip>,
+    div,
+  );
+  ReactDOM.unmountComponentAtNode(div);
+});
```
matiasbavera, line 18: I'll be good to have a snapshot and a `enabled={false}` test too

matiasbavera (commented):
Thanks for the changes Val, left a couple of comments

packages/dashboard/src/components/commands-panel.tsx
```diff
@@ -156,6 +157,13 @@ export const CommandsPanel = React.memo((props: CommandsPanelProps) => {
               expandIcon={<ExpandMoreIcon />}
             >
               <Typography variant="h5">Loop Request</Typography>
+              <Tooltip
+                title="send a request for target fleet to perform tasks repeatedly"
```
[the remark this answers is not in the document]
vallq: oh okay, sure! how about the definition for the dispenser request? is that alright?

packages/dashboard/src/components/ui-contexts.tsx
vallq: hm I put it in a separate category because to me the notifications bar is a result of interaction with RMF (by displaying whenever there is a conflict) and tooltips are purely UI. but we could move it into the `app-context` file, and also include it in the app context provider as per your prev comment!

packages/dashboard/src/components/tests/tooltips.test.tsx
vallq: okay, will include that too!

matiasbavera:
Hey @koonpeng  since we are using `react-testing-library` on react-components, should we start doing the new tests of the RomiDashboard with `react-testing-library` or at least the test of components we should move to the react component library?

packages/dashboard/src/components/ui-contexts.tsx
matiasbavera: Yeah, the notification bar, in theory, can be used by any component, it's for display a message of error or success, so it just not for the RMF it's a pure UI component. So, it can with the app context too. =)

koonpeng:
If it's not too much work, can you move the tooltip component into `react-components`? We could also make it "managed", meaning the component itself handles showing/hiding the tooltip on mouse hover etc.

As for the tests, which library does everyone prefer? After trying out `test-library` I personally prefer that, enzyme feels like too much unneeded complexity, espically with how its selectors only targets certain elements, making it incredibly tricky to use (afaik it differentiate between "react" and "native" DOM elements).

But I don't mind using enzyme as well with others prefer it, we should vote and decide which is the primary library to use.

matiasbavera:
> As for the tests, which library does everyone prefer? After trying out `testing-library` I personally prefer that, enzyme feels like too much unneeded complexity, espically with how its selectors only targets certain elements making it incredibly tricky to use (afaik it differentiate between "react" and "native" DOM elements).
> 
> But I don't mind using enzyme as well if others prefer it, we should vote and decide which is the primary library to use.

I understand, Ok I created this poll https://github.com/osrf/romi-dashboard/issues/158

> If it's not too much work, can you move the tooltip component into `react-components`? We could also make it "managed", meaning the component itself handles showing/hiding the tooltip on mouse hover etc.

Makes sense

vallq:
tooltips
this took longer than expected (': but I've managed to add the tooltip to the stories! the default will render the tooltip when it is hovered over, I could add other examples but I am not sure if those would be necessary. let me know your comments!

as for tests, I'm more comfortable with react testing library!

matiasbavera:
> this took longer than expected (': but I've managed to add the tooltip to the stories! the default will render the tooltip when it is hovered over, I could add other examples but I am not sure if those would be necessary. let me know your comments!

I think this is fine

> as for tests, I'm more comfortable with react testing library!

cool!

packages/dashboard/src/components/tests/tooltip.test.tsx
```diff
@@ -0,0 +1,48 @@
+import React from 'react';
```
matiasbavera, line 1: If we use enable on the `react-components` package we should move those tests there

packages/dashboard/src/components/tests/__snapshots__/tooltip.test.tsx.snap
```diff
@@ -0,0 +1,19 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
matiasbavera, line 1: If we use enable on the `react-components` package we should delete this because AFAIK we are not  using snapshots con the `react-components` package

matiasbavera (commented):
Thanks for the changes Val, left a couple of comments

packages/react-components/tests/tooltip.test.tsx
```diff
@@ -0,0 +1,29 @@
...
+      <DashboardTooltip title="test" id="test" enabled={true}>
+        <Typography variant="h5">Tooltip is enabled</Typography>
+      </DashboardTooltip>,
+    );
+
+    fireEvent.mouseEnter(screen.getByRole('heading'));
+
+    expect(screen.getByTestId('test-tooltip')).toBeTruthy();
+  });
+  it('does not render the tooltip when disabled', () => {
```
matiasbavera, line 18: nit: space between functions

matiasbavera (approved):
Thanks for this Val! Once that comment is addressed you can merge it =)

packages/react-components/tests/tooltip.test.tsx
vallq: okay! thanks @matiasbavera

PR #156, opened 2020-11-02
Reducer refactor

matiasbavera:
## What's new

[Describe your changes]

## Self-checks

## Discussion

[Questions for reviewers]

### DIFF
--- a/packages/dashboard/src/components/appbar.tsx
+++ b/packages/dashboard/src/components/appbar.tsx
@@ -29,10 +29,6 @@ export default function AppBar(props: AppBarProps): React.ReactElement {
   const authenticator = React.useContext(AuthenticatorContext);
   const user = React.useContext(UserContext);
 
-  // toggleShowOmniPanel={() => setShowOmniPanel(!showOmniPanel)}
-  // showSettings={setShowSettings}
-  // showHelp={setShowHelp}
-
   async function handleLogout(): Promise<void> {
     try {
       await authenticator.logout();
--- a/packages/dashboard/src/components/appbar.tsx
+++ b/packages/dashboard/src/components/appbar.tsx
@@ -13,22 +13,25 @@ import SettingsIcon from '@material-ui/icons/Settings';
 import React from 'react';
 import { AuthenticatorContext, UserContext } from './auth/contexts';
 import HelpIcon from '@material-ui/icons/Help';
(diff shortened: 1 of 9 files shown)


matiasbavera:
Hi @koonpeng, this is not even remotely for review, still WIP, but I'll like to get your opinion on this refactor before I continue with it. 

In this PR I am implementing the `React.useReducer` to decouple the states from the Dashboard file. The result of this is that we only do `prop drilling` of a` dispacher` (with which the states can be changed) instead of several handlers and this makes the components handle their logic and can change the global state. This is good and bad depending on where you look at it, by doing this you are attaching the component so that it works as the central state is (it is possible to test this component because you only have to mock up the `reducer` functions). The good thing is that this will force us to have two types of components, the reusable ones that will eventually have to go to the `react-component` package and those that are not reusable and have specific logic like ʻAppBar`

In this case, I am using two large states, `stateDashboard` and `stateMenu`, but we can make this a large state with state dictionaries inside it, with a central trigger. This is basically what Redux does and we will only have one source of truth.

koonpeng:
I can see how it will make some of the components much less complex, as mentioned the main downside is that they will not be reusable and their functions are tied to the dashboard.

I often have the dilemna of making a component simpler/easier, at the cost of reducing its reusability. I guess this is a shortcoming of react and it makes sense to separate components into different categories. The reusable ones will be in `react-components` and those specific to the webapp will be in `dashboard`.

Now that we are moving reusable components into its own package, I think it makes sense for the design goal of the non-reusable components (appbar, settings etc) to be more simple/flat (less prop passing/drilling). They are not meant to be reusable anyway so it should be ok to tie application specific logic into it.

matiasbavera:
> I can see how it will make some of the components much less complex, as mentioned the main downside is that they will not be reusable and their functions are tied to the dashboard.
> 
> I often have the dilemna of making a component simpler/easier, at the cost of reducing its reusability. I guess this is a shortcoming of react and it makes sense to separate components into different categories. The reusable ones will be in `react-components` and those specific to the webapp will be in `dashboard`.
> 
> Now that we are moving reusable components into its own package, I think it makes sense for the design goal of the non-reusable components (appbar, settings etc) to be more simple/flat (less prop passing/drilling). They are not meant to be reusable anyway so it should be ok to tie application specific logic into it.

Awesome! I'm glad to hear we are on the same page.

PR #161, opened 2020-11-09
Refactor: move task components to react-components

matiasbavera:
## What's new

* Removed snapshot test from TaskSummary
* Moved trashbin-button-group to react-components package
* Add react-testing-library tests to the trashbin-button-group
* Using  `SimpleInfo, SimpleInfoData` on `TaskSummaryPanel` component
* Created a general reducer for restore and clear actions (in this case it is used on snapshot-button-group) 
* Added React-Testing-Library to the Dashboard Package
* Extended simpleInfo props to support class overriding
* Added fixes after react-component migration merge
* Added e2e test

## Self-checks

- [X] I tested the behavior of the components that interact with the backend, with an e2e test


packages/dashboard/src/components/tests/task-summary-panel.test.tsx
```diff
@@ -116,3 +116,38 @@ test('Get name of the actor from status', () => {
...
+  beforeEach(() => {
+    task = Object.values(fakeTaskSummary())[0];
+  });
+
+  test('Shows ACTIVE label', () => {
+    task.state = RomiCore.TaskSummary.STATE_ACTIVE;
+    const root = mount(<TaskSummaryPanelInfo task={task} />);
+    expect(root.html()).toContain('ACTIVE');
+    root.unmount();
+  });
```
matiasbavera, line 131: Should we start doing this with react-testing-library? 

I think yes, we can install the library on this PR as well and start to refactor that too.

packages/dashboard/src/components/task-summary-panel.tsx
```diff
@@ -24,6 +23,45 @@ export const getActorFromStatus = (status: string) => {
...
+      return 'COMPLETED';
+    case RomiCore.TaskSummary.STATE_FAILED:
+      return 'FAILED';
+    default:
+      return 'UNKNOWN';
+  }
+};
+
+interface TaskSummaryPanelInfoProps {
+  task: RomiCore.TaskSummary;
```
koonpeng, line 46: The only prop is the list of task, we should be able to migrate this to `react-components` as well.

While we are at it, we should also move the save/restore functions out into a HOC or hook. Try and see if you can make it more generic, what we need is a way to
  * save the current state
  * restore the saved state
  * merge the restored state with the current state
  * clear the saved state
I think it should be possible to write a generic hook to handle that.

This component should only handle rendering a list of tasks, the task panel in `dashboard` will then use this component, along with the HOC/hook to handle the saving/clearing logic.

packages/react-components/lib/tree-button-group.tsx
```diff
@@ -55,3 +55,5 @@ export const TreeButtonGroup = (props: TreeButtonGroupProps) => {
     </div>
   );
 };
+
+export default TreeButtonGroup;
```
koonpeng, line 59: The buttons are pretty explicit so I wouldn't call them something generic like `TreeButtonGroup`. Since it deals with managing "snapshots", we could name it something like `SnapshotControlButtonGroup`.

Other things to consider.
* Forwarding refs.
* Extending and forwarding other props to the root div.
* rename `handle*` to `on*`, following conventions.
* add a `save` button and a prop to control which buttons are shown.

packages/react-components/stories/tree-button-group.stories.tsx
```diff
@@ -0,0 +1,12 @@
+import { Story, Meta } from '@storybook/react';
+import React from 'react';
+import { TreeButtonGroup } from '../lib';
+
+export default {
+  title: 'Tree Button Group',
+  component: TreeButtonGroup,
+} as Meta;
+
+export const SimpleTreeButtonGroup = (): React.ReactElement => {
```
koonpeng, line 10: Make use of storybook's control addon to provide a UI to adjust the props. See https://storybook.js.org/docs/react/essentials/controls.

packages/react-components/tests/tree-button-group.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const handleRestoreClick = jest.fn();
+  const root = render(
+    <TreeButtonGroup
+      handleResetClick={handleResetClick}
+      handleClearClick={handleClearClick}
+      handleRestoreClick={handleRestoreClick}
+    />,
+  );
+
+  const clearButton = root.container.querySelector('button#clear-button');
```
koonpeng, line 32: RTL encourages use of the provided selectors when possible, in this case we could use `getByText`.

packages/react-components/tests/tree-button-group.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const resetButton = root.container.querySelector('button#reset-button');
+  const restoreButton = root.container.querySelector('button#restore-button');
+
+  // FireEvent should not receive null
+  clearButton && fireEvent.click(clearButton);
```
koonpeng, line 37: use `userEvents.click` instead.

koonpeng (changes requested):

packages/react-components/tests/tree-button-group.test.tsx
matiasbavera: Makes sense! Thanks!

packages/react-components/tests/tree-button-group.test.tsx
matiasbavera: Yeah, you're right! Thanks for the comment!

packages/dashboard/src/components/tests/task-summary-panel.test.tsx
matiasbavera: https://github.com/osrf/romi-dashboard/issues/163

matiasbavera:
Left to do: 
* Order the tasks by status
* No custom hook/HOC seen that general actions + useReducer is the way to go

packages/react-components/lib/snapshot-control-button-group-reducer.ts
```diff
@@ -0,0 +1,57 @@
...
+  const newContents = Object.assign({}, currentContent);
+  Object.keys(storedContent).forEach((element) => {
+    if (replaceCurrentContent) newContents[element] = storedContent[element];
+    else {
+      // If the element is already on the currentContent do not replace it
+      if (!(element in currentContent)) newContents[element] = storedContent[element];
+    }
+  });
+  return newContents;
+}
```
matiasbavera, line 39: Since this is a general state or reducer, should we keep it with `any`? or we should add the possible types with `|`?

packages/react-components/lib/task-summary/task-summary-accordion.tsx
```diff
@@ -0,0 +1,180 @@
...
+import { TreeItem, TreeView } from '@material-ui/lab';
+import ExpandMoreIcon from '@material-ui/icons/ExpandMore';
+import ChevronRightIcon from '@material-ui/icons/ChevronRight';
+import PlayCircleFilledWhiteIcon from '@material-ui/icons/PlayCircleFilledWhite';
+import PauseCircleFilledIcon from '@material-ui/icons/PauseCircleFilled';
+import { Theme } from '@material-ui/core/styles/createMuiTheme';
+import SimpleInfo, { SimpleInfoData } from '../simple-info';
+import { formatStatus, getActorFromStatus, getStateLabel } from './task-summary-utils';
+
+const debug = Debug('OmniPanel:TaskSummaryAccordion');
```
koonpeng, line 14:
```suggestion
const debug = Debug('Tasks:TaskSummaryAccordion');
```

packages/react-components/lib/task-summary/task-summary-accordion.tsx
```diff
@@ -0,0 +1,180 @@
...
+      case RomiCore.TaskSummary.STATE_ACTIVE:
+        return classes.active;
+      case RomiCore.TaskSummary.STATE_COMPLETED:
+        return classes.completed;
+      case RomiCore.TaskSummary.STATE_FAILED:
+        return classes.failed;
+      default:
+        return 'UNKNOWN';
+    }
+  };
```
koonpeng, line 69: The comment says added useCallback but it is not added?

packages/react-components/lib/task-summary/task-summary-accordion.tsx
```diff
@@ -0,0 +1,180 @@
...
+
+  const renderActor = (taskStatus: string): React.ReactElement | undefined => {
```
koonpeng, line 71: I don't know when to use `React.ReactElement` or `JSX.Element` I guess both is fine, but I think we should return `null` instead of `undefined`.

packages/react-components/lib/task-summary/task-summary-utils.ts
```diff
@@ -0,0 +1,73 @@
...
+      return 'FAILED';
+    default:
+      return 'UNKNOWN';
+  }
+};
+
+/**
+ * Sorts tasks by state so what is active is always at the top of the list.
+ */
+export const sortTasksByState = (
```
koonpeng, line 35: Should we also do a secondary sort based on the submission time?

packages/react-components/tests/task-summary/task-summary-accordion.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+      'Moving [tinyRobot/tinyRobot2]: ( 9.81228 -6.98942 -3.12904) -> ( 6.26403 -3.51569  1.16864) | Remaining phases: 1 | Remaining phases: 6',
+    submission_time: { sec: 0, nanosec: 0 },
+    task_id: 'am8faee9-84ca-41ea-8bb6-8493cc9f8249',
+  },
+};
+
+describe('Renders correctly', () => {
+  let task: RomiCore.TaskSummary;
+  beforeEach(() => {
+    task = Object.assign({}, Object.values(taskSummary)[0]);
```
koonpeng, line 31: This can "taint" the test data if a test modifies an object inside it, e.g.

```ts
it('bad test', () => {
  task.start_time.sec = 10;
});

it('other test', () => {
  expect(task.start_time.sec).toBe(0); // fails
});
```

I recommend to use a function to create a new copy of task for each test.

packages/react-components/tests/task-summary/task-summary-accordion.test.tsx
```diff
@@ -0,0 +1,129 @@
...
+describe('Components gets the correct style on specifics states', () => {
+  let task: RomiCore.TaskSummary;
+  beforeEach(() => {
+    task = Object.assign({}, Object.values(taskSummary)[0]);
+  });
+
+  test('active style is applied ', () => {
+    task.state = RomiCore.TaskSummary.STATE_ACTIVE;
+    const root = render(<TaskSummaryAccordion tasks={[task]} />);
+    expect(root.getByText(task.task_id).parentElement?.className).toContain('makeStyles-active');
```
koonpeng, line 66: I'm not sure which practice to follow. I looked at material-ui's test and there are a lot of tests on the css class like this. But on the other hand RTL does not recommend testing class names because that is not what an user cares about.

Which philosophy do you think makes more sense?

koonpeng (changes requested):

packages/react-components/lib/task-summary/task-summary-accordion.tsx
matiasbavera: Sorry it's a leftover, it is not inside a use effect anymore

packages/react-components/lib/task-summary/task-summary-accordion.tsx
matiasbavera: Cool I think `React.ReactElement` it's ok here

JSX.Element is a ReactElement, with the generic type for props and type being any. It exists, as various libraries can implement JSX in their own way, therefore JSX is a global namespace that then gets set by the library, React sets it like this:

```
declare global {
  namespace JSX {
    interface Element extends React.ReactElement<any, any> { }
  }
}
```


Source https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement

packages/react-components/tests/task-summary/task-summary-accordion.test.tsx
matiasbavera: yeah, that's a better solution, I was assigning another reference on each test. I created a getTaskObject(). Thanks!

packages/react-components/lib/snapshot-control-button-group.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+export interface SnapshotControlButtonGroupProps extends React.HTMLAttributes<HTMLDivElement> {
+  disableReset?: boolean;
+  disableClear?: boolean;
+  disableRestore?: boolean;
+  disableSave?: boolean;
+  showReset?: boolean;
+  showClear?: boolean;
+  showRestore?: boolean;
+  showSave?: boolean;
```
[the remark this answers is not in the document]
matiasbavera: You're right! Thanks!

matiasbavera:
A lot of things broke when making the reducer generic. I'm seeing how to fix it

packages/react-components/tests/task-summary/task-summary-accordion.test.tsx
matiasbavera: Yeah, I think in our case both will help, but we should establish priority. I think we should go with the RTL way as a top priority for all our components. I'm imagining that the RTL way will help a lot with the code refactor, because since we don't have to look for a specific class or id, if we change the id or the component in the refactor and the component looks the same we will not have to maintain (update) that test.

For the material-ui components, If there are components associated with `Material-ui` and getting the values from those components became too convoluted (e.g.: check the background's color in a component with a lot of parents htmlElement.parent().parent().parent().leftchild()....), I think we should go with the material-ui way and check for a css class.

The problem with using both philosophies as I proposed is where to draw a line to say that something is "convoluted" and worth using the css class verification.

packages/react-components/lib/task-summary/task-summary-utils.ts
matiasbavera: Done!

packages/react-components/tests/task-summary/task-summary-accordion.test.tsx
```diff
@@ -0,0 +1,143 @@
...
+  expect(tasks[0].state).toBe(RomiCore.TaskSummary.STATE_ACTIVE);
+  expect(tasks[1].state).toBe(RomiCore.TaskSummary.STATE_QUEUED);
+});
+
+test('Sorts task by submission time correctly', () => {
+  const tasks = Object.values(getTaskObject());
+  const sortedTasks = sortTasksBySubmissionTime(tasks);
+  expect(sortedTasks[0].submission_time.nanosec).toBe(1519129858900);
+  expect(sortedTasks[1].submission_time.nanosec).toBe(1519129853500);
+});
```
koonpeng, line 143: Should we add a test to sort by multiple values?

Task 1: Active, time 500
Task 2: Active, time 1000
Task 3: Queued, time 1500

Expected result, [ Task2, Task1, Task3 ];

koonpeng (changes requested):

packages/react-components/lib/snapshot-control-button-group-reducer.ts
```diff
@@ -0,0 +1,65 @@
...
+  state: SnapshotState<T>,
+  action: SnapshotAction<T>,
+): SnapshotState<T> {
+  switch (action.type) {
+    case SnapshotActionType.Clear:
+      action.payload.current = Object.assign({}, state[SnapshotStateType.Content]);
+      return { ...state, [SnapshotStateType.Content]: {} as T };
+    case SnapshotActionType.Restore:
+      const mergedContent = mergeContent(state.content, action.payload.current, false);
+      const newState = { ...state, [SnapshotStateType.Content]: mergedContent };
```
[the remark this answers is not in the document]
matiasbavera: Solved, created a function to handle the logic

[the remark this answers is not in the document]
matiasbavera: Thanks for adding the linter to the build

packages/react-components/lib/snapshot-control-button-group.tsx
```diff
@@ -0,0 +1,93 @@
...
+import RestoreFromTrashIcon from '@material-ui/icons/RestoreFromTrash';
+import SaveIcon from '@material-ui/icons/Save';
+import React from 'react';
+
+export interface SnapshotControlButtonGroupProps extends React.HTMLAttributes<HTMLDivElement> {
+  disableReset?: boolean;
+  disableClear?: boolean;
+  disableRestore?: boolean;
+  disableSave?: boolean;
+  showReset?: boolean; // default value = true
```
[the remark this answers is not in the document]
matiasbavera: I used `jsdoc`

packages/react-components/tests/task-summary/task-summary-accordion.test.tsx
matiasbavera: Added

packages/dashboard/src/components/task-summary-panel.tsx
```diff
@@ -1,247 +1,85 @@
...
-    setSelected(nodeIds);
+  const initialValues = {
+    [SnapshotStateType.Content]: {},
   };
 
-  const handleResetTasks = () => {
-    setExpanded([]);
-    setSelected('');
-  };
+  const [stateSnapshot, dispatchSnapshot] = React.useReducer(snapshotReducer, initialValues);
```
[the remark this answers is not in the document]
matiasbavera: Thanks for those examples, it really helped me, sadly, none of them worked, it returned `Record<string, unknown>` as a type. I found a related workaround though, wrap the `React.useReducer` and explicitly assign the types on the state and dispatcher

```
export function useSnapshotReducer<T>(
  initialValues: SnapshotState<Record<string, T>>,
): useSnapshotReducerType<T> {
  const [stateSnapshot, dispatchSnapshot] = React.useReducer(snapshotReducer, initialValues);
  return {
    stateSnapshot: stateSnapshot as SnapshotState<Record<string, T>>,
    dispatchSnapshot: dispatchSnapshot as React.Dispatch<SnapshotAction<Record<string, T>>>,
  };
}
```


```
 const { stateSnapshot, dispatchSnapshot } = useSnapshotReducer<RomiCore.TaskSummary>(
    initialValues,
  );
```

matiasbavera:
Rebased onto master

koonpeng (changes requested):
image

The switch to `SimpleInfo` seems to have broken the margins/paddings.

koonpeng:
I pushed a commit to rework the `snapshot-*` to `trashbin-*` to better support typescript, also I feel trash bin is a better name for what it does. Please help test it if the functionalities work as expected.

packages/react-components/lib/task-summary/task-summary-accordion.tsx
```diff
@@ -0,0 +1,179 @@
...
+      <Typography variant="body1" id="task-actor" className={classes.taskActor}>
+        {actor}
+      </Typography>
+    );
+  };
+
+  const renderTaskTreeItem = (task: RomiCore.TaskSummary) => {
+    return (
+      <TreeItem
+        data-component="TreeItem"
```
koonpeng, line 83: Avoid using `data-component`, see if there is any suitable aria roles (listitem or row maybe?) for this. Need to change the e2e tests as well after we changed this.

koonpeng (changes requested):

matiasbavera:
> image
> 
> The switch to `SimpleInfo` seems to have broken the margins/paddings.

Oh man, you're right, good catch. I'll fix that. Thanks

matiasbavera:
> > image
> > The switch to `SimpleInfo` seems to have broken the margins/paddings.
> 
> Oh man, you're right, good catch. I'll fix that. Thanks

Fixed 
image

matiasbavera:
> I pushed a commit to rework the `snapshot-*` to `trashbin-*` to better support typescript, also I feel trash bin is a better name for what it does. Please help test it if the functionalities work as expected.

Sorry, I rebase it onto master and I lost the changes =/. But I renamed all over again.

packages/react-components/lib/task-summary/task-summary-accordion.tsx
matiasbavera: Ok, I'm using a role now, `role='treeitem'`

koonpeng:
Sorry, I made other changes as well which change the implementation of the reducer. Can you try to merge these changes and see how it goes?

[trashbin.zip](https://github.com/osrf/romi-dashboard/files/5627226/trashbin.zip)

matiasbavera:
> ange the implementation of the reduc

It's working correctly

packages/react-components/lib/trashbin-control-button-group-reducer.tsx
```diff
@@ -0,0 +1,105 @@
+import React from 'react';
+
+export type TrashBinState<T> = {
+  current: T;
+  trash: T;
```
matiasbavera, line 5: I was afraid that when we empty the `trash` state triggers a component re-render, that's why I was using the ref, but it seems not to be the case. So, good solution =)

koonpeng (approved):
LGTM!

PR #162, opened 2020-11-11
allows to use env var to specify ros args

koonpeng:
## What's new

This adds a way to specify the ros args using environment variables, this is needed to run rclnodejs with SROS. It requires v0.1.3 (not released yet) of [romi-js-rclnodejs-transport](https://github.com/osrf/romi-js-rclnodejs-transport).

## Self-checks

## Discussion

[Questions for reviewers]


matiasbavera:
On hold until `romi-js-rclnodejs-transport`  v0.1.3 is released?

Now the `RclnodejsTransport.create(config.ros2NodeName);`
supports only one parameter
`static create(nodeName: string): Promise<RclnodejsTransport>;`

koonpeng:
Yep, I was testing out some stuff before releasing v0.1.3

koonpeng:
Turns out we don't really need this as SROS2 already allows us to specify the args in a env var. But I went ahead and publish v0.1.3 anyway since it's the feature is already done.

matiasbavera (approved):
LGTM

PR #169, opened 2020-11-23
Fix/show dispensers with unknown state

ChawinTan:
## What's new

Dispensers in the resource manager but without states are now shown on the omnipanel. 
Updated the dispenser display panel unit test to reflect the change in logic and dependencies on resource context

### fixes # [related issue, if exist]

Related to #169 

### [Describe your changes]

Import the dispenser resource context and use the list of dispensers obtained from its context to compare it against the dispenser state. If the dispenser in the resource list is not in the state, make an empty state and add it as a new entry to the sate object. If it exist, simply add its existing state to the state object.
## Self-checks

### [Questions for reviewers]

Right now, I simply added a an unknown state for dispensers in the list without a state. Would it be better to modify the Omnipanel UI to include a new field that displays something like `state unknown`? Personally, I don't think it is needed as having an OFFLINE state and an unknown request id queue also serves the purpose


matiasbavera:
> Related to #169

I fell into the trap twice =). The above is pointing to itself (this PR).

packages/dashboard/src/components/tests/dispensers-panel.test.tsx
```diff
@@ -14,8 +17,16 @@ beforeEach(() => {
...
+    <ResourcesContext.Provider value={resources}>
+      <DispenserPanel dispenserStates={dispenserStates} />
+    </ResourcesContext.Provider>,
+  );
   const dispenserElements = root.find(DispenserItem);
-  expect(dispenserElements.length).toBe(Object.keys(dispenserStates).length);
+  const dispenersInResource: number = resources.dispensers
+    ? Object.keys(resources.dispensers.dispensers).length
+    : 0;
+  expect(dispenserElements.length).toBe(dispenersInResource);
```
matiasbavera, line 30: A related comment (but not for now, it's ok at it is now), in the very near future we are going to change Enzyme to RTL, as well as philosophy, for which we should be verifying by the `Unknown` text and not only by quantity.

packages/dashboard/src/components/dispensers-panel.tsx
```diff
@@ -18,6 +19,27 @@ export const DispenserPanel = React.memo((props: DispenserPanelProps) => {
...
+        mode: 2,
+        request_guid_queue: [],
+        seconds_remaining: 0,
+        time: { sec: 0, nanosec: 0 },
+      };
+      updateDispenserStates[dispenser] = unknownStates;
+    } else {
+      updateDispenserStates[dispenser] = dispenserStates[dispenser];
+    }
+  });
```
matiasbavera, line 41: we can put this inside a useMemo too. So it'll change only if `dispenserList` changes.

we can change this name `updateDispenserStates` to a storage kind of name and not as an action. Maybe, `newDispenserStates` or `updatedDispenseStates` or `dispenserStatesUpdated`

matiasbavera (commented):
Thanks for this! I left a couple of comments

packages/dashboard/src/components/dispensers-panel.tsx
koonpeng: Instead of hardcoding something as an unknown state, we should have the UI display something different indicating the state is unknown.

packages/dashboard/src/components/dispensers-panel.tsx
```diff
@@ -51,15 +73,15 @@ export const DispenserPanel = React.memo((props: DispenserPanelProps) => {
...
       ...prev,
       [spotlight.value]: true,
     }));
     ref.scrollIntoView({ behavior: 'smooth' });
   }, [spotlight]);
 
-  const listItems = Object.keys(dispenserStates).map((guid, i) => {
-    const state = props.dispenserStates[guid];
+  const listItems = Object.keys(updateDispenserStates).map((guid, i) => {
+    const state = updateDispenserStates[guid];
```
koonpeng, line 84: See above comment on hardcoded state, this should be something like
```tsx
Object.values(dispensers).map((dispenser) => {
  const state = dispenserStates[dispenser.guid]; // undefined if no state
  return state ? <DispenserItem>... : /* something else */;
}
```

koonpeng (changes requested):

ChawinTan:
Rewrote the fix after the latest merge to migrate dashboard components into `react-components`

koonpeng (changes requested):

ChawinTan:
Updated storybook with the correct types and an unknown item story

packages/react-components/stories/unknown-state.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+    margin: '1rem 0',
+  },
+};
+
+export const itemsWithUnknownState: Story = (args) => (
+  <div>
+    <Typography variant="body1">
+      There are situations where the state of a component (for example, dispenser or robots) is not
+      known due to them being disconnected from RMF or other reasons, resulting in them not being
+      displayed in the omnipanel.
```
koonpeng, line 25: Better to remove description of the omni panel, these are meant to be standalone components not tied to it.

packages/react-components/stories/unknown-state.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+const styles: Record<string, React.CSSProperties> = {
+  root: {
+    margin: '0 auto',
+  },
+  spacing: {
+    margin: '1rem 0',
+  },
+};
+
+export const itemsWithUnknownState: Story = (args) => (
```
koonpeng, line 20: PascalCase
```suggestion
export const ItemsWithUnknownState: Story = (args) => (
```

packages/react-components/lib/item-unknown-state.tsx
```diff
@@ -0,0 +1,27 @@
...
+  typography: {
+    padding: '0.5rem 0',
+  },
+}));
+
+interface ItemUnknownProps {
+  unknownItem: string;
+}
+
+export default function ItemUnknownState(props: ItemUnknownProps): JSX.Element {
```
koonpeng, line 15: I'm wondering how useful would this be given that it is pretty much just a Typography. The only place this is used now is in the dispenser accordion and it still would need to make sure the AccordionSummary uses the unknown state.

i.e. Usage of this needs to be coupled with an appropriate AccordionSummary for it to make sense, and even so, other accordions would probably use their own unknown representation as they need to display different things.

packages/react-components/lib/dispensers/dispenser-accordion.tsx
```diff
@@ -48,18 +49,21 @@ const useStyles = makeStyles((theme) => ({
   statusLabelOffline: { borderColor: theme.palette.error.main },
 }));
 
+// for valid dispenser states,
+// a precondition must dispenser === dispenserState.guid
```
koonpeng, line 53: This would not appear in intellisense, see below suggestion.

packages/react-components/lib/dispensers/dispenser-accordion.tsx
```diff
@@ -48,18 +49,21 @@ const useStyles = makeStyles((theme) => ({
...
 export interface DispenserAccordionProps extends Omit<AccordionProps, 'children'> {
-  dispenserState: RomiCore.DispenserState;
+  dispenserState: RomiCore.DispenserState | null;
```
koonpeng, line 55:
```suggestion
  /**
   * Pre-condition: `dispenser === dispenserState.guid`
   */
   dispenserState: RomiCore.DispenserState | null;
```

koonpeng (changes requested):

packages/react-components/lib/item-unknown-state.tsx
ChawinTan: I wanted to use this component in the robot Accordion as well. I was thinking that for items (lift, doors, robots, dispensers etc ...) with unknown states, we can use a generic component to simply tell users that the state of the item is currently not known and there is no need for a more detailed explanation of why it is unknown. But I guess using Typography would also be just fine in the accordion itself.

packages/react-components/lib/dispensers/dispenser-accordion.tsx
ChawinTan: Didn't know that this can be done!

ChawinTan: Is there a link to all the possible syntax? I can't seem to find the correct docs on vscode

koonpeng: vscode tries to find the docs from several sources, one of which is jsdoc.

ChawinTan:
Shifted UnknowItem story into design-decisions, added docs and remove item-unknown component

packages/react-components/lib/item-unknown-state.tsx
koonpeng: It would be nice if we have sometime more fancy than a simple typography. For now since it's very simple and not used any where else I say just have "inlined" in the dispenser accordion. You can make a separate PR to standardize all the components unknown representation if you would like to.

koonpeng (approved):
:+1:

PR #176, opened 2020-12-04
refactor: standalone ros2 bridge

koonpeng:
## What's new

Other than the semantic change renaming api-server to ros2-bridge,
* Removed trajectory plugin
* update rclnodejs

A previously forgotten bug in rclnodejs has been fixed so we can now update it.

The trajectory plugin has always been a wip and is not being used, going forward we should put both ros2-bridge and trajectory server behind a reverse proxy or api gateway instead of trying to implement everything in a single process.

## Self-checks

## Discussion

[Questions for reviewers]


matiasbavera (approved):
LGTM. I tested and it's working correctly.

Issue #166, opened 2020-11-20
Robots: Robots without a state are not shown


koonpeng:
We are only displaying robot's with known state, so a robot that lost connection to rmf will then disappear from the dashboard. The robot's state also does not contain all the info that we need, for example, a robot's footprint is only available from the trajectory so we have to use a hardcoded value now.

ChawinTan:
Seems like fleet manager is handling all the states of the robots. Can I assume that all robots returned by the resource manager will be at the same level? Or there is no need to display the icon but just display the robot in the omni-panel accordian, like the dispensers?

koonpeng:
afaik, the robot resources only contains the icons, they do not list all the known robots in a map. I am not aware of any rmf interfaces that provides that atm.

ChawinTan:
I see, so it differs from dispensers in that the location of all known dispensers is hardcoded in the json resource files due to rmf limitations whereas the known robots will be handled by rmf instead.

So for the case of robots, some sort of cache to store all known robots, and then referencing them later if they are disconnected would be more appropriate.

cnboonhan:
Closing this as another development should be addressing this.

PR #179, opened 2020-12-08
Fix/show robots with unknown state

ChawinTan:
### fixes # [related issue, if exist]

#166 

Display robots on the visualizer and the omnipanel that have been previously cached but have lost connection to the backend.

Make a temporary display for fleets that do not send signal upon initialization.

### [Describe your changes]

- Implemented a display in place of the accordion when a fleet fails to send signal upon start up
No_fleet_signal
- Implemented a display to show robots that were previously cached but have disconnected in the accordion
no_signal_accordion
one_signal_one_no
- Implemented a caching system to store new fleet, robots and update their respective state for comparison
- Robot overlay uses cachedRobots to display icons and related data instead

## Self-checks

### [Questions for reviewers]

Right now, no icons are displayed if no robot have been cached. Should we hardcode the default starting positions of the robots in `resourceManager` in this scenario to display the icons?

Currently, when a fleet fails to send any signal from any of its robots, the omnipanel displays a message as shown in the first picture. Is this sufficient for now or should we implement something more robust, for example, if the fleet is still not returning states after 50 subscriptions, then we send an sms to the operator?


ChawinTan:
Will add the stories later when things are more confirmed

koonpeng (changes requested):
I would suggest to have proper backend support instead of doing client side caching. The problem with it is that each cache will have its own cache, so eventually different clients will report different robots in some scenarios, for example

1. client A starts
2. client A receives fleet state with RobotA and RobotB
3. client B starts
4. both client receive fleet state with RobotA

now client A says RobotB has unknown state, but client B says there is no such thing as RobotB. So we need a backend that can provide the "single truth" of all robots managed by rmf and their current status, we could have a server that listens to fleet states and attempts to synchronize the robot statuses based on various factors like how long since the last state received. But imo it's better if rmf can provide that information since it should have more reliable data on when robot is added/removed from the system.

koonpeng:
converting this to a draft until the missing backend support is available.

mxgrey:
In this context, does "backend" refer to the whole distributed RMF system, or do we anticipate a specific backend webserver application (maybe that runs on the main RMF server machine) specifically to aggregate data to send off to the dashboards?

PR #180, opened 2020-12-08
Upgraded CRA and RTL. Fixed issues caused by the upgrade.

### LINKED ISSUE #178: Update CRA 3.4.3 -> v4.0.1  RTL v11.0.4  -> v11.2.2
CRA 3.4.3 -> v4.0.1
RTL v11.0.4  -> v11.2.2

Tests are working correctly upgrading and downgrading these versions. There is an error when running `npm run start` on the api-server when we upgrade  to CRA 4.0.1:

Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.  TS1205

```
     5 | import { WebSocketMiddleware } from './websocket-connect';
     6 | 
  >  7 | export { CustomLogger as Logger };
       |          ^
     8 | 
     9 | export interface RpcRequest<T = unknown> {
    10 |   version: string;
 
```

So I think we should upgrade versions of RTL and CRA on a FUP PR and fix that issue too.

_Originally posted by @matiasbavera in https://github.com/osrf/romi-dashboard/pull/170#discussion_r537734574_

matiasbavera:
## What's new

fixes #178

Upgraded CRA and RTL. Fixed issues caused by the upgrade.

Upgrading the `react-script` package caused problems with the Typescript types and the way to export some interfaces. These errors occurred when running `react-script start`. I guess it could be because of these changes
```
TypeScript 4 support # 9734
ESLint 7 # 8978
    New Jest and React Testing Library rules # 8963
```
https://github.com/facebook/create-react-app/releases

## Self-checks


packages/dashboard/tsconfig.json
```diff
@@ -18,6 +18,7 @@
     "isolatedModules": true,
     "noEmit": true,
     "jsx": "react",
+    "noFallthroughCasesInSwitch": true
```
koonpeng, line 21: From some googling, this is added to workaround a bug in react-scripts with typescript >= 4.0. According to this https://github.com/facebook/create-react-app/blob/master/CHANGELOG.md it has been fixed?

matiasbavera: Yes, it's fixed but it's added automatically to tsconfig.json every time we run `start:mock`.
image

Maybe we can issue a ticket to their project to solve that.

koonpeng: I guess if we were to create a new typescript cra the option will be in tsconfig by default.

koonpeng (approved):

PR #181, opened 2020-12-08
Added style picker according to state priority on lifts

### LINKED ISSUE #171: Is intentional Having those lifts with the same color, even with different states?
Related PR https://github.com/osrf/romi-dashboard/pull/157

matiasbavera:
## What's new

fixes #171

Added style picker according to state priority on lifts.

## Self-checks

## Discussion

Should I add tests on the lift-overlay that checks the lift style "business logic"? or a test of the variant on the react-component it's enough? 

I think having both will be great. One to check the component's correct color assignation and the other one to pick the style in the correct order of priority.

### DIFF
--- a/packages/dashboard/src/components/schedule-visualizer/lift-overlay.tsx
+++ b/packages/dashboard/src/components/schedule-visualizer/lift-overlay.tsx
@@ -1,13 +1,36 @@
 import * as RomiCore from '@osrf/romi-js-core-interfaces';
 import Debug from 'debug';
 import React, { useContext } from 'react';
-import { LiftMarker as LiftMarker_, LiftMarkerProps } from 'react-components';
+import { LiftMarker as LiftMarker_, LiftMarkerProps, useLiftMarkerStyles } from 'react-components';
 import { viewBoxFromLeafletBounds } from '../../util/css-utils';
 import { LiftStateContext } from '../rmf-contexts';
 import SVGOverlay, { SVGOverlayProps } from './svg-overlay';
 
 const debug = Debug('ScheduleVisualizer:LiftsOverlay');
 const LiftMarker = React.memo(LiftMarker_);
+
+const getLiftModeVariant = (
+  currentFloor: string,
+  liftState?: RomiCore.LiftState,
+): keyof ReturnType<typeof useLiftMarkerStyles> | undefined => {
+  if (!liftState) return 'unknown';
+
+  const liftMode = liftState.current_mode;
+
+  if (liftMode === RomiCore.LiftState.MODE_FIRE) return 'fire';
+  if (liftMode === RomiCore.LiftState.MODE_EMERGENCY) return 'emergency';
+  if (liftMode === RomiCore.LiftState.MODE_OFFLINE) return 'offLine';
+  if (liftState.current_floor === currentFloor) {
+    if (liftMode === RomiCore.LiftState.MODE_HUMAN) return 'human';
+    if (liftMode === RomiCore.LiftState.MODE_AGV) return 'onCurrentFloor';
+  } else {
+    if (liftMode === RomiCore.LiftState.MODE_HUMAN) return 'moving';
+    if (liftMode === RomiCore.LiftState.MODE_AGV) return 'moving';
+  }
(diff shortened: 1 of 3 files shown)


koonpeng (commented):
>Should I add tests on the lift-overlay that checks the lift style "business logic"? or a test of the variant on the react-component it's enough?
>
>I think having both will be great. One to check the component's correct color assignation and the other one to pick the style in the correct order of priority.

Yeah, both would be good, although testing style changes is not really recommended by RTL. 

What do you think of eventually dropping jest/jsdom and move to browser based testing with karma/jasmine? That would allow us to do screenshot based snapshot testing and solves all the problem with missing web apis in jsdom. This might be more important if we want to move to a canvas based (possibly with webGL) map renderer.

matiasbavera:
> What do you think of eventually dropping jest/jsdom and move to browser based testing with karma/jasmine? That would allow us to do screenshot based snapshot testing and solves all the problem with missing web apis in jsdom. This might be more important if we want to move to a canvas based (possibly with webGL) map renderer.

`Jest` is built on top of `Jasmine` and it's significative faster than `Karma` because of `Jsdom`. But you totally sold me the idea with this point `This might be more important if we want to move to a canvas based (possibly with webGL) map renderer`. The downside I'm seeing right now is that we wouldn't be able to test the Reducer change of states, I mean we will but we will do it through the browser (as a black box).

I'm afraid to say that we can use both of them for different scenarios, because of all the noise we can cause. 

Update: 
There are also some packages that use `Jest` to test WebGL-canvas. I think we should give it a try to make the best choice (but it's pretty sure that we are going to find missing web apis in jsdom problems). 

https://www.npmjs.com/package/jest-webgl-canvas-mock

matiasbavera:
I tried to add this test too

```

  test('renders lift childs', () => {
    const lifts = getLift();
    const bounds = new L.LatLngBounds([0, 25.7], [-14, 0]);
    const root = render(
      <LMap >
        <LiftsOverlay data-testid="test" currentFloor={'L1'} bounds={bounds} lifts={lifts} />
      </LMap>,
    );
    for (let iterator in lifts) {
      const lift = officeMap.lifts[iterator];
      // expect(root.container.querySelector(`#Lift-${lift.name}`)).toBeTruthy();
      // expect(root.getByTestId('test')).toBeTruthy();
    }
  });
```

It seems to be rendering just the `LMap` and not the child components. So I can't access nor find the lifts

koonpeng (approved):
>Should I add tests on the lift-overlay that checks the lift style "business logic"? or a test of the variant on the react-component it's enough?

I will approve this now, go ahead and merge if you don't plan to add the variant tests in react-components.

koonpeng:
>The downside I'm seeing right now is that we wouldn't be able to test the Reducer change of states, I mean we will but we will do it through the browser (as a black box).

karma/jasmine tests are not black boxes. I had some confusion about karma and wdio/selenium before but afaik, the difference is that wdio tests orchestrates the browser from "outside" while karma runs the tests "inside" the browser. jasmine itself is ran inside the browser so we can still use spies to test reducers.

We definitely need some experiments to see if it's possible to migrate our tests to karma though.

>There are also some packages that use Jest to test WebGL-canvas. I think we should give it a try to make the best choice (but it's pretty sure that we are going to find missing web apis in jsdom problems).

That is the reason why I don't like jest, we must rely on these mocks which might not behave as they do on real browsers. There are also many apis not being mocked by default and we must rely on third party packages which are not maintained and might not even work anymore (last published version is 1 year ago).

>It seems to be rendering just the LMap and not the child components. So I can't access nor find the lifts

That is one of the problem with jest, leaflet uses dom apis not supported by jsdom so it's behaviour is unknown. There are also several places which we can't test with jsdom, computed svg properties for example.

matiasbavera:
> karma/jasmine tests are not black boxes. I had some confusion about karma and wdio/selenium before but afaik, the difference is that wdio tests orchestrates the browser from "outside" while karma runs the tests "inside" the browser. jasmine itself is ran inside the browser so we can still use spies to test reducers.

Yeah, you're right, we can spy the reducers.
Moving the discussion here https://github.com/osrf/rmf-web/issues/185

PR #189, opened 2020-12-18
Moved from jest to jasmine-karma

matiasbavera:
## What's new

Moved react-components tests from `jest/jsdom` to `jasmine/karma`.

Observations:
* Changed `.test` extension to `.spec` for karma to work.
* Syntax of the assertion library changed (we are using jasmine now).
* We lost the `jest.fn()`. But we can use an empty function and `spyOn`.
* Added a karma config file and a Webpack configuration file for pre-processing the project's files. There is a `typescript-karma` compiler but it's not working great when we deal with third-party libraries and CSS. Webpack do the pre-processing for us and it's more flexible so, I opted to go with Webpack. 
* Added puppeteer to install and open a headless chrome in Github's actions server, for testing with Karma.
 

## Self-checks


matiasbavera:
Hi @koonpeng, I moved and ran all the react-component tests with Jasmine and Karma. Now I am fighting to integrate the coverage test.

matiasbavera:
We have coverage now.

image

matiasbavera:
We have this warning: 

It looks like there are several instances of `@material-ui/styles` initialized in this application.
This may cause theme propagation issues, broken class names, specificity issues, and makes your application bigger without a good reason.

It's an open issue. 
https://github.com/mui-org/material-ui/issues/15745

matiasbavera:
If we want html coverage reports we should use "istanbul-instrumenter-loader": "^0.2.0",

matiasbavera:
Having this error on the CI 

```

Unable to detect ROS, please make sure a supported version of ROS is sourced
gyp: Call to 'node scripts/ros_distro.js' returned exit status 1 while in binding.gyp. while trying to load binding.gyp
gyp ERR! configure error 
gyp ERR! stack Error: `gyp` failed with exit code: 1
gyp ERR! stack     at ChildProcess.onCpExit (/opt/hostedtoolcache/node/12.20.0/x64/lib/node_modules/npm/node_modules/node-gyp/lib/configure.js:351:16)
gyp ERR! stack     at ChildProcess.emit (events.js:314:20)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:276:12)
gyp ERR! System Linux 5.4.0-1032-azure
gyp ERR! command "/opt/hostedtoolcache/node/12.20.0/x64/bin/node" 
```

matiasbavera:
@koonpeng  Ready for firsts comments. =)

packages/react-components/lib/task-summary/task-summary-accordion.tsx
```diff
@@ -52,11 +52,16 @@ export const TaskSummaryAccordion = React.memo((props: TaskSummaryAccordionProps
...
   const [selected, setSelected] = React.useState<string>('');
 
-  const handleToggle = (event: React.ChangeEvent<Record<string, unknown>>, nodeIds: string[]) => {
+  /**
+   * Here exists a type conflict, the linter told us to use Record<string, unknown> if we use `{}` as
+   * it's on the type definition, but on build time we get this error:
+   * Type '(event: React.ChangeEvent<Record<string, unknown>>, nodeIds: string) => void' is not
+   * assignable to type '((event: ChangeEvent<{}>, nodeIds: string) => void) | ((event: ChangeEvent<{}>, nodeIds: string[]) => void) | undefined'.
+   */
+  const handleToggle = (event: React.ChangeEvent<unknown>, nodeIds: string[]) => {
```
koonpeng, line 61: We can disable the linter for this line, or see if you can work around it by explicitly specifying the type.

```ts
const handleToggle: SomeType = (event, nodeIds) => {}
```

packages/react-components/package.json
```diff
@@ -13,28 +13,49 @@
...
+    "awesome-typescript-loader": "^5.2.1",
+    "concurrently": "^5.3.0",
+    "enzyme-to-json": "^3.5.0",
     "eslint": "^7.11.0",
     "eslint-plugin-react": "^7.21.4",
     "eslint-plugin-react-hooks": "^4.1.2",
     "fontsource-roboto": "^3.0.3",
+    "jasmine": "^3.6.3",
+    "jasmine-core": "^3.6.0",
     "jest": "^26.5.2",
```
koonpeng, line 47: Remove jest if we are no longer using it?

packages/react-components/tests/dispensers/dispenser-marker.spec.tsx
```diff
@@ -2,19 +2,27 @@ import { render } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import React from 'react';
 import { DispenserMarker } from '../../lib';
+import { mockOnClick } from '../test-utils';
+
+it('triggers onClick callback when button is clicked', () => {
+  const handler = mockOnClick();
+  spyOn(handler, 'onClick');
```
koonpeng, line 9: Not sure how jasmine spy works, can we just create a spy directly instead of using an object? https://stackoverflow.com/questions/9510148/using-jasmine-to-spy-on-a-function-without-an-object.

packages/react-components/tests/omni-panel.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+      <OmniPanelView viewId={0}></OmniPanelView>
+    </OmniPanel>
+  );
+}
+
+describe('OmniPanel component', () => {
+  let handler: ReturnType<typeof mockOnClick>;
+
+  beforeEach(() => {
+    // TextEncoder is not available in node
```
koonpeng, line 20: Since we are running the tests in browser, this comment shouldn't be valid anymore.

packages/react-components/tests/robots/robot-marker.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+    colorManager = new ColorManager();
+    handler = {
+      onClick: () => console.log('mock'),
+      robotColor: async () => 'black',
+    };
+
+    spyOn(handler, 'onClick');
+    spyOn(handler, 'robotColor');
+
+    colorManager.robotPrimaryColor = handler.robotColor;
```
koonpeng, line 22: Since we are running the tests in browser, this comment shouldn't be valid anymore. There also shouldn't be a need to mock `ColorManager` now.

packages/react-components/tests/test-utils.ts
```diff
@@ -0,0 +1,15 @@
...
+  };
+};
+
+export const buildMockObject = (functionNames: string[]): Record<string, () => void> => {
+  const newMockObject: Record<string, () => void> = {};
+  functionNames.forEach((element) => {
+    newMockObject[element] = () => console.log('mock');
+  });
+  return newMockObject;
+};
```
koonpeng, line 15: We probably don't need this if we can create a spy function directly.

packages/react-components/tsconfig.json
```diff
@@ -9,7 +9,7 @@
     "allowSyntheticDefaultImports": true,
     "strict": true,
     "forceConsistentCasingInFileNames": true,
-    "module": "esnext",
+    "module": "commonjs",
```
koonpeng, line 12: I think we should be using `esnext` https://github.com/mui-org/material-ui/blob/b4d4a325c4fb4411276b2d5cd2b6f4779ea92b3a/tsconfig.json#L3. We probably need to change the webpack/karma config so that it works.

koonpeng (changes requested):
Can you make another PR (or in this PR) to see if we can replace the smoke tests with screenshot snapshot diffing?

koonpeng:
```
We have this warning:

It looks like there are several instances of @material-ui/styles initialized in this application.
This may cause theme propagation issues, broken class names, specificity issues, and makes your application bigger without a good reason.
```

Since we are using a monorepo, the `@material` packages are in the root package to avoid multiple instances being loaded. Maybe we need to change the webpack/karma config so it bundles correctly? Or maybe this warning comes from the way karma load tests, resulting in each test creating its own `@material/styles` instance?

packages/react-components/package.json
```diff
@@ -13,28 +13,49 @@
     "clean": "tsc --build --clean",
     "test": "jest --watch",
     "test:coverage": "jest --ci --coverage",
+    "test:karma": "./node_modules/karma/bin/karma start karma.conf.js",
```
koonpeng, line 16: nit:
Is it possible to use the karma cli?

matiasbavera:
Alright, I'll open a new PR

koonpeng:
Looks like this got nuked when we change `master` to `main`, go ahead and open a new PR!

packages/react-components/package.json
matiasbavera: we can use it if we install karma globally which is no ideal.

packages/react-components/tests/dispensers/dispenser-marker.spec.tsx
matiasbavera: Yeah we can. I'll create it this way `const mockOnClick = jasmine.createSpy();`

packages/react-components/tests/robots/robot-marker.spec.tsx
matiasbavera: Had a lot of errors without instantiating ColorManager. 

```
ERROR in [at-loader] ./tests/robots/robot-marker.spec.tsx:23:32 
    TS2739: Type 'Spy<Func>' is missing the following properties from type 'ColorManager': conflictHighlight, robotPrimaryColor, robotColorFromCache, _robotColorFromId, _robotColorCache

ERROR in [at-loader] ./tests/robots/robot-marker.spec.tsx:32:32 
    TS2322: Type 'Spy<Func>' is not assignable to type 'ColorManager'.
```
Also, I think we need that for this test to work:

```
colorManager.robotPrimaryColor = jasmine.CreateSpy();

  it("uses ColorManager to determine robot's color", async () => {
    await render(<RobotMarker robot={makeRobot()} fleetName="test_fleet" footprint={1} />);
    expect(colorManager.robotPrimaryColor).toHaveBeenCalled();
  });
```

packages/react-components/tsconfig.json
matiasbavera: Done!

PR #192, opened 2020-12-28
fix ros2-bridge not exiting on SIGINT

koonpeng:
## What's new

A recent rclnodejs update change the SIGINT behavior to no longer forcefully exit the process, instead it just stops all contexts. The "problem" comes from the fact that because it installs a signal handler, the default nodejs handler no longer gets called so we need to install our own handler to close the server.

## Self-checks

## Discussion

[Questions for reviewers]


codebot (approved):
LGTM thanks for chasing this down.

Issue #183, opened 2020-12-09
React-component to represent states of components that are unknown


ChawinTan:
Right now, components with unknown states are represented with a simple message in a Typography. Ideally we would want a re-usable component that can be used across the accordions of all components with more bling to it :sparkles:, or we could also notify operators in other ways, eg: (sms, an ivr?)

Feel free to give suggestions!

Related comment:
https://github.com/osrf/rmf-web/pull/169#discussion_r538964840

Related PRs:
#169 
#179

koonpeng:
Avoid posting asana links here as that is private.

ChawinTan:
Updated

PR #195, opened 2021-01-04
Feature/unknown state representation

ChawinTan:
## What's new

#183 

### [Describe your changes]

- Used the `simpleInfo` component to represent the information
- removed noWrap in `simpleInfo` Typography
- included the location into information. I think it will be useful in future for things like robots that will move around.
iten_unknown_component

## Self-checks

## Discussion

[Questions for reviewers]


packages/react-components/lib/item-unknown.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface Location {
+  x: number;
+  y: number;
+  yaw: number;
+  level_name: string;
+}
+
+export interface ItemUnknownProps {
+  name: string;
+  location: Location | undefined;
```
koonpeng, line 13: There's an issue with requiring context specific props, we can't be sure all consumers will have these props, it is also a problem if a consumers has information it wants to display on unknown state that does not fit into `name` and `location`.

There is also UX problems when the state suddenly change from something unrelated to a table with name, location and error fields.

packages/react-components/lib/item-unknown.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+  const data = [
+    { name: 'Name', value: name },
+    {
+      name: 'Location',
+      value: location
+        ? `${location.level_name} (${location.x.toFixed(3)}, ${location.y.toFixed(3)})`
+        : 'Unable to locate',
+    },
+    { name: 'Error', value: errorMsg ? errorMsg : 'No error message provided' },
```
koonpeng, line 28: I would prefer `Unknown error`.

koonpeng (changes requested):

ChawinTan:
- Made the location field optional
- included a generic `fields` parameter for displaying item related information in a table
- Included headers, icons and different coloring to provide for better visual cues when changing to an error state. I suppose this would make for a better UX to make it known that an error has occurred

packages/react-components/lib/item-unknown.tsx
```diff
@@ -0,0 +1,64 @@
...
+}
+
+type FieldsKeys = 'name' | 'value';
+type DataValueTypePrimitive = number | string;
+type DataValueTypeArray = DataValueTypePrimitive[];
+type DataValueType = DataValueTypePrimitive | DataValueTypeArray;
+
+export interface ItemUnknownProps {
+  // fields object require both key values in FieldKeys
+  fields: { [key in FieldsKeys]: DataValueType }[];
```
koonpeng, line 30: I believe `name` should be a string.

packages/react-components/lib/item-unknown.tsx
```diff
@@ -0,0 +1,64 @@
...
+  location?: Location;
```
koonpeng, line 31: Can location be part of the fields as well?

koonpeng (changes requested):
This doesn't really fix the problem with item using other presentation suddenly turning into a SimpleInfo table when there is an error. I reckon that this will fit more as a variant of `SimpleInfo`.

Better yet, we can try something that wraps a child component and overlays it with an error state. e.g. Blurring the underlying component and showing a error message on top. Does it sound like a good idea to you?

Can you add a story for unknown item as well?

ChawinTan:
> This doesn't really fix the problem with item using other presentation suddenly turning into a SimpleInfo table when there is an error. I reckon that this will fit more as a variant of `SimpleInfo`.

I think I misunderstood earlier. Do you mean that there will be other items that will display information not using simpleInfo table?

> Better yet, we can try something that wraps a child component and overlays it with an error state. e.g. Blurring the underlying component and showing a error message on top. Does it sound like a good idea to you?

Like blurring the simpleInfo table when an error state is received? Sounds good to me!

koonpeng:
> I think I misunderstood earlier. Do you mean that there will be other items that will display information not using simpleInfo table?

Yeah, as an example, suppose we have a new fancy component to display robot states with nice graphs etc. Using `ItemUnkown` will cause it to switch from the fancy component to a `SimpleInfo` table whenever there is an error, this wouldn't be nice in terms of UX.

The impressive I get from `ItemUnknown` is that it should be used any other components than just dispenser, so it should support more use cases than something that uses a `SimpleInfo` and has a location field. If your intention is to just use it for dispenser then it's better to put it in the dispenser dir and avoid exporting it.

> Like blurring the simpleInfo table when an error state is received? Sounds good to me!

Yeah, to be more clear, not just a `SimpleInfo`, it should be able to work with any components.

ChawinTan:
- Allow nesting of components within `ItemUnknown` so that it can be used across all components
- Nested component will be blurred if has a state when the `showError` turns true. If not state is provided/nested component is provided, then only the error message will be shown.
```
<ItemUnknown errorMsg={'This is an error message'} showError={true}>
        <TestComponent />
</ItemUnknown>
```

packages/react-components/lib/simple-info.tsx
```diff
@@ -89,14 +89,12 @@ export const SimpleInfo = (props: SimpleInfoProps): JSX.Element => {
...
     disabled,
-    wrap,
   }: SimpleInfoData<DataValueTypePrimitive>) => (
     <>
       <Typography className={classes.displayName} variant="body1">
         {name}
       </Typography>
       <Typography
-        noWrap={!wrap}
         variant="body1"
```
matiasbavera, line 98: Just checking, isn't this affecting the task-summary-details?

packages/react-components/stories/item-unknown.stories.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import { Meta, Story } from '@storybook/react';
+import { ItemUnknown, SimpleInfo } from '../lib';
+
+export default {
+  title: 'Item Unknown',
+  component: ItemUnknown,
+} as Meta;
+
+export const LiftsPanel: Story = (args) => {
```
matiasbavera, line 10: getting an error here 
image

```
Cannot read property 'displayName' of undefined
TypeError: Cannot read property 'displayName' of undefined
    at displayName (http://localhost:6006/vendors~main.169c47d3206e22fd5981.bundle.js:60084:22)
    at parseReactElement (http://localhost:6006/vendors~main.169c47d3206e22fd5981.bundle.js:228970:21)
    at http://localhost:6006/vendors~main.169c47d3206e22fd5981.bundle.js:228986:12
    at Array.map (<anonymous>)
    at parseReactElement (http://localhost:6006/vendors~main.169c47d3206e22fd5981.bundle.js:228985:106)
    at reactElementToJsxString (http://localhost:6006/vendors~main.169c47d3206e22fd5981.bundle.js:229449:21)
    at http://localhost:6006/vendors~main.169c47d3206e22fd5981.bundle.js:60097:76
    at mapSingleChildIntoContext (http://localhost:6006/vendors~main.169c47d3206e22fd5981.bundle.js:236638:26)
```
Just checking, Is `LiftsPanel` the name you want?

matiasbavera (commented):
Thanks for this. Left a couple of comments.

packages/react-components/lib/simple-info.tsx
ChawinTan: Restored it to the original form.

packages/react-components/stories/item-unknown.stories.tsx
ChawinTan: Changed the story name.

I can't seem to replicate this error on my environment. Are there any more detailed error messages or perhaps you can update your environment just to rule out that the error is not from your side?

packages/react-components/lib/dispensers/dispenser-accordion.tsx
```diff
@@ -91,20 +102,18 @@ export const DispenserAccordion = React.forwardRef(
...
-              The state of <b>{dispenser}</b> dispenser is unknown. Please check if{' '}
-              <b>{dispenser} </b>
-              is working properly.
-            </Typography>
-          </React.Fragment>
-        )}
+        <ItemAccordionDetails>
+          <ItemUnknown
+            errorMsg={'Dispenser is not sending states. Please check if it is working properly.'}
+            showError={!dispenserState}
```
koonpeng, line 108: Should we simplify the props by removing `showError` and use `errorMsg` to test if we should enable the blur?

packages/react-components/lib/item-unknown.tsx
```diff
@@ -0,0 +1,49 @@
...
+  },
+}));
+
+export interface ItemUnknownProps {
+  errorMsg?: string;
+  showError: boolean;
+  children: JSX.Element | null;
+}
+
+export const ItemUnknown = (props: ItemUnknownProps): JSX.Element => {
```
koonpeng, line 25: Recommend renaming this to something more meaning. `UnknownOverlay`, `ErrorOverlay` or something else that conveys what the component do.

packages/react-components/lib/item-unknown.tsx
```diff
@@ -0,0 +1,49 @@
...
+        <Grid item>
+          <Typography color="error" variant="h6" align="center">
+            Error
+          </Typography>
+        </Grid>
+        <Grid item>
+          <ErrorIcon className={classes.errorIcon} />
+        </Grid>
+      </Grid>
+      <Typography className={classes.errorMsg} color="error" variant="body1" align="center">
```
koonpeng, line 41: I don't think it's needed now but we could look at adding a prop/props to let the consumer to style the error message.

koonpeng (changes requested):
Can you show the error message over the underlying component instead of on top of it?

The underlying component is still selectable, we should disable that with `user-select`.

ChawinTan:
Here's the current state of `ErrorOverlay`, with an example of an error message style override.

error_overlay

koonpeng:
Some padding would probably make it look nicer.

We should also see if we can make the error message more readable, maybe backdrops, opacity or backgrounds. If we can't do that while still making it look nice then let's go back to error msg on top of the component.

matiasbavera:
> Some padding would probably make it look nicer.
> 
> We should also see if we can make the error message more readable, maybe backdrops, opacity or backgrounds. If we can't do that while still making it look nice then let's go back to error msg on top of the component.

+1 to this

ChawinTan:
err_msg_opacity

- increase font size
- add padding
- add opacity to background and backdrop

matiasbavera:
> err_msg_opacity
> 
> * increase font size
> * add padding
> * add opacity to background and backdrop

Since "Error" is a title, what do you think about making the title slightly bigger than the description text? other than that look good to me.

packages/react-components/lib/error-overlay.tsx
```diff
@@ -0,0 +1,76 @@
...
+  },
+}));
+
+export interface ErrorOverlayProps {
+  errorMsg?: string;
+  children: JSX.Element | null;
+  overrideErrorStyle?: string;
+}
+
+export const ErrorOverlay = (props: ErrorOverlayProps): JSX.Element => {
```
matiasbavera, line 39: Can you check if this component is re-rendering non-stop on the dashboard, with the chrome performance tool or the React profiler? 

If it is we can try adding a `React.memo`

packages/react-components/lib/dispensers/dispenser-accordion.tsx
```diff
@@ -91,20 +114,21 @@ export const DispenserAccordion = React.forwardRef(
...
+              !dispenserState
+                ? 'Dispenser is not sending states. Please check if it is working properly.'
+                : undefined
+            }
+          >
+            {dispenserState ? (
+              <DispenserInfo dispenser={dispenserState} />
+            ) : previousState ? (
+              <DispenserInfo dispenser={previousState} overrideStyle={overrideClasses} />
+            ) : null}
```
matiasbavera, line 129: just a nit, and it's up to you, an alternative to avoid nested ternary operators can be this

```
{dispenserState && <DispenserInfo dispenser={dispenserState} />}
{(!dispenserState && previousState) && <DispenserInfo dispenser={previousState} overrideStyle={overrideClasses}}
```

If you think this logic is convoluted you can leave it as it is.

packages/react-components/lib/dispensers/dispenser-accordion.tsx
```diff
@@ -91,20 +114,21 @@ export const DispenserAccordion = React.forwardRef(
...
-              is working properly.
-            </Typography>
-          </React.Fragment>
-        )}
+        <ItemAccordionDetails>
+          <ErrorOverlay
+            errorMsg={
+              !dispenserState
+                ? 'Dispenser is not sending states. Please check if it is working properly.'
+                : undefined
```
matiasbavera, line 122: A question to follow the logic. If a dispenser has a state the error message it's undefined first, and in a moment the backend stops sending the state to the UI, so now the errorMsg is 'Dispenser is not sending states. Please check if it is working properly.' If the backend starts sending the state again you will assign the errorMsg as undefined again, right?

nit: If that the case I think we should be using null. If not,  you can disregard this comment

packages/react-components/lib/dispensers/dispenser-accordion.tsx
```diff
@@ -65,6 +76,18 @@ export const DispenserAccordion = React.forwardRef(
...
+    function usePrevDispenserState(dispenserState: RomiCore.DispenserState | null) {
+      const ref = useRef<RomiCore.DispenserState | null>(null);
+      useEffect(() => {
+        if (dispenserState) {
+          ref.current = dispenserState;
+        }
+      });
+      return ref.current;
+    }
+    const previousState = usePrevDispenserState(dispenserState);
```
matiasbavera, line 90: Not for now and just a comment, if we are going to use the error state and the same logic for other dashboard components as well, maybe creating a custom hook to handle this should be a good idea.

matiasbavera (commented):
Thanks for the changes. Left a couple of comments, HTH.

koonpeng:
It looks much better now! There is a small problem though, it looks like the backdrop is not correctly covering the underlying component, you can see the line where the backdrop starts.

Another small nit, I think it is convention to put the icon before the message https://www.google.com/search?q=error+dialog&tbm=isch&ved=2ahUKEwjL7JbVo6TuAhVFS3wKHXyHCiYQ2-cCegQIABAA&oq=error+dialog&gs_lcp=CgNpbWcQAzIECCMQJzIECCMQJzICCAAyAggAMgIIADICCAAyAggAMgIIADIGCAAQBRAeMgYIABAFEB46BAgAEB5QmY0CWMiOAmDvkAJoAHAAeACAASWIAUOSAQEymAEAoAEBqgELZ3dzLXdpei1pbWfAAQE&sclient=img&ei=W90EYIvTH8WW8QP8jqqwAg&bih=1327&biw=2560

packages/react-components/lib/dispensers/dispenser-accordion.tsx
ChawinTan: Good Idea!

Depending on how the states are going to be handled in future, we may or may not need this, so I will make a comment and keep it as it is for now

ChawinTan:
Some ui changes and made some optimizations that matias recommended

matiasbavera:
image

Do you think is it possible to make the icon bigger?

matiasbavera (commented):
LGTM, left a comment about the icon size. After that, It should be ready to go.

matiasbavera:
You may need to rebase this.

koonpeng:
the unknown state story is not display correctly, maybe because there is no dispenser state?
image

packages/react-components/lib/dispensers/dispenser-accordion.tsx
```diff
@@ -1,20 +1,21 @@
-import { Accordion, AccordionProps, makeStyles, Divider } from '@material-ui/core';
-import Typography from '@material-ui/core/Typography';
+import { Accordion, AccordionProps, makeStyles } from '@material-ui/core';
 import * as RomiCore from '@osrf/romi-js-core-interfaces';
 import Debug from 'debug';
-import React from 'react';
+import React, { useEffect, useRef } from 'react';
```
koonpeng, line 4: stick to a convention whether to use default import for "destructured" import, in general we are using default import for react. I don't really mind if you prefer "destructured", if you would like to use it, change the other references to remove `React.` as well.

packages/react-components/lib/error-overlay.tsx
```diff
@@ -0,0 +1,78 @@
...
+  overrideErrorStyle?: string;
+}
+
+export const ErrorOverlay = React.memo(
+  (props: ErrorOverlayProps): JSX.Element => {
+    const classes = useStyles();
+    const { errorMsg, children, overrideErrorStyle } = props;
+
+    return errorMsg ? (
+      <React.Fragment>
```
koonpeng, line 45: We don't need `React.Fragment` here as the `div` already encapsulates everything

packages/react-components/lib/error-overlay.tsx
```diff
@@ -0,0 +1,78 @@
...
+                </Typography>
+              </Grid>
+            </Grid>
+            <Typography
+              className={overrideErrorStyle ? overrideErrorStyle : classes.errorMsg}
+              color="error"
+              variant="h6"
+              align="center"
+            >
+              {errorMsg ? errorMsg : 'Unknown error'}
```
koonpeng, line 69: `errorMsg` will never be falsy here because of the check above.

packages/react-components/lib/error-overlay.tsx
```diff
@@ -0,0 +1,78 @@
...
+    display: 'grid',
+  },
+  disableSelect: {
+    userSelect: 'none',
+  },
+}));
+
+export interface ErrorOverlayProps {
+  errorMsg?: string | null;
+  children: JSX.Element | null;
```
koonpeng, line 35: use `ReactNode` to correctly handle any children types.

koonpeng (changes requested):
image
do you think it would look better if we center the error vertically?

ChawinTan:
- Update ItemUnknown in design decision
- align error message vertically
- address comments

koonpeng (approved):

PR #197, opened 2021-01-05
build: rework package management

koonpeng:
## What's new

 The previous implementation chains npm install/ci with the "preinstall" script, this is bad for
 several reasons.
   1. npm exports some env variables, chaining npm in scripts causes the commands to inherit
     the context of the parent npm, this may cause corrupted installs or undefined behaviors.
   2. The "preinstall" script is always ran in a npm install/ci, even for transitive dependencies,
     this causes all packages to be not "publishable" as those "preinstall" scripts shouldn't be
     there on an "external" install.

The new solution simply use a bootstrap script to prepare all the packages, docs are also updated to use this new workflow.

We may want to move to "meta" package managers in the future (lerna, rush). The downside of these packages managers is that they take over package installations and management, i.e. you must not run certain npm commands or it may "corrupt" your `node_modules`.

We may also consider moving to yarn workspaces or npm workspaces (require npm >= 7).

## Self-checks

## Discussion

[Questions for reviewers]


codebot (changes requested):
request keeping Ubuntu 20.04 specific installation instructions (labeling them as such in a section header) as well as providing the links to the portable (much longer) installation options as done here. Otherwise LGTM, thank you for factoring out bootstrap from preinstall

PR #199, opened 2021-01-07
fix redirection of unknown routes to NotFoundPage

codebot:

## What's new

The router previously would accept any route and deliver the `Dashboard` component, because by default the `path` was empty. This PR says that the Dashboard route should only be provided by `/`. This means that after this PR, if you go to `localhost:3000/foo` it now properly displays the `NotFoundPage` component.

## Self-checks

## Discussion

[Questions for reviewers]


koonpeng (commented):
Hmm, should we instead fix BASE_PATH by making sure it always starts with a `/`? See `url.ts`, I feel this should make more sense, so the route for dashboard would be `BASE_PATH`, and for others would be `${BASE_PATH}/my_path`.

If we do that, we should update the docs as well, there is a line that says it defaults to `''`

koonpeng:
I made a commit to enforce BASE_PATH to start with '/', can you try if it fixes things?

koonpeng:
interesting, chrome doesn't complain on `//login`.

I think we need to make `BASE_PATH` always start with `/` and never end with `/`, unless it is `/`.

koonpeng:
hmm, yea you are right. But the other way has problem also, if `BASE_PATH` is some other path `some/path`, then `${BASE_PATH}/other` would be `some/path/other`, no leading slash. I'm not sure if this will screw up routing also.

PR #202, opened 2021-01-12
build: always run npm regardless if node_modules is present

koonpeng:
## What's new

this is useful when switching branches or pulling new changes to ensure everything is up to date.

The old behavior is present because it used to be that we recursively install dependent packages, this is no longer the case, we now require the user to manually run bootstrap.

## Self-checks

## Discussion

[Questions for reviewers]

### DIFF
--- a/scripts/bootstrap.js
+++ b/scripts/bootstrap.js
@@ -36,16 +36,8 @@ scope.forEach((pkg) => {
 });
 targets.forEach((pkg) => {
   const cwd = `${__dirname}/../${pkg}`;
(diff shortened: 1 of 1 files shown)


codebot (approved):
Seems good to me; the only difference so far as I understand from reading this is that the `bootstrap` script may take a bit longer to run, but it's guaranteed now to get all the dependencies correct?

koonpeng:
yea, first run should be the same, later runs will take longer but I think its worth it to ensure things are correct.

PR #205, opened 2021-01-14
react-components: system summary

ChawinTan:
## What's new

- Notifications Dialog panel
- An item status panel on the omni panel for better visual cues

## fixes # [related issue, if exist]

- remove the existing snackbar implementation #206 

## [Describe your changes]

- Added a status accordion panel for better visual cues when a system has gone down
- Added a new notifications dialog panel for more detailed error messages
- Remove the existing snackbar notifications

Here is the view of the main menu: 
When a part of the system goes down, the indicator will turn red
main_menu

View when all systems are up:
status_accordion_dashboard

View in storybook, when some systems are down:
Status_accordion

Notification Dialogue when all systems are up:
notification-dialogue_empty

Notification Dialogue when there are messages:
notifications_dialogue

The idea is that when something happens and the indicator turns red, the user will be alerted about an error. A notification will also be sent at the same time and the user can then click on the alert icon to see more detailed messages.

Opening the omni panel also lets the user know which item is currently facing problems.

## Self-checks

## Discussion

[Questions for reviewers]


ChawinTan:
@koonpeng @matiasbavera This pr is ready for the first round of review. Once the design is more or less confirmed, I can then add the unit test in `react-components`.

packages/dashboard/src/components/dashboard/dashboard.tsx
```diff
@@ -274,6 +278,36 @@ export default function Dashboard(_props: {}): React.ReactElement {
     [dashboardDispatch, appController],
   );
 
+  // temp code with mock data to display status
```
ChawinTan, line 281: These are code that I write to show how it looks like in the dashboard for demonstration purposes. Should we keep these just for display purposes or should we remove it from the dashboard once it is ready to be commited?

matiasbavera:
Just an idea, If the number of elements we should keep track of grows, maybe it will be a good idea to have an accordion opened by default? So users can close "irrelevant" information (things that are working right) 

image

ChawinTan:
Redesign the omni-panel to include a summary of system and shifted the notifications button to the omni panel as well. Right now, I am categorizing several modes together as either `operational` or `out of order`, with 2 more modes, `idle` and `charging` for robots.

Also made some changes to the notifications dialog base on mattias's suggestions above.

new_omni_panel_1
rework_omni_panel_2
notifications_dialog_new

matiasbavera:
> Redesign the omni-panel to include a summary of system and shifted the notifications button to the omni panel as well. Right now, I am categorizing several modes together as either `operational` or `out of order`, with 2 more modes, `idle` and `charging` for robots.
> 
> Also made some changes to the notifications dialog base on @matiasbavera suggestions above.

Oh thanks for these changes, I feel that interface tell me a lot more than the other one.

new_omni_panel_1

I'll mention one point to open a discussion here (please tell me if you don't understand me):

I think that from a UX point of view, when a user opens this panel will have to look at every single red squared searching for an error (if there is one, if there isn't one the user will lose time looking for one). So that will be "extra-work" that the end-user would do every time he wants to check the general status. I think that saying extra-work is exaggerated but, colors attract more brain attention than letters and I think we can take advantage of that.

So maybe a good idea could be for the unactive states no to appear or they can stay with a gray color where they can pass unnoticed. With this approach, the end-user will have to search for a different color from gray and green to find a problem or something different quickly.

Let me know what you think. 

Let's also ask for some feedback. 

Just in case you have time to give some feedback on the UI, CC: @koonpeng  @codebot @vallq

ChawinTan:
> I think that from a UX point of view, when a user opens this panel will have to look at every single red squared searching for an error (if there is one, if there isn't one the user will lose time looking for one). So that will be "extra-work" that the end-user would do every time he wants to check the general status. I think that saying extra-work is exaggerated but, colors attract more brain attention than letters and I think we can take advantage of that.
> 
> So maybe a good idea could be for the unactive states no to appear or they can stay with a gray color where they can pass unnoticed. With this approach, the end-user will have to search for a different color from gray and green to find a problem or something different quickly.

Not sure if I understood correctly, but I think what you are trying to say is that we can be more efficient if we don't show boxes if all equipment in a category is all green? In that case, I think we can simply display a message like `All equipment green` and then a message like `2 robots out of order` when robots are out of order with a red font color

packages/dashboard/src/components/dashboard/main-menu.tsx
```diff
@@ -1,29 +1,64 @@
-import { Divider, List, ListItem, makeStyles, Typography } from '@material-ui/core';
+import { Divider, makeStyles, Typography, Button } from '@material-ui/core';
+import * as RomiCore from '@osrf/romi-js-core-interfaces';
 import Debug from 'debug';
 import React from 'react';
 import DashboardTooltip from 'react-components/lib/tooltip';
 import { TooltipsContext } from '../app-contexts';
 import { OmniPanelViewIndex } from './dashboard';
+import { MainMenuItemState } from './main-menu-item-state';
```
koonpeng, line 8: `main-menu-item-state` depends on `main-menu` as well. Try to avoid recursive dependencies whenever possible, I didn't check but this may also apply to `main-menu-task-state` etc.

koonpeng (changes requested):
Can you move the new components (system summary particularly) to `react-components` so that it is easier to test and iterate on the design.

koonpeng:
> > I think that from a UX point of view, when a user opens this panel will have to look at every single red squared searching for an error (if there is one, if there isn't one the user will lose time looking for one). So that will be "extra-work" that the end-user would do every time he wants to check the general status. I think that saying extra-work is exaggerated but, colors attract more brain attention than letters and I think we can take advantage of that.
> > So maybe a good idea could be for the unactive states no to appear or they can stay with a gray color where they can pass unnoticed. With this approach, the end-user will have to search for a different color from gray and green to find a problem or something different quickly.
> 
> Not sure if I understood correctly, but I think what you are trying to say is that we can be more efficient if we don't show boxes if all equipment in a category is all green? In that case, I think we can display a message like `All equipment green` and then a message like `2 robots out of order` when robots are out of order with a red font color.
> 
> Another way could be to use only 2 colors: green/grey for equipment that are ok and red for something that is out of order.
> 
> I initially wanted this to be more of a summary of the systems and data displayed are not only meant for errors/alerts

Some opinions:

1. There isn't much / any separation between different components.
2. The headers (Doors/Lifts etc) looks too small to me, also the "details" button is quite small compared to the status boxes.
3. Capitalization and underscores.
4. User can't easily find what is out of order, you only see a number "X out of order", I assume you have to click `notifications` and filter through the messages and try to find which message is the cause of problem.

I don't really mind the current approach, but a more efficient design would be better I guess. We could try to combine multiple boxes? Something like

```
---------------
|     3/3     |
| Operational |
---------------
```

It would be green if everything is working, and yellow or red depending on the highest severity. But it doesn't work for components with more than 2 states though.

We could also make use of the banner from the previous design

It would be green only if all components are OK, and yellow or red depending on the highest severity, note that idling robots should not be considered a problem.

matiasbavera:
> I don't really mind the current approach, but a more efficient design would be better I guess. We could try to combine multiple boxes? Something like
> 
> ```
> ---------------
> |     3/3     |
> | Operational |
> ---------------
> ```
> 
> It would be green if everything is working, and yellow or red depending on the highest severity. But it doesn't work for components with more than 2 states though.
> 
> We could also make use of the banner from the previous design
> It would be green only if all components are OK, and yellow or red depending on the highest severity, note that idling robots should not be considered a problem.

Hi @ChawinTan , this is what I was trying to express, KP put it in simple words, and it is better understood. =)

ChawinTan:
- Added back the banner
- Increased the font size of the headers and details button
- Change the box layout
- Added a new panel to show the equipment that are out of order
Top_omnipanel
midway_omnipanel

- Modified the robot state calculation. `charging` and `idle` states were in their own category previously. Now I included both states into the `operational` category. 2 functional robots that are charging will be shown in the `charging` and `operational` box at the same time.
robot_pics

matiasbavera:
> * Added back the banner
> * Increased the font size of the headers and details button
> * Change the box layout
> * Added a new panel to show the equipment that are out of order

¡Estos cambios se ven geniales! Gracias por esto.

packages/dashboard/src/components/dashboard/main-menu.tsx
```diff
@@ -52,55 +88,220 @@ export const MainMenu = React.memo((props: MainMenuProps) => {
...
+    const getSpoiltDoor = getDoorSummary().summary.outOfOrder;
+    const getSpoiltLift = getLiftSummary().summary.outOfOrder;
+    const getSpoiltDispensers = getDispenserSummary().summary.outOfOrder;
+    const getSpoiltRobots = getRobotSummary().summary.outOfOrder;
+
+    return getSpoiltDoor + getSpoiltLift + getSpoiltDispensers + getSpoiltRobots !== 0;
+  };
+
+  const getSpoiltEquipment = (): string[] => {
+    const getSpoiltDoor = getDoorSummary().outOfOrder;
```
koonpeng, line 226: nit: This and `getBannerColor` repeats the same iteration through `getDoorSummary()` etc.

packages/dashboard/src/components/dashboard/main-menu.tsx
```diff
@@ -52,55 +88,220 @@ export const MainMenu = React.memo((props: MainMenuProps) => {
...
-      </ListItem>
-      <Divider />
-
-      <ListItem data-item="Commands" button={true} onClick={handleMainMenuCommandsClick}>
-        <DashboardTooltip
-          title="This panel shows the commands that a user can request and RoMi will allocate the most suitable robot for the task"
-          id="commands-tooltip"
-          enabled={showTooltips}
+    <React.Fragment>
+      <MainMenuBanner bannerUrl={'http://localhost:3000/favicon.ico'} isError={getBannerColor()} />
```
koonpeng, line 235: Do not hardcode the url, use `/favicon.ico` if possible.

packages/dashboard/src/components/notification-bar.tsx
```diff
@@ -1,34 +0,0 @@
-import { Snackbar } from '@material-ui/core';
```
koonpeng, line 1: This may still be useful, let's keep the code but leave it unused. Let's also rename it to `AlertSnackBar` or something similar to avoid confusion with the new notification dialog.

packages/dashboard/src/stories/baseComponents/notificationBar-component.tsx
```diff
@@ -1,55 +0,0 @@
-import React, { useState } from 'react';
```
koonpeng, line 1: This may still be useful, let's keep the code but leave it unused.

packages/dashboard/src/stories/notification-bar.stories.tsx
```diff
@@ -1,15 +0,0 @@
-import React from 'react';
```
koonpeng, line 1: This may still be useful, let's keep the code but leave it unused.

packages/dashboard/src/serviceWorker.ts
```diff
@@ -15,9 +15,7 @@ const isLocalhost = Boolean(
     // [::1] is the IPv6 localhost address.
     window.location.hostname === '[::1]' ||
     // 127.0.0.1/8 is considered localhost for IPv4.
-    window.location.hostname.match(
```
koonpeng, line 18: Was anything else changed aside from formatting?

packages/react-components/lib/systems-summary/system-summary-type.tsx
```diff
@@ -0,0 +1,41 @@
...
+  charging?: number;
+}
+
+export interface ItemSummary {
+  item: string;
+  summary: ItemSummaryState;
+  outOfOrder: string[];
+}
+
+export interface MainMenuItemStateProps {
```
koonpeng, line 17: I would prefer to put these interfaces in the same file as the components, @matiasbavera do you have any preference?

packages/react-components/lib/systems-summary/main-menu-item-state.tsx
```diff
@@ -0,0 +1,126 @@
...
+  robotGrid: {
+    marginTop: theme.spacing(1),
+  },
+}));
+
+export const MainMenuItemState = (props: MainMenuItemStateProps): JSX.Element => {
+  const classes = useStyles();
+  const { itemSummary, handleClick } = props;
+
+  const totalItem: number = React.useMemo(() => {
```
koonpeng, line 42: nit: I think this is simple enough that we don't need `useMemo`

packages/react-components/lib/systems-summary/main-menu-item-state.tsx
```diff
@@ -0,0 +1,126 @@
...
+    return itemSummary.summary.operational + itemSummary.summary.outOfOrder;
+  }, [itemSummary]);
+
+  const operationalItem: number = React.useMemo(() => {
+    return itemSummary.summary.operational;
```
koonpeng, line 47: This has no advantage in being memoed.

packages/react-components/lib/systems-summary/system-summary-type.tsx
```diff
@@ -0,0 +1,41 @@
...
+  [key: string]: number;
+}
+
+export interface MainMenuBannerProps {
+  bannerUrl: string;
+  isError: boolean;
+}
+
+export interface MainMenuSpoiltItemsProps {
+  spoiltItems: string[];
```
koonpeng, line 40: Allow an optional message to show the underlying cause of problem?

packages/react-components/lib/systems-summary/system-summary-type.tsx
```diff
@@ -0,0 +1,41 @@
...
+export interface MainMenuAlertProps {
+  notifications: Notification[];
+}
+
+export interface MainMenuTaskStateProps {
+  tasks: RomiCore.TaskSummary[];
+}
+
+export interface TaskSummaryState {
+  [key: string]: number;
```
koonpeng, line 31: use explicit keys since we know all the states that a task can be in.

packages/react-components/lib/notifications-dialog.tsx
```diff
@@ -0,0 +1,247 @@
...
+export const NotificationsDialog = (props: NotificationDialogProps): JSX.Element => {
+  const classes = useStyles();
+
+  const { showNotificationsDialog, setShowNotifications, notifications } = props;
+
+  const [level, setLevel] = React.useState('');
+  const [rmfNotifications, setRmfNotifications] = React.useState(notifications);
+  // a copy of rmfNotifications
+  // ensure notifications can be restored to correct state after filtering
+  const [notficationsCopy, setNotificationsCopy] = React.useState(notifications);
```
koonpeng, line 134: Instead of maintaining 3 sets of notification, you can `useMemo` to filter the notifications that should be shown.

packages/react-components/lib/notifications-dialog.tsx
```diff
@@ -0,0 +1,247 @@
...
+        if (notification.severity === val) {
+          filterNotifications.push(notification);
+        }
+      });
+      setRmfNotifications(filterNotifications);
+    }
+  };
+
+  // delete notifications once marked read
+  const deleteReadNotifications = (i: number) => {
```
koonpeng, line 165: I suggest making this a prop, I think this implementation wouldn't work when the parent component reopens the dialog, the deleted notifications will come back.

packages/react-components/lib/notifications-dialog.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+export interface Notification {
+  time: string;
+  error: string;
+  severity: string;
+}
+
+export interface NotificationDialogProps {
+  showNotificationsDialog: boolean;
+  setShowNotifications: (payload: boolean) => void;
```
koonpeng, line 27: Should this be `onClose` or `onOpen` instead?

packages/react-components/lib/systems-summary/system-summary-type.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+export interface ItemSummary {
+  item: string;
+  summary: ItemSummaryState;
+  outOfOrder: string[];
+}
+
+export interface MainMenuItemStateProps {
+  itemSummary: ItemSummary;
+  handleClick: () => void;
```
koonpeng, line 19: I believe the standard is to be called `onClick`

packages/react-components/stories/systems-summary.stories.tsx
```diff
@@ -0,0 +1,137 @@
...
+
+const spoiltEquipment: MainMenuSpoiltItemsProps = {
+  spoiltItems: ['lift1 - fire', 'door1 - unknown', 'robot1 - unknown'],
+};
+
+export const MainMenuItemStateStory: Story = (args) => (
+  <React.Fragment>
+    <MainMenuItemState
+      itemSummary={itemStateDataDoor.itemSummary}
+      handleClick={itemStateDataDoor.handleClick}
```
koonpeng, line 97: If this is called `onClick` storybook should pick this up and expose it as an action.

koonpeng (changes requested):

packages/dashboard/src/serviceWorker.ts
ChawinTan: Nope, I copied and paste the original version in the main branch

packages/react-components/lib/notifications-dialog.tsx
ChawinTan: Will there be a back end sending deleted notifications or will a reducer implementation suffice for now?

koonpeng: Yes, we will need to keep track of all notifications status in a database. This should not affect this component though, the job of updating the state should be left to the parent component, it will do whatever is needed and send a new `notifications` props without the deleted notifications.

ChawinTan:
Address comments and added unit tests

commented out the `deleteNotifications` function for now and left a TODO comment

packages/react-components/tests/systems-summary/main-menu-alert.test.tsx
```diff
@@ -0,0 +1,16 @@
+import React from 'react';
+import { MainMenuAlert } from '../../lib';
+import { render } from '@testing-library/react';
+import { notifications, deletedNotifications } from './test.utils';
+
+test('button should be disabled when there are no notifications', () => {
```
koonpeng, line 6: Need some tests for the onClick behavior.

packages/react-components/tests/systems-summary/main-menu-banner.test.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import { MainMenuBanner } from '../../lib';
+import { render } from '@testing-library/react';
+
+test('smoke test', () => {
```
koonpeng, line 5: needs some tests for the `isError` prop.

packages/react-components/tests/systems-summary/main-menu-banner.test.tsx
```diff
@@ -0,0 +1,7 @@
...
+  render(<MainMenuBanner bannerUrl={''} isError={false} />);
```
koonpeng, line 6: Maybe make `bannerUrl` optional? Also suggest renaming it to `imageSrc`, `imageUrl`, `image` or similar.

packages/react-components/tests/systems-summary/main-menu-item-state.test.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { MainMenuItemState } from '../../lib';
+import { itemSummary } from './test.utils';
+import { render } from '@testing-library/react';
+
+test('smoke test', () => {
+  render(<MainMenuItemState itemSummary={itemSummary} onClick={jest.fn()} />);
+});
+
+test('idle and charging panels to be rendered when item value is Robots', () => {
```
koonpeng, line 10: This seems like a red flag to me, it shouldn't behave differently depending on the "item value".

packages/react-components/tests/systems-summary/main-menu-spoilt-items.test.tsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+import { MainMenuSpoiltItems } from '../../lib';
+import { render } from '@testing-library/react';
+
+test('smoke test', () => {
```
koonpeng, line 5: Needs tests on the `summary` and `errorMessage` props.

packages/react-components/lib/systems-summary/main-menu-task-state.tsx
```diff
@@ -1,6 +1,17 @@
 import React from 'react';
 import { makeStyles, Typography, Grid, Paper } from '@material-ui/core';
-import { MainMenuTaskStateProps, TaskSummaryState } from './index';
+import * as RomiCore from '@osrf/romi-js-core-interfaces';
+
+export interface MainMenuTaskStateProps {
```
koonpeng, line 5: It would be helpful to make this clickable.

packages/react-components/tests/systems-summary/main-menu-task-state.test.tsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import { MainMenuTaskState } from '../../lib';
+import { render } from '@testing-library/react';
+import { tasks } from './test.utils';
+
+test('smoke test', () => {
```
koonpeng, line 6: Needs tests on the `tasks` props, check that the correctly number of tasks is render for each category.

koonpeng (changes requested):
It's abit weird that there are a bunch of `MainMenu*` components in `react-components` but the actual `MainMenu` is not there. See if you can move `MainMenu` into `react-components`, or rename all the `MainMenu*` components to be more generic.

packages/dashboard/src/components/dashboard/main-menu.tsx
```diff
@@ -52,55 +90,235 @@ export const MainMenu = React.memo((props: MainMenuProps) => {
...
+    const getRobots = getRobotSummary();
+    return {
+      door: getDoor,
+      lift: getLift,
+      dispenser: getDispensers,
+      robot: getRobots,
+    };
+  };
+
+  const toggleBannerColor = (): boolean => {
```
koonpeng, line 231: Is seems misleading, it's called `toggleBannerColor` but doesn't actually toggle anything and returns a boolean.

packages/dashboard/src/components/dashboard/main-menu.tsx
```diff
@@ -178,23 +180,23 @@ export const MainMenu = React.memo((props: MainMenuProps) => {
     fleetKeys.forEach((fleet) => {
       fleets[fleet].robots.forEach((robot) => {
         switch (robot.mode.mode) {
-          case 8:
-          case 5:
+          case RomiCore.RobotMode.MODE_ADAPTER_ERROR:
+          case RomiCore.RobotMode.MODE_EMERGENCY:
```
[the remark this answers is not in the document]
ChawinTan: I assume there would be a separate health state manager that is independent of the component state managers that we have now?

packages/react-components/lib/index.ts
```diff
@@ -24,5 +24,6 @@ export * from './trashbin-control-button-group-reducer';
 export * from './utils';
 export * from './waypoints';
 export * from './error-overlay';
+export * from './material-table-icons';
```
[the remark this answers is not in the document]
ChawinTan: This is a separate line of code from another pr that was conflicting with my code. Will remove it.

packages/react-components/lib/systems-summary/main-menu-alert.tsx
```diff
@@ -1,8 +1,12 @@
 import React from 'react';
 import { makeStyles, Typography, Button, Badge } from '@material-ui/core';
 import NotificationsIcon from '@material-ui/icons/Notifications';
-import { NotificationsDialog } from '../index';
-import { MainMenuAlertProps } from './index';
+import { NotificationsDialog, Notification } from '../index';
+
+export interface MainMenuAlertProps {
+  notifications: Notification[];
+  deletedNotifications: Notification[];
```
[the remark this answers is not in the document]
ChawinTan: I think I got it the other way around

packages/react-components/lib/systems-summary/main-menu-spoilt-items.tsx
```diff
@@ -20,9 +28,9 @@ export const MainMenuSpoiltItems = (props: MainMenuSpoiltItemsProps): JSX.Elemen
       </Typography>
       {spoiltItems.map((item) => {
         return (
-          <Paper className={classes.paper} key={item} elevation={3}>
+          <Paper className={classes.paper} key={item.summary} elevation={3}>
```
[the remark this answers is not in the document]
ChawinTan: It's a concatenation of the name and string. Will give it a better naming

packages/react-components/lib/systems-summary/main-menu-item-state.tsx
```diff
@@ -1,7 +1,24 @@
...
+  operational: number;
+  outOfOrder: number;
+  idle?: number;
+  charging?: number;
+}
+
+export interface ItemSummary {
+  item: string;
+  summary: ItemSummaryState;
+  outOfOrder: string[];
```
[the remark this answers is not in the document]
ChawinTan: It is used in `main-menu` to collect all the spoilt equipment. Will rename it to something less confusing

ChawinTan:
Addressed comments

- refactored to allow clicking of spoilt items and added tests
- refactored to allow clicking of task summary panel and added tests
- refactored to subscribe health status from a context

packages/dashboard/src/managers/rmf-health-state-manager.ts
```diff
@@ -0,0 +1,175 @@
...
+  dispenser: ItemSummary;
+  robot: ItemSummary;
+}
+
+export interface ItemState {
+  doors: Record<string, RomiCore.DoorState>;
+  lifts: Record<string, RomiCore.LiftState>;
+  robots: Record<string, RomiCore.FleetState>;
+  dispensers: Record<string, RomiCore.DispenserState>;
+}
```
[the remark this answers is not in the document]
koonpeng: It is also very detrimental to performance, each item send updates at its own discretion, with 10 items sending at 10hz, it is already 100 updates a second, each update will cause a re-calculation of all items, so we are already looking at 10,000 calculations a second.

packages/dashboard/src/components/rmf-app/rmf-app.tsx
```diff
@@ -230,6 +236,41 @@ function TaskContextsProvider(props: React.PropsWithChildren<{}>): JSX.Element {
   return <TasksContext.Provider value={tasks}>{props.children}</TasksContext.Provider>;
 }
 
+function RmfHealthContextsProvider(props: React.PropsWithChildren<{}>): JSX.Element {
+  const doorStates = React.useContext(DoorStateContext);
+  const liftStates = React.useContext(LiftStateContext);
+  const dispenserStates = React.useContext(DispenserStateContext);
+  const robotStates = React.useContext(FleetStateContext);
```
koonpeng, line 243: This is very detrimental to performance, if 10 items send at 10hz, we are already doing 10,000 calculations per second.

packages/react-components/tests/systems-summary/systems-summary-spoilt-items.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+      spoiltLiftClick={liftClick}
+      spoiltDispenserClick={dispenserClick}
+      spoiltRobotClick={robotClick}
+    />,
+  );
+
+  fireEvent.click(screen.getByText('door - state'));
+  fireEvent.click(screen.getByText('lift - state'));
+  fireEvent.click(screen.getByText('dispenser - state'));
+  fireEvent.click(screen.getByText('robot - state'));
```
koonpeng, line 64: RTL recommends using `userEvent` instead.

packages/react-components/tests/systems-summary/systems-summary-task-state.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+test('expect all different task states to be calculated and rendered', () => {
+  render(<SystemSummaryTaskState tasks={tasks} onClick={jest.fn()} />);
+  // 1 task for each state, total 4 states
+  expect(screen.queryAllByText('1').length).toEqual(4);
+});
+
+test('should call onClick when panel is clicked', () => {
+  const onClick = jest.fn();
+  render(<SystemSummaryTaskState tasks={tasks} onClick={onClick} />);
+  fireEvent.click(screen.getByLabelText('panel'));
```
koonpeng, line 15: RTL recommends using `userEvent` instead.

packages/react-components/tests/systems-summary/systems-summary-alert.test.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { SystemSummaryAlert } from '../../lib';
+import { render, fireEvent, screen } from '@testing-library/react';
+import { notifications } from './test.utils';
+
+test('button should be disabled when there are no notifications', () => {
+  const root = render(<SystemSummaryAlert notifications={[]} />);
+  fireEvent.click(screen.getByRole('button'));
```
koonpeng, line 8: RTL recommends using `userEvent` instead.

packages/react-components/tests/notifications-dialog.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+test('should call onClose when close button is clicked', () => {
+  const mockOnClose = jest.fn();
+  render(
+    <NotificationsDialog
+      onClose={mockOnClose}
+      showNotificationsDialog={true}
+      notifications={notifications}
+    />,
+  );
+  fireEvent.click(screen.getByText(/CLOSE/i));
```
koonpeng, line 21: RTL recommends using `userEvent` instead.

packages/react-components/tests/notifications-dialog.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+test('should call onClose when close icon at the top is clicked', () => {
+  const mockOnClose = jest.fn();
+  render(
+    <NotificationsDialog
+      onClose={mockOnClose}
+      showNotificationsDialog={true}
+      notifications={notifications}
+    />,
+  );
+  fireEvent.click(screen.getByLabelText(/close/i));
```
koonpeng, line 34: RTL recommends using `userEvent` instead.

packages/react-components/tests/notifications-dialog.test.tsx
```diff
@@ -0,0 +1,52 @@
...
+test('should not render other severity levels when a particular level is selected', () => {
+  const mockOnClose = jest.fn();
+  render(
+    <NotificationsDialog
+      onClose={mockOnClose}
+      showNotificationsDialog={true}
+      notifications={notifications}
+    />,
+  );
+  fireEvent.change(screen.getByLabelText(/filter-input/i).childNodes[1], {
```
koonpeng, line 47: RTL recommends using `userEvent` instead.

koonpeng (changes requested):

packages/dashboard/src/components/dashboard/main-menu.tsx
```diff
@@ -52,55 +101,118 @@ export const MainMenu = React.memo((props: MainMenuProps) => {
...
-          id="commands-tooltip"
-          enabled={showTooltips}
+    <React.Fragment>
+      <SystemSummaryBanner imageSrc={'/favicon.ico'} isError={bannerIsError(healthStatus)} />
+      <div className={classes.root}>
+        <SystemSummaryAlert notifications={notifications} />
+        <Divider className={classes.divider} />
+
+        {spoiltEquipmentList.length > 0 ? (
+          <React.Fragment>
```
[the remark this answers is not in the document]
ChawinTan: There is a Divider below `SystemSummarySpoiltItems`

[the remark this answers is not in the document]
koonpeng: You are right, thanks for point out.

packages/react-components/lib/systems-summary/systems-summary-spoilt-items.tsx
```diff
@@ -0,0 +1,118 @@
...
+  spoiltItems: SpoiltItem[];
+  doors: RomiCore.Door[];
+  lifts: RomiCore.Lift[];
+  dispensers: string[];
+  robots: Record<string, RomiCore.FleetState>;
+  spoiltDoorClick?(door: RomiCore.Door): void;
+  spoiltLiftClick?(lift: RomiCore.Lift): void;
+  spoiltRobotClick?(fleet: string, robot: RomiCore.RobotState): void;
+  spoiltDispenserClick?(event: React.MouseEvent, guid: string): void;
+}
```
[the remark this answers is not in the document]
ChawinTan: But we would need to know which item and the type of item that is spoilt for the clicking to work. Do you mean replacing the `spoiltItem` array with just the list of doors and lifts etc ...?

ChawinTan:
Removed the temp logic, changes to notifications component, update tests and changed the way spoilt item is received and handled.

- Remove the dependency on the list of Romicore items (doors, lifts etc ...) and spoilt items
- Shift the fields of `SpoiltItem` into one object with the spoilt item.

This should make it easier for the backend to organize data and the front end to reason with it, I think.

packages/dashboard/src/managers/rmf-health-state-manager.ts
```diff
@@ -0,0 +1,74 @@
+import { DoorSummary, LiftSummary, DispenserSummary, RobotSummary } from 'react-components';
+
+export interface HealthStatus {
```
koonpeng, line 3: I feel the design of this could be improved, there are several problem with this

1. `ItemSummaryState.outOfOrder` seems redundant, wouldn't it be equal to `spoiltDoors.length` etc?
2. I don't think the use of `ItemSummary.item` is required, the categories are already differentiatied based on different props (door, lifts etc).
3. `ItemSummary.itemSummary` and `ItemSummary.robotSummary` are optional, and `RobotSummaryState` is a subclass of `ItemSummaryState`. This make things unnecessarily confusing and breaks many advantages of type checking. Since doors etc and robots are already differentiated by different props, it's better to just let have use different interface, e.g.
```ts
export interface RobotSummary {
  spoiltRobots: SpoiltRobot[];
  summary: RobotSummaryState;
}
```

packages/react-components/lib/notifications-dialog.tsx
```diff
@@ -0,0 +1,233 @@
...
+                    severity={notification.severity}
+                  />
+                  <Typography variant="body1" align="left">
+                    {moment(notification.time, 'MMMM Do YYYY, h:mm:ss').fromNow()}
+                  </Typography>
+                  <Typography variant="body1" align="left">
+                    {notification.error}
+                  </Typography>
+                  <Typography align="right">
+                    <IconButton className={classes.removeNotificationIcon}>
```
koonpeng, line 216: no handler for this button, related to the bug mentioned in the story.

packages/react-components/lib/systems-summary/systems-summary-item-state.tsx
```diff
@@ -0,0 +1,163 @@
...
+export interface RobotSummaryState extends ItemSummaryState {
+  idle: number;
+  charging: number;
+}
+
+export interface ItemSummary {
+  item: string;
+  itemSummary?: ItemSummaryState;
+  robotSummary?: RobotSummaryState;
+}
```
koonpeng, line 20: Design issues as mentioned in rmf-health-manager.ts

packages/react-components/lib/systems-summary/systems-summary-item-state.tsx
```diff
@@ -0,0 +1,163 @@
...
+  let totalItem = 0;
+  let operationalItem = 0;
+
+  if (itemSummary.itemSummary) {
+    totalItem = itemSummary.itemSummary.operational + itemSummary.itemSummary.outOfOrder;
+    operationalItem = itemSummary.itemSummary.operational;
+  } else if (itemSummary.robotSummary) {
+    totalItem = itemSummary.robotSummary.operational + itemSummary.robotSummary.outOfOrder;
+    operationalItem = itemSummary.robotSummary.operational;
+  }
```
koonpeng, line 89: Since robot states is different from other states, I think it's better to separate them into different components instead of changing behavior based on the type of the props.

This is also not strictly type safe as `itemSummary` and `robotSummary` could both be undefined.

packages/react-components/stories/notifications-dialog.stories.tsx
```diff
@@ -0,0 +1,93 @@
...
+      <NotificationsDialog
+        notifications={notifications}
+        showNotificationsDialog={showDialogBox}
+        onClose={() => setShowDialogBox(false)}
+      />
+    </>
+  );
+};
+
+export const NotificationsDialogDisplay: Story = () => {
```
koonpeng, line 87: Clicking on the red X (not the close X) does not do anything.

koonpeng (changes requested):

packages/dashboard/src/components/rmf-app/rmf-app.tsx
```diff
@@ -230,6 +232,16 @@ function TaskContextsProvider(props: React.PropsWithChildren<{}>): JSX.Element {
   return <TasksContext.Provider value={tasks}>{props.children}</TasksContext.Provider>;
 }
 
+function RmfHealthContextsProvider(props: React.PropsWithChildren<{}>): JSX.Element {
+  const healthManager = new RmfHealthStateManager();
```
koonpeng, line 236: This creates a new manager each render, you need to `useState` or `useMemo` here.

packages/dashboard/src/managers/rmf-health-state-manager.ts
```diff
@@ -0,0 +1,74 @@
...
+  door: DoorSummary;
+  lift: LiftSummary;
+  dispenser: DispenserSummary;
+  robot: RobotSummary;
+}
+
+export default class RmfHealthStateManager {
+  getDoorSummary = (): DoorSummary => {
```
koonpeng, line 11: I think these are stub values, add TODO and FIXME here?

packages/dashboard/src/managers/rmf-health-state-manager.ts
```diff
@@ -0,0 +1,74 @@
...
+    return {
+      item: 'Door',
+      itemSummary: { operational: 2, outOfOrder: 1 },
+      spoiltDoors: [
+        {
+          itemNameAndState: 'hardware_door - unknown',
+          door: {
+            name: 'hardware_door',
```
koonpeng, line 19: Does it cause any bugs or inconsistency if a non-existing door is named here? If so, better to just provide an empty array.

packages/dashboard/src/managers/rmf-health-state-manager.ts
```diff
@@ -0,0 +1,74 @@
...
+  lift: LiftSummary;
+  dispenser: DispenserSummary;
+  robot: RobotSummary;
+}
+
+export default class RmfHealthStateManager {
+  getDoorSummary = (): DoorSummary => {
+    return {
+      item: 'Door',
+      itemSummary: { operational: 2, outOfOrder: 1 },
```
koonpeng, line 14: I think this will cause inconsistency if  using any maps that does not have exactly 3 doors, this is not easily fixed so I might suggest to copy the new `main-menu.tsx` into a temporary `wip-main-menu-plusplus.tsx` (or similar) and revert the main menu changes.

In other words, this will be a new feature that is unused and disabled by default, the downside is that you need to ensure this temp main-menu is keep up to date with other main menu changes until it can be merged into it (when the backend is ready).

packages/dashboard/src/managers/rmf-health-state-manager.ts
koonpeng: Ignore this if you move the new main menu into a new temp file and revert the main-menu changes.

koonpeng (commented):
Replying to an earlier comment as I can't reply there for some reason,

> Can't we just use the list of door, lifts props etc instead of doing the extra filtering item with the spoiltItems prop?

> But we would need to know which item and the type of item that is spoilt for the clicking to work. Do you mean replacing the spoiltItem array with just the list of spoilt doors and lifts etc ...?

The `spoiltDoorClick` etc props separate the categories, so in theory the parent can map the item name to the object, but this mapping is extra work and is quite tricky to perform when using memoized component. Including the doors/lifts themselves makes it slightly less re-usable, but I'm resolving the conversation as this is probably a good comproise to make it easier to use.

packages/react-components/lib/systems-summary/systems-summary-spoilt-items.tsx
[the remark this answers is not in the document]
koonpeng: The `spoiltDoorClick` etc props separate the categories, so in theory the parent can map the item name to the object, but this mapping is extra work and is quite tricky to perform when using memoized component. Including the doors/lifts themselves makes it slightly less re-usable, but I'm resolving the conversation as this is probably a good comproise to make it easier to use.

packages/react-components/stories/systems-summary.stories.tsx
```diff
@@ -0,0 +1,159 @@
...
+    />
+    <SystemSummaryItemState
+      itemSummary={itemStateDataRobot.itemSummary}
+      onClick={itemStateDataRobot.onClick}
+      {...args}
+    />
+  </React.Fragment>
+);
+
+export const SystemSummaryItemAlertStory: Story = (args) => (
```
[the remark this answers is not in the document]
ChawinTan: This will be an api call to the backend once it is up. I have removed it for now since the backend is not up yet. Will a TODO comment for now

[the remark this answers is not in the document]
koonpeng: note that this being in `react-components`, it must not depend on any backend. There is currently no way for the story to handle the button because there is no handler props it can use, the story should be able to perform the needed actions even without a backend.

It can do so by using something along the lines of
```tsx
<SystemSummaryAlert notifications={notifications} onNotificationDismiss={(id) => setNotifications((prev) => prev.filter((x) => x.id !== id)} {...args} />
```

On the actual dashboard, `onNotificationDismiss` will do whatever is needed to keep things in sync (might include updating the db etc).

packages/dashboard/src/components/dashboard/tests/main-menu.test.tsx
```diff
@@ -1,9 +1,141 @@
...
+    },
+  };
+  const root = mount(
+    <RmfHealthContext.Provider value={healthStatus}>
+      <MainMenu pushView={jest.fn()} tasks={[]} notifications={[]} />
+    </RmfHealthContext.Provider>,
+  );
+  expect(
+    root.find('SystemSummaryItemState').find('.MuiPaper-root').find('h6').at(3).text(),
+  ).toEqual('1/1');
```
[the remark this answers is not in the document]
ChawinTan: It should be 2/2 my bad. An operational robot can also be either idling or charging.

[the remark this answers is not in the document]
koonpeng: yea, you are right, should be 2/2

packages/react-components/stories/systems-summary.stories.tsx
```diff
@@ -0,0 +1,159 @@
...
+      },
+      itemNameAndState: 'door - unknown',
+    },
+  ],
+  lifts: [],
+  dispensers: [],
+  robots: [],
+};
+
+export const SystemSummaryItemStateStory: Story = (args) => (
```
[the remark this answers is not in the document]
ChawinTan: The clicks on the Detail button is registered in the Actions panel of storybook. Is it meant to drill into the actual details of the items itself?

ChawinTan:
refactored rmf health status and created a new robotSummary component. Refactored stories and moved the existing main menu into a `wip` file.

Still can't get some actions to be registered in storybook though and I can't find out why despite following the conventions already. I am going to take the next couple of days to find out why

ChawinTan:
So after separating the systems-summary components into each of their individual stories, with each of their component specified in the `export default {...}` field, the controls and actions are being registered in storybook's control panel.

koonpeng (changes requested):
Can't post comments at the appropriate places,

at `system-summary-item-state.tsx`

```tsx
export interface ItemSummary {
  operational: number;
  spoiltItem: SpoiltDoor[] | SpoiltLift[] | SpoiltDispenser[];
}
```
recommend to split this into `DoorSummary` etc, using union types here like this removes the advantages of type checking, as you can easily pass a list of lifts to a spoilt door component.

---

Thanks for the hard work, just a couple of small changes and it's good to merge

ChawinTan:
separated out item summary props, change the name of robotSummaryState props and address other comments

packages/dashboard/src/components/dashboard/wip-main-menu-plusplus.tsx
```diff
@@ -0,0 +1,215 @@
...
+        <SystemSummaryAlert
+          notifications={allNotifications}
+          onNotificationsDismiss={handleDismissNotification}
+        />
+        <Divider className={classes.divider} />
+
+        <SystemSummarySpoiltItems
+          doors={healthStatus.door.spoiltItem as SpoiltDoor[]}
+          lifts={healthStatus.lift.spoiltItem as SpoiltLift[]}
+          dispensers={healthStatus.dispenser.spoiltItem as SpoiltDispenser[]}
```
koonpeng, line 129: You don't need a cast here anymore.

koonpeng:
@ChawinTan I noticed that the test coverage for this PR is very low, see https://codecov.io/gh/open-rmf/rmf-web/tree/dedfe267719b8fd18f7d455cd4d80ea9021b5726/packages/react-components/lib/systems-summary. Not sure if it is a bug or if there are really insufficient tests, let's fix it before merging it again.

PR #208, opened 2021-01-19
dashboard: allow schedulervisualier reuse

koonpeng:
## Problem
The current design has the `Dashboard` as a top level component, containing the `ScheduleVisualizer` among many other components. There is a strong coupling between `Dashboard` and `ScheduleVisualizer` which makes it very hard to re-use `ScheduleVisualizer` for other pages/layout we will have in the future.

## What's new
This PR refactor many of the components contained within `Dashboard` to mainly 2 other components:

* AppBase: This provides many of the "app level" functions the `Dashboard` previously does, things like the loading screen, settings, help drawers, notification bar etc. Various contexts are provided to let children children components read and control them.
* RmfApp: This provides contexts to rmf data, things like door states, fleet states etc. Children components can make use of these contexts to get rmf data.

These 2 components are provided in the top-top level `app.tsx` so any descendent components will get access to all the contexts.

There are also other small changes:

* `MainMenuReducer` is now renamed to `DashboardReducer`, it only provides dispatches for dashboard components, namely all the settings, help, hotkey dialog dispatches are removed, it now only has omnipanel related dispatches.
* With the dashboard (and omnipanel) no longer the whole app, the "dashboard" button is removed from the app bar. Instead a custom leaflet control is added to `Dashboard` to show/hide the omnipanel.

## Self-checks

## Discussion

[Questions for reviewers]


codebot (approved):
Thanks for this mega refactoring! Everything works for me except command sending, but that is expected due to the mega-PR adding the task dispatcher in https://github.com/osrf/rmf_demos/pull/181 so I think we should merge this as-is so that we can work in parallel on the fixes required to deal with the recent mega-PR's in rmf_core and rmf_demos :+1:

PR #213, opened 2021-01-22
Add fluentd conf

matiasbavera:
## What's new

[Describe your changes]

## Self-checks

## Discussion

[Questions for reviewers]


koonpeng:
Should we really be logging the UI? Most/all actions will hit the backend anyway so can we just log based on that?

As I think about it, log collection looks like a deployment problem, we don't control where the services run, or if they will be ran inside a container or kubernetes or other cloud infrastructure. I havent used fluetd before, but I guess the config would be different for each scenario?

The problem though is that some features require the logs, so we kind of need a standardized way to query them.

koonpeng:
Here are some scenarios I can think of

1. https://kubernetes.io/docs/concepts/cluster-administration/logging/#using-a-node-logging-agent. Suppose we deploy with kubernetes, each of the backend will be ran in a pod, they will output their logs to stdout and kubernetes will collect them in a file, fluetd will then process the files and upload them to a database (which will also be a k8s pod).
  * in this case, the logging would be a deployment problem, as far as we are concern, we just output stuff to stdout.
  * a similar variant is the "Sidecar container with a logging agent", but it doesn't change anything on the app side.
  * since some features require the logs, we need a way to query the logs, this is very tricky because we don't control the deployment, therefore we don't know the database or format of the logs. This can be sidestepped if we push the log viewer to deployment problem as well.

2. https://kubernetes.io/docs/concepts/cluster-administration/logging/#exposing-logs-directly-from-the-application. In this case, we will handle the logging directly, unlike option 1, the logging backend is part of the "system" instead of a deployment problem, we need to build support to send data to our choice of backend (or support multiple options). fluetd could be the backend, then we need a backend for fluetd itself as well.
In this case, every backend service should publish logs to fluetd themselves.
  * There should be an option in every backend to configure the fluetd address and port.
  * There should be an option to disable logging as well.
  * What if the fluetd service is not running or not contactable? Should the backend fail to start?
  * Since some features require logs, we need to launch fluetd as well in `npm start` of dashboard.
  * I don't think we can we can output the logs to a file and have fluetd watch the file ala option 1 because we don't control the deployment and we can't guarantee fluetd is ran in the same machine as the backend.

Option 1 has the advantage of being able to support any log collector/logging backend easily, suppose a deployment would like to use splunk or kibana, it can easily be done by configuring fluetd or alternative collectors like logstash. The downside is as mentioned above, we don't know the logging backend so it makes it very hard to query the logs.

In option 2, we (can) control the whole stack, that makes it possible for us to write our own log viewer. But the downside is that every backend needs to support sending logs from inside the app, it complicates things by requiring them to include logging logic.

We probably need option 2 if we are not able to find a usable log analysis tool like kibana, but if we do, then the extra complexity in option 2 doesn't seem like it would be worth the trouble.

matiasbavera:
> Should we really be logging the UI? Most/all actions will hit the backend anyway so can we just log based on that?

Yes, that's true, but I think this would help us to identify problems that happen on the interface itself, that not necessarily use the backend. Eg: "error: this could not be loaded because of a missing attribute" kind of stuff. Or maybe capturing global errors like when react explodes or when it throws you a white page, etc. I'm reading some things that we could do from this place https://www.sitepoint.com/logging-errors-client-side-apps/#capturingglobalerrors

Let me know what you think

matiasbavera:
Closing this because it's using kind

PR #216, opened 2021-01-26
ci: re-organize workflows to remove repeating

koonpeng:
## What's new

refactor the workflows to group by packages instead of events. This removes the duplicate when we need to run a check in multiple events. The downside of this is that there will always be a "build docker images" checks that is skipped for pull requests, and some changes in the status badges.

## Self-checks

## Discussion

[Questions for reviewers]


matiasbavera (approved):
This is much better organized and clearer. Thanks. LGTM.

> The downside of this is that there will always be a "build docker images" checks that is skipped for pull requests, and some changes in the status badges.

I think we can live with that, and IMO I'll be happy to pay that price to keep things organized. =)

PR #217, opened 2021-01-26
ci: update pr template

koonpeng:
## What's new

Add some guidelines on what kind of descriptions should a PR include. Make some of the templates like `fixes #` and `[Describe your changes]` a comment so that it is more obvious they should not be part of the pr message.

## Self-checks


matiasbavera (approved):
I really like these changes. Thanks

PR #218, opened 2021-01-27
dashboard: ScheduleVisualizer now uses the fleet context directly

codebot:

## What's new

The ScheduleVisualizer component now uses the fleet context directly, rather than receiving it via `props`. Then the parent component doesn't need to render every time the fleet context is updated. This is helpful for the mosaic layout manager (will come in a future PR, currently it's sitting in a branch).

## Self-checks


koonpeng (approved):
Thanks! This will make it easier to integrate ScheduleVisualizer.

PR #219, opened 2021-01-27
react-components: Add web based tests with karma

matiasbavera:
## What's new

Moved react-components tests from jest/jsdom to jasmine/karma.

Observations:

* Changed .test extension to .spec for karma to work.
* Syntax of the assertion library changed (we are using jasmine now).
* We lost the jest.fn(). But we can use an empty function and spyOn.
* Added a karma config file and a Webpack configuration file for pre-processing the project's files. There is a typescript-karma compiler but it's not working great when we deal with third-party libraries and CSS. Webpack do the pre-processing for us and it's more flexible so, I opted to go with Webpack.
* Added puppeteer to install and open a headless chrome in Github's actions server, for testing with Karma.

## Self-checks

## Discussion

[FUP PR]  Check if we can replace the smoke tests with screenshot snapshot diffing?


matiasbavera:
```
We have this warning:

It looks like there are several instances of @material-ui/styles initialized in this application.
This may cause theme propagation issues, broken class names, specificity issues, and makes your application bigger without a good reason.
```
from KP
> Since we are using a monorepo, the `@material` packages are in the root package to avoid multiple instances being loaded. Maybe we need to change the webpack/karma config so it bundles correctly? Or maybe this warning comes from the way karma load tests, resulting in each test creating its own `@material/styles` instance?


Currently looking into this

matiasbavera:
> # [Codecov](https://codecov.io/gh/osrf/rmf-web/pull/219?src=pr&el=h1) Report
> > Merging [#219](https://codecov.io/gh/osrf/rmf-web/pull/219?src=pr&el=desc) ([302141b](https://github.com/osrf/rmf-web/commit/302141b91a98ce36070aed679748ee648edc0a9e)) into [main](https://codecov.io/gh/osrf/rmf-web/commit/ba3c1604d1eedd6b38a3da394883953b6f65ed8c?el=desc) ([ba3c160](https://github.com/osrf/rmf-web/commit/ba3c1604d1eedd6b38a3da394883953b6f65ed8c)) will **decrease** coverage by `6.69%`.
> > The diff coverage is `n/a`.
> 
> [Impacted file tree graph](https://codecov.io/gh/osrf/rmf-web/pull/219?src=pr&el=tree)

About the codecov, it seems strange to me that the coverage average has dropped so much. Maybe this `itsbull` library may be generating the report in another way.

  Or maybe when converting from ts to js in webpack and then running the tests with Karma in js it generates the reports in another way (?). The latter does not make so much sense to me because in the case of jest from ts it passes to js through the ts compiler, and then the tests are run in commojs in nodejs.

koonpeng:
Maybe it's not using the sourcemap? I hear jest uses istabul under the hood as well.

matiasbavera:
Coverage test from main with jest: 

----------------------------------------|---------|----------|---------|---------|-------------------
File                                    | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------------------------------------|---------|----------|---------|---------|-------------------
All files                               |   87.12 |    65.03 |   84.62 |   86.03 |                   
 lib                                    |   80.26 |    70.71 |   73.33 |   79.89 |                   
  alert-dialog.tsx                      |     100 |    91.67 |     100 |     100 | 145               
  ant-tab.tsx                           |     100 |      100 |     100 |     100 |                   
  confirmation-alert-dialog.tsx         |     100 |      100 |     100 |     100 |                   
  css-utils.ts                          |     100 |      100 |     100 |     100 |                   
  date-time-picker.tsx                  |     100 |      100 |     100 |     100 |                   
  error-overlay.tsx                     |     100 |    16.67 |     100 |     100 | 47-66             
  geometry-utils.ts                     |   29.69 |        0 |      40 |   31.67 | ...76-209,213-215 
  index.ts                              |     100 |      100 |     100 |     100 |                   
  item-accordion-details.tsx            |     100 |      100 |     100 |     100 |                   
  item-accordion-summary.tsx            |     100 |       75 |     100 |     100 | 36                
  material-table-icons.tsx              |   67.65 |      100 |   35.29 |   67.65 | 25-31,37-38,40-41 
  omni-panel-view.tsx                   |     100 |      100 |     100 |     100 |                   
  omni-panel.tsx                        |     100 |    83.33 |     100 |     100 | 58                
  simple-info.tsx                       |   92.11 |    69.23 |     100 |   91.67 | 153-158           
  spotlight-accordion.tsx               |   96.43 |       75 |     100 |   96.15 | 76                
  stack-navigator.ts                    |   93.33 |      100 |   83.33 |   92.86 | 23                
  status-label.tsx                      |     100 |      100 |     100 |     100 |                   
  tooltip.tsx                           |     100 |      100 |     100 |     100 |                   
  ...n-control-button-group-reducer.tsx |   27.27 |    14.29 |      25 |   27.27 | 38-79,101         
  trashbin-control-button-group.tsx     |     100 |    96.15 |     100 |     100 | 56                
  utils.ts                              |     100 |      100 |     100 |     100 |                   
 lib/auth                               |     100 |      100 |     100 |     100 |                   
  index.ts                              |     100 |      100 |     100 |     100 |                   
  login-card.tsx                        |     100 |      100 |     100 |     100 |                   
  login-page.tsx                        |     100 |      100 |     100 |     100 |                   
 lib/commands                           |   96.52 |    74.38 |     100 |   95.79 |                   
  delivery-request-form.tsx             |   96.85 |    77.91 |     100 |   96.19 | 113-114,147,156   
  form-styles.ts                        |     100 |      100 |     100 |     100 |                   
  index.ts                              |     100 |      100 |     100 |     100 |                   
  loop-request-form.tsx                 |   95.96 |    70.27 |     100 |   95.06 | 72-73,103,111     
 lib/dispensers                         |   90.67 |     56.1 |   92.31 |   89.55 |                   
  default-marker-icon.tsx               |     100 |      100 |     100 |     100 |                   
  dispenser-accordion.tsx               |   86.05 |    48.48 |     100 |      85 | 39-43,100-104     
  dispenser-marker.tsx                  |   95.65 |     87.5 |      75 |   94.44 | 68                
  index.ts                              |     100 |      100 |     100 |     100 |                   
 lib/doors                              |   80.99 |       50 |     100 |   79.55 |                   
  door-accordion.tsx                    |   66.67 |    28.57 |     100 |   65.38 | ...,69-71,110-118 
  door-marker.tsx                       |   87.14 |    60.71 |     100 |   85.48 | ...68,234-240,244 
  enums.ts                              |     100 |      100 |     100 |     100 |                   
  index.ts                              |     100 |      100 |     100 |     100 |                   
 lib/lifts                              |   82.66 |    47.56 |   84.85 |   80.39 |                   
  index.ts                              |     100 |      100 |     100 |     100 |                   
  lift-accordion.tsx                    |      85 |       50 |   66.67 |   82.86 | 54,97-104,112     
  lift-marker.tsx                       |    82.5 |    57.14 |     100 |   81.58 | 14,18-22,31-33,37 
  lift-request-form.tsx                 |     100 |       80 |     100 |     100 | 88,136            
  lift-utils.ts                         |   41.38 |    11.11 |      40 |   41.38 | 4-16,21-29,34-42  
 lib/log-management                     |   87.88 |    66.67 |   64.52 |   88.99 |                   
  custom-lookup-filter.tsx              |   68.18 |    58.33 |   42.86 |   68.42 | 61-71             
  index.ts                              |     100 |      100 |     100 |     100 |                   
  log-level.ts                          |     100 |      100 |     100 |     100 |                   
  log-management.tsx                    |   85.19 |    66.67 |   66.67 |   94.44 | 42                
  log-table.tsx                         |     100 |     87.5 |     100 |     100 | 122               
  search-filter.tsx                     |     100 |      100 |     100 |     100 |                   
  search-log-form.tsx                   |   86.49 |        0 |   28.57 |   82.14 | 39,44,51,57,61    
 lib/robots                             |   92.62 |    54.84 |    96.3 |   90.82 |                   
  default-marker.tsx                    |     100 |       75 |     100 |     100 | 28,65             
  image-marker.tsx                      |     100 |       50 |     100 |     100 | 51-58             
  index.ts                              |     100 |      100 |     100 |     100 |                   
  robot-accordion.tsx                   |   74.19 |    11.11 |     100 |   72.41 | 20-30,34-36       
  robot-marker.tsx                      |   96.15 |       80 |      75 |   95.65 | 79                
  utils.ts                              |     100 |      100 |     100 |     100 |                   
 lib/task-summary                       |   93.81 |       84 |      90 |   93.18 |                   
  index.ts                              |     100 |      100 |     100 |     100 |                   
  task-summary-accordion.tsx            |   94.44 |    90.91 |      80 |   93.62 | 61,65,79          
  task-summary-utils.ts                 |   92.68 |    78.57 |     100 |   92.31 | 28,66-67          
 lib/waypoints                          |     100 |       50 |     100 |     100 |                   
  index.ts                              |     100 |      100 |     100 |     100 |                   
  waypoint-marker.tsx                   |     100 |       50 |     100 |     100 | 42-49             
----------------------------------------|---------|----------|---------|---------|-------------------

matiasbavera:
With karma + istambul

----------------------------------------|---------|----------|---------|---------|-------------------
File                                    | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
----------------------------------------|---------|----------|---------|---------|-------------------
All files                               |   77.54 |    60.41 |   78.39 |   75.44 |                   
 lib                                    |   70.37 |    70.59 |   73.91 |   68.29 |                   
  alert-dialog.tsx                      |     100 |    91.67 |     100 |     100 | 145               
  ant-tab.tsx                           |     100 |      100 |     100 |     100 |                   
  confirmation-alert-dialog.tsx         |     100 |      100 |     100 |     100 |                   
  css-utils.ts                          |     100 |      100 |     100 |     100 |                   
  date-time-picker.tsx                  |     100 |      100 |     100 |     100 |                   
  error-overlay.tsx                     |     100 |    16.67 |     100 |     100 | 47-66             
  geometry-utils.ts                     |   12.24 |        0 |      40 |   12.77 | 99-215            
  item-accordion-details.tsx            |     100 |      100 |     100 |     100 |                   
  item-accordion-summary.tsx            |     100 |       75 |     100 |     100 | 36                
  material-table-icons.tsx              |   38.89 |      100 |   35.29 |   38.89 | 25-31,37-38,40-41 
  omni-panel-view.tsx                   |     100 |      100 |     100 |     100 |                   
  omni-panel.tsx                        |     100 |    83.33 |     100 |     100 | 58                
  simple-info.tsx                       |   90.63 |    69.23 |     100 |      90 | 153-158           
  spotlight-accordion.tsx               |   95.83 |       75 |     100 |   95.45 | 76                
  stack-navigator.ts                    |   92.31 |      100 |   83.33 |   92.31 | 23                
  status-label.tsx                      |     100 |      100 |     100 |     100 |                   
  svg-text.tsx                          |   66.67 |    66.67 |     100 |   64.29 | 21-22,27-28,35    
  tooltip.tsx                           |     100 |      100 |     100 |     100 |                   
  ...n-control-button-group-reducer.tsx |      20 |    14.29 |      25 |      20 | 38-101            
  trashbin-control-button-group.tsx     |     100 |    96.15 |     100 |     100 | 56                
  utils.ts                              |     100 |      100 |     100 |     100 |                   
 lib/auth                               |     100 |      100 |     100 |     100 |                   
  login-card.tsx                        |     100 |      100 |     100 |     100 |                   
  login-page.tsx                        |     100 |      100 |     100 |     100 |                   
 lib/commands                           |   96.31 |    74.38 |     100 |   95.48 |                   
  delivery-request-form.tsx             |   96.72 |    77.91 |     100 |      96 | 113-114,147,156   
  form-styles.ts                        |     100 |      100 |     100 |     100 |                   
  loop-request-form.tsx                 |   95.74 |    70.27 |     100 |   94.74 | 72-73,103,111     
 lib/dispensers                         |   87.04 |     56.1 |   92.31 |   84.78 |                   
  default-marker-icon.tsx               |     100 |      100 |     100 |     100 |                   
  dispenser-accordion.tsx               |   82.35 |    48.48 |     100 |   80.65 | 39-43,100-104     
  dispenser-marker.tsx                  |   93.75 |     87.5 |      75 |   90.91 | 68                
 lib/doors                              |    77.5 |       50 |     100 |   75.45 |                   
  door-accordion.tsx                    |   58.14 |    28.57 |     100 |    56.1 | ...,69-71,110-118 
  door-marker.tsx                       |   85.71 |    60.71 |     100 |   83.64 | ...68,234-240,244 
  enums.ts                              |     100 |      100 |     100 |     100 |                   
 lib/lifts                              |   78.26 |    47.56 |   84.85 |   74.58 |                   
  lift-accordion.tsx                    |   78.57 |       50 |   66.67 |   73.91 | 54,97-104,112     
  lift-marker.tsx                       |   78.79 |    57.14 |     100 |   77.42 | 14,18-22,31-33,37 
  lift-request-form.tsx                 |     100 |       80 |     100 |     100 | 88,136            
  lift-utils.ts                         |   26.09 |    11.11 |      40 |   26.09 | 4-42              
 lib/log-management                     |   84.91 |    66.67 |   64.52 |   85.54 |                   
  custom-lookup-filter.tsx              |      65 |    58.33 |   42.86 |   64.71 | 61-71             
  log-level.ts                          |     100 |      100 |     100 |     100 |                   
  log-management.tsx                    |   82.61 |    66.67 |   66.67 |   92.86 | 42                
  log-table.tsx                         |     100 |     87.5 |     100 |     100 | 122               
  search-filter.tsx                     |     100 |      100 |     100 |     100 |                   
  search-log-form.tsx                   |   83.33 |        0 |   28.57 |   76.19 | 39,44,51,57,61    
 lib/robots                             |   48.04 |    18.67 |   51.02 |   43.75 |                   
  default-marker.tsx                    |   93.33 |     37.5 |      90 |   94.74 | 48                
  image-marker.tsx                      |     100 |       50 |     100 |     100 | 51-58             
  robot-accordion.tsx                   |   65.22 |    11.11 |     100 |    61.9 | 20-30,34-36       
  robot-marker.tsx                      |   94.44 |       80 |      75 |   93.33 | 79                
  trajectory-marker.tsx                 |   11.11 |        0 |       0 |   11.11 | 34-59             
  trajectory-paths.tsx                  |    8.51 |        0 |       0 |    9.09 | ...89-123,157-188 
  trajectory.ts                         |       0 |      100 |       0 |       0 | 26-71             
  utils.ts                              |     100 |      100 |     100 |     100 |                   
 lib/task-summary                       |   92.68 |       84 |      90 |   91.78 |                   
  task-summary-accordion.tsx            |   92.86 |    90.91 |      80 |   91.43 | 61,65,79          
  task-summary-utils.ts                 |    92.5 |    78.57 |     100 |   92.11 | 28,66-67          
 lib/waypoints                          |     100 |       50 |     100 |     100 |                   
  waypoint-marker.tsx                   |     100 |       50 |     100 |     100 | 42-49             
----------------------------------------|---------|----------|---------|---------|-------------------

matiasbavera:
Hi @koonpeng I finally found the difference between the jest coverage and karma + Istanbul: 

1.  I added the following files now (which decrease the coverage number)
``` javascript
coveragePathIgnorePatterns: [	
    'lib/color-manager.ts', // uses many apis not available it jsdom	
    'lib/robots/trajectory*', // uses svg methods not supported by jsdom	
    'lib/svg-text.ts', // uses computed text length which is not supported by jsdom
]
```

2. Jest assumes that all of the `index.ts` files are fully tested files, so that increment the coverage. So, it looks like we were relying on a fake coverage number.

matiasbavera:
Captain log about: 
```
We have this warning:

It looks like there are several instances of @material-ui/styles initialized in this application.
This may cause theme propagation issues, broken class names, specificity issues, and makes your application bigger without a good reason.
```

1. I isolated just the `react-component` package. With this action, I took "monorepo dependencies tree problem" out of the equation.
2. Since we have two warnings I thought that maybe importing the components on tests may cause a double import on the bundle. So, I removed all the `tsx` test, still got two warnings. So the problem is when we bundle the `/lib` package with webpack.
3. I tried to load the material library to the karma's window (similar to <script></script>). [Op1](https://stackoverflow.com/questions/22228399/karma-is-it-possible-to-load-javascript-files-statically-before-the-requirejs-f)
4. I tried bundle splitting as well to keep the material UI in a different bundle. [This article](https://survivejs.com/webpack/building/bundle-splitting/)



Random Actions:
* Removed storybook to check if the bundler was getting the files too.

koonpeng:
Thanks for investigating this, material is installed in the root package to avoid the monorepo problem, have you checked the node_modules of react-components and confirm there are no material packages inside?

Is there any complains on multiple react being loaded, if it works for react, then it must be something wrong with the material bundle? You may try to look at the difference with how react and material is bundled.

matiasbavera:
> have you checked the node_modules of react-components and confirm there are no material packages inside?

Yeah, and they are not inside.

> Is there any complaints on multiple react being loaded, if it works for react, then it must be something wrong with the material bundle?

Now that are you saying this, there are no complaints about React. It's definitely something wrong with the material bundle.

> You may try to look at the difference with how react and material is bundled.
makes sense I'll do that

matiasbavera:
Hey @koonpeng 
karma2


I created this image to explain what is happening with that warning. As you said, it is a bundle problem with that library.  So, the problem goes deeper, maybe we can create our own "library" to bundle those tests and connect those to karma. But I think is time to ask, should I keep digging to fix it or we should ignore it, the downside of keeping this is that we will increase the size of each bundle but since this is not a production bundle, we could afford to have this extra library loaded. If this package could grow exponentially, this will cost us extra time to compute the bundle.

koonpeng:
Thanks for the diagram, so the problem is with how webpack decides to optimize the shared bundles, I wonder if there is any way to control how webpack make the shared bundled. Thanks for the reproducible repo, I will take a look at it and see if I can find any solution.

I'm not really worried about the bundle size, as mentioned, it is not a production bundle and everything should be served on localhost, I'm more worried that it causes styling and class name issues, that would invalidate our test cases as they are not running material ui "correctly".

koonpeng:
> This project is a framework and preprocessor for Karma that combines test files and dependencies into 2 shared bundles and 1 chunk per test file. It relies on webpack to generate the bundles/chunks and to keep it updated during autoWatch=true.

I think this is only for the next version of karma-webpack, the current release does not have this statement https://github.com/ryanclark/karma-webpack/tree/v4.0.2. If my understanding is correct, it should work as it combines dependencies (material-ui) into 1 bundle and one chunk per test files.

The current version seems to work differently, the webpack message has this
```
[./node_modules/@material-ui/core/esm/Typography/index.js] 39 bytes {tests/confirmation-alert-dialog.spec} {tests/problematic.spec} [built]
```
from what I learn from the docs, it means that the `Typography` module (and other material modules) is bundled into `tests/confirmation-alert-dialog.spec` and `tests/problematic.spec` chunks, this would explain why there are duplicate instances being loaded.

Digging into the docs more, I found out that webpack has a plugin to address this https://webpack.js.org/guides/code-splitting/#splitchunksplugin, from what I understand, it splits duplicated dependencies into its own bundle, aka it should split material-ui into its own bundle but it doesn't work when I follow the instructions to use it. Digging into the source code for `karma-webpack` I found this

```js
    // For webpack 4+, optimization.splitChunks and optimization.runtimeChunk must be false.
    // Otherwise it hangs at 'Compiled successfully'
    if (webpackOptions.optimization) {
      webpackOptions.optimization.splitChunks = false;
      webpackOptions.optimization.runtimeChunk = false;
    }
```

it explicitly disables chunk splitting because of some issues.

As for the solution for this situation, I guess we could try and see if a webpack bug or part of the design, if it is a bug, maybe it has been fixed and we can remove the check in `karma-webpack`? Or we could wait/update to the dev version, which looks like it reworked the webpack integration and should "just work" if my understanding of the statement in the docs is correct?

matiasbavera:
> As for the solution for this situation, I guess we could try and see if a webpack bug or part of the design, if it is a bug, maybe it has been fixed and we can remove the check in `karma-webpack`? Or we could wait/update to the dev version, which looks like it reworked the webpack integration and should "just work" if my understanding of the statement in the docs is correct?

Oh, you're right, that solved the problem. I'll update to webpack 5 now. Thank you.

matiasbavera:
To solve a module resolution problem with mjs files I added this:

 "resolutions": {
    "@babel/runtime": "^7.12.18"
  }

https://github.com/webpack/webpack/issues/11467

matiasbavera:
I updated to webpack 5. Now we have a warning when we create the bundle due `awesome-typescript-loader`

Warning: 
```

BREAKING CHANGE: No more changes should happen to Compilation.assets after sealing the Compilation.
        Do changes to assets earlier, e. g. in Compilation.hooks.processAssets.
        Make sure to select an appropriate stage from Compilation.PROCESS_ASSETS_STAGE_*.
```
This warning happened with other libraries/loaders as well. I think we can wait for them to update the library.

As an alternative, we use `ts-loader` I already tested and it's working just fine.

The reason I would stick with `awesome-typescript-loader` is because it is much faster, creating the bundle than ts-loader

.github/workflows/react-components.yml
```diff
@@ -29,8 +30,10 @@ jobs:
         working-directory: .
       - name: build
         run: npm run build:storybook && npm run build
+      - name: Setup kernel for Karma tests, increase watchers
+        run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
       - name: test
-        run: npm run test:coverage
```
koonpeng, line 33: I think this is a mistake from the merge.

.github/workflows/react-components.yml
```diff
@@ -8,6 +8,7 @@ on:
       - main
 env:
   CI: true
+  NODE_OPTIONS: "--max-old-space-size=8192"
```
koonpeng, line 11: Can you explain why this is needed? It can mean there are memory leaks if we need to resort to using this.

packages/react-components/package.json
```diff
@@ -10,36 +10,55 @@
     "build:watch": "tsc --build --watch",
     "build:storybook": "tsc --build stories && ../../node_modules/.bin/build-storybook -s stories",
     "clean": "tsc --build --clean",
-    "test": "jest --watch",
-    "test:coverage": "jest --ci --coverage",
+    "test": "./node_modules/karma/bin/karma start karma.conf.js --singleRun",
+    "test:watch": "./node_modules/karma/bin/karma start karma.conf.js",
```
koonpeng, line 14: I think we can use `karma` directly, scripts are ran with the bin PATH added.

packages/react-components/tests/robots/robot-marker.spec.tsx
```diff
@@ -0,0 +1,63 @@
...
+import { makeRobot } from './test-utils';
+
+describe('robot-markers', () => {
+  let colorManager: ColorManager;
+  let fakeOnClick: ReturnType<typeof jasmine.createSpy>;
+
+  beforeEach(() => {
+    colorManager = new ColorManager();
+    fakeOnClick = jasmine.createSpy();
+    colorManager.robotPrimaryColor = jasmine.createSpy();
```
koonpeng, line 14: I don't think we need a mock here anymore as karma should support the apis needed, you may add a FIXME and revisit it in another PR if it doesn't work.

koonpeng (commented):
Thanks for this PR, it must be a lot of work to figure out the compatibility between jest and karma.

.github/workflows/react-components.yml
matiasbavera: Oh sorry, the comment went away with the rebase. Sure, yeah it can mean that, but in this case, I think it's just a node 12 "feature" the default it's 512mb. From node 12 > it configures the heap limits based on more sensible defaults. Regarding the 8192, I just set the maximum number for not having problems in the future. Or we can reduce the number and then increase due to need.

I'll try to run it without this option now (I may consume fewer resources now that we fix the material-ui bundle)

packages/react-components/package.json
matiasbavera: I'll try it, I had to install it globally. To make it work on my pc.

.github/workflows/react-components.yml
matiasbavera: There is nothing here, maybe you looked at this in the outdated version?

packages/react-components/package.json
matiasbavera: It worked!

.github/workflows/react-components.yml
matiasbavera: It's working without the flag now.

.github/workflows/react-components.yml
koonpeng: Maybe it's because I did a merge main, `run: npm run test:coverage` has been changed back to `run: npm run test`, we should be using the coverage script.

.github/workflows/react-components.yml
```diff
@@ -29,8 +30,10 @@ jobs:
         working-directory: .
       - name: build
         run: npm run build:storybook && npm run build
+      - name: Setup kernel for Karma tests, increase watchers
+        run: echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p
```
koonpeng, line 33: is this really needed? it should just run once and finish in ci so it shouldn't need to watch for changes.

.github/workflows/react-components.yml
matiasbavera: Oh sorry, I misunderstood what you told me =/. Yeah, the `npm run test` now has the coverage too. I can create another karma file run all test (without the coverage) and switch the name to the one that we have now to `test:coverge`

.github/workflows/react-components.yml
matiasbavera: No, looks like we don't, this is a leftover same as  --max-old-space-size. It was fixed when we solved the material-ui bundle problem

koonpeng (approved):

packages/react-components/package.json
```diff
@@ -10,36 +10,55 @@
     "build:watch": "tsc --build --watch",
     "build:storybook": "tsc --build stories && ../../node_modules/.bin/build-storybook -s stories",
     "clean": "tsc --build --clean",
-    "test": "jest --watch",
-    "test:coverage": "jest --ci --coverage",
+    "test": "karma start karma.conf.js --singleRun",
```
koonpeng, line 13: Maybe basic question, how does coverage work for watch mode? Does it patch the coverage data every run?

Should we keep the previous impl by having "test" be watch mode, unless the "CI" env is set, which it then change to single run mode? Or pass the "--singleRun" arg separately? aka `npm run test -- --singleRun`.

matiasbavera: > Maybe basic question, how does coverage work for watch mode? Does it patch the coverage data every run?

Yeah! I'll separate the coverage

> Should we keep the previous impl by having "test" be watch mode, unless the "CI" env is set, which it then change to single run mode?

This makes more sense. I'll do that. Thanks

PR #220, opened 2021-01-28
add note about using the docker group and perhaps rebooting

codebot:

## What's new
Just an update to the installation README about adding the user to the `docker` group.

## Self-checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,11 +11,12 @@ Install docker and docker-compose
 sudo apt update && sudo apt install docker.io docker-compose
 ```
 
-To avoid having to supply an admin password every time docker containers start (and to avoid losing the password prompt in the massive console log stream if using a window manager that doesn't support pop-up privilege escalation), we suggest adding your user to the `docker` group:
-```bash
-sudo usermod -aG docker $USER
-```
-You'll need to logout/login to reload your group memberships. NOTE: on some window managers like `i3`, you may need to reboot the machine in order to reload your group memberships. Not sure why. You can use the `groups` command to verify that you are in the `docker` group before proceeding.
+Note that starting docker containers requires superuser privileges.
+If you're using GNOME or KDE, there will be a pop-up window to ask for privilege escalation when running the Docker container which runs Keycloak, the authentication mechanism we are currently using.
+If you're using `i3` or other "unusual" window managers, this pop-up may not occur, which can be confusing since a password prompt can be easily lost in the console text stream.
+Although it is possible to allow Docker to start without a privilege-escalation password by adding your user account to the `docker` group, this convenience has security implications, because it's essentially allowing password-less sudo.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -11,6 +11,12 @@ Install docker and docker-compose
 sudo apt update && sudo apt install docker.io docker-compose
 ```
 
+To avoid having to supply an admin password every time docker containers start (and to avoid losing the password prompt in the massive console log stream if using a window manager that doesn't support pop-up privilege escalation), we suggest adding your user to the `docker` group:
+```bash
+sudo usermod -aG docker $USER
+```
+You'll need to logout/login to reload your group memberships. NOTE: on some window managers like `i3`, you may need to reboot the machine in order to reload your group memberships. Not sure why. You can use the `groups` command to verify that you are in the `docker` group before proceeding.
```
matiasbavera, line 18: Just an idea, maybe you could add the source in case someone is interested in learning more about it (?) https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user

matiasbavera (approved):

koonpeng:
maybe its just me, I don't feel good suggesting people to add users to the docker group, it gives about as much privileges as passwordless sudo. Instead we can just point out that option.

For the case to update the groups, i feel it depends on too much factors maybe just give a vague statement like "you may need to logout/in or restart depending on your os and enviornment" that is how i seen some docs put it.

README.md
```diff
@@ -11,6 +11,13 @@ Install docker and docker-compose
 sudo apt update && sudo apt install docker.io docker-compose
 ```
 
+Note that starting docker containers requires superuser privileges.
+If you're using GNOME or KDE, there will be a pop-up window to ask for privilege escalation when running the Docker container which runs Keycloak, the authentication mechanism we are currently using.
+If you're using `i3` or other "unusual" window managers, this pop-up may not occur, which can be confusing since a password prompt can be easily lost in the console text stream.
+Although it is possible to allow Docker to start without a privilege-escalation password by adding your user account to the `docker` group, this convenience has security implications, because it's essentially allowing password-less sudo.
```
koonpeng, line 17: How about this?

---

You can add yourself to the `docker` group to start without requesting for password.

```
sudo usermod -aG docker <your-user>
```
You may need to logout/login or restart your system depending on your OS/environment.

Keep in mind that this has security implications, this tradeoff is described in more detail in the Docker documentation:
https://docs.docker.com/engine/install/linux-postinstall/#manage-docker-as-a-non-root-user.

codebot: Thank you for the feedback. I have incorporated this (I think) in the commit below.

codebot: @koonpeng are you OK with the latest changes?

koonpeng (approved):
lgtm

Issue #206, opened 2021-01-15
Add a flag for the user to decide to show trajectory conflicts and negotiations


matiasbavera:
The idea is to have trajectory conflicts and negotiations completely hidden except when the user specifically opens some kind of panel to look at that information (or maybe activate a flag). We don't generally need to report failed negotiations. 

Regarding repeatedly failing negotiations
We could have a separate error channel for escalating repeatedly failing negotiations to the operator's attention, and that message could be channeled into an operator notification.

koonpeng:
I think we should always show trajectory conflicts, but not the snack bar alert. Trajectory conflicts are considered normal operations in rmf so it should not raise any attention. Only when a robot failed to resolve the conflict after a certain amount of time then we consider it abnormal, but this requires a server to keep track of that data.

For negotiation, I thought the original design is to only show it when the negotiation panel is opened. There shouldn't be a need to show the negotiation trajectories layer checkbox in leaflet. It will always show when the user is in the negotiation panel and be hidden otherwise, all normal trajectories will also be hidden when we are showing negotiation trajectories. It looks like the design never got in so this change would be very welcomed.

matiasbavera:
Ok, thanks for the feedback. I'll update the description.






> For negotiation, I thought the original design is to only show it when the negotiation panel is opened. There shouldn't be a need to show the negotiation trajectories layer checkbox in leaflet. It will always show when the user is in the negotiation panel and be hidden otherwise, all normal trajectories will also be hidden when we are showing negotiation trajectories. It looks like the design never got in so this change would be very welcomed.

Yeah, that was the idea. Oh yeah, the negotiation trajectories are showed on top of the current trajectories, only when you click on a negotiation item (on the Omni-panel). So, if I understand correctly, the idea is to remove the current trajectory from the map and just show the negotiation when a negotiation is clicked, right?

matiasbavera:
> I think we should always show trajectory conflicts, but not the snack bar alert.

Got it. I think we should remove the snack alert when we implement the "grid" UI. But we can hide it with a flag for now

koonpeng:
fixed by #222

PR #222, opened 2021-02-03
Refactor/negotiation trajectory

ChawinTan:
## What's new

#206 - following the discussion on removing negotiation trajectories in the leaflet overlay checkbox

https://user-images.githubusercontent.com/29136440/106719686-91cb7380-663d-11eb-8367-4687f8b629cd.mp4

- removed the negotiations trajectory option in the leaflet overlay checkbox
- only show negotiation trajectory when user is in the negotiations panel and hide the normal trajectories
- when user navigates out of the negotiations panel, go back to showing normal trajectories

Showing the different trajectories is done by passing the current index of the omni panel to visual schedulizer. If the index matches the index of negotiations panel, the existing trajectories will be cleared.

Two other minor issues:
- home button currently requires 2 clicks to go back to current main menu. 
- the close button currently still opens to existing view despite calling `resetView`.

Solved that by creating a new object in `dashboard reducer` with the current view set to the result of `stack.stackNavigator.reset()`

Right now, users still have to click on the negotiation to show the negotiations trajectory. Is there any intention to automatically select the latest accepted negotiations that removes the need for clicking?

## Self-checks


koonpeng:
Thanks, I like how it looks, does this PR also fix the new loop and delivery request changes?

> Two other minor issues:
>
> home button currently requires 2 clicks to go back to current main menu.
> the close button currently still opens to existing view despite calling resetView.

These are non related issues right? I guess they are introduced when I did the schedule visualizer/dashboard refactor.

ChawinTan:
> Thanks, I like how it looks, does this PR also fix the new loop and delivery request changes?
> 
> >

Nope, I am on an old version of RMF demos



> These are non related issues right? I guess they are introduced when I did the schedule visualizer/dashboard refactor.

Yup, they are not related. But the UI does not look in sync so I guess they are indirectly related

packages/dashboard/src/components/schedule-visualizer/index.tsx
```diff
@@ -195,17 +196,22 @@ export default function ScheduleVisualizer(props: ScheduleVisualizerProps): Reac
           },
         });
         debug('set trajectories');
-        setTrajectories((prev) => ({
-          ...prev,
-          [curMapFloorLayer.level.name]: resp,
-        }));
+        // negotiations panel index is 6 on the omni panel
+        if (omniPanelView === 6) {
```
matiasbavera, line 200: Can we use some sort of enum here to avoid hardcoding the index position?

matiasbavera (commented):
Thanks for these changes, looks good. Left a comment.

packages/dashboard/src/components/schedule-visualizer/index.tsx
ChawinTan: Re-factored out the enum

packages/dashboard/src/components/schedule-visualizer/index.tsx
```diff
@@ -43,6 +44,7 @@ export interface MapFloorLayer {
 export interface ScheduleVisualizerProps extends React.PropsWithChildren<{}> {
   buildingMap: RomiCore.BuildingMap;
   negotiationTrajStore: Readonly<Record<string, NegotiationTrajectoryResponse>>;
+  omniPanelView: number;
```
koonpeng, line 47: This coupled it with the omnipanel, instead, leave the logic whether to show/hide trajectories to `Dashboard`, which has knowledge of both `OmniPanel` and `ScheduleVisualizer` to make that decision.

ChawinTan: Used a boolean check at the dashboard instead

packages/dashboard/src/components/schedule-visualizer/index.tsx
```diff
@@ -44,7 +43,7 @@ export interface MapFloorLayer {
 export interface ScheduleVisualizerProps extends React.PropsWithChildren<{}> {
   buildingMap: RomiCore.BuildingMap;
   negotiationTrajStore: Readonly<Record<string, NegotiationTrajectoryResponse>>;
-  omniPanelView: number;
+  showTrajectories: boolean;
```
koonpeng, line 46: Suggest to make this optional and default to `true`.

koonpeng (changes requested):

packages/dashboard/src/components/schedule-visualizer/index.tsx
ChawinTan: set it to optional

Added a check for showTrajectories === undefined which will result in the default behaviour when showTrajectory is true

koonpeng (approved):

PR #227, opened 2021-02-15
dashboard: cleanups

koonpeng:
## What's new

This started out as an attempt to use kubernetes for e2e tests, however, the workflow currently used is not compatible with it, mainly because of the dependency on rmf.

The main reason is because the goal of the e2e tests is to test the dashboard against a rmf deployment, since ros2-bridge depends on rmf, we need to build it on the same environment as the rmf deployment, since we don't do versioning, we want to run the tests on the latest commit (or whatever version the user has installed), that means requiring to build rmf from source. 

This is incompatible with how containers work, we can't build ros2-bridge on the host and copy the output to the container, as we can't guarantee that all the system and ros2 libraries used are the same versions between the host and container. We also can't build ros2-bridge inside a container as that would mean pinning the rmf version, remember we want to run the tests on whatever version of rmf you have installed.

As a result, I concluded that it is not feasible for us to run e2e with kubernetes, unless we make some fundamental changes to how we depend on rmf.

With that explained, there are multiple changes I did during the experiment, this PR is a collection of some of the things I changed:

* possibly fix dockert not working on some tiling managers
* remove REACT_APP_MOCK launch type
  * there have been lots of new features that requires an actual running rmf, using mocks is no longer a good replacement.
* remove docker launch type
  * similar to the kubernetes problem, and the reason we changed the CI from "running tests in docker" to "running the test environment in docker". If desired, you should switch to running the whole dev environment inside docker.
* use the standard PUBLIC_URL env instead of the non-standard REACT_APP_BASE_URL.
* loads `AppIntrinsincs` only when rendering app routes
  * previous code has a bug where the appbar is loaded on the not found page because it is possible to be authenticated and navigate to a not found url.
* remove some very large snapshot tests which are not useful

## Self-checks


packages/dashboard/src/components/tests/app.test.tsx
```diff
@@ -10,6 +10,6 @@ jest.mock('./../schedule-visualizer', () => () => null);
 
 it('renders correctly', async () => {
   await act(async () => {
-    expect(mount(<App />)).toMatchSnapshot();
+    expect(() => mount(<App />)).not.toThrow();
```
matiasbavera, line 13: Just FYI @vallq  this may cause you conflicts if it gets merged before your PR

packages/dashboard/src/components/dashboard/tests/negotiations-panel.test.tsx
```diff
@@ -124,24 +123,6 @@ it('renders negotiations correctly', () => {
...
-      negotiationStatusUpdateTS={0}
-      setNegotiationTrajStore={setNegotiationTrajStore}
-    />,
-  );
-
-  expect(toJson(root)).toMatchSnapshot();
-
-  root.unmount();
-});
-
```
matiasbavera, line 144: @vallq this too

packages/dashboard/src/components/auth/tests/private-route.test.tsx
```diff
@@ -26,14 +28,18 @@ describe('PrivateRoute', () => {
...
+      mount(
+        <AuthenticatorContext.Provider value={new FakeAuthenticator()}>
+          <Router history={history}>
+            <Switch>
+              <PrivateRoute path="/private" exact />
+            </Switch>
+          </Router>
+        </AuthenticatorContext.Provider>,
+      );
+    });
```
matiasbavera, line 42: @vallq  here too, but as you can notice it has a context provider now.

Context: Val is replacing Enzyme for RTL.

matiasbavera (commented):
Code LGTM, but there is a `pipenv` error on the GH runner. This probably just happened after the merge with the backend.

```
> root@undefined install /__w/rmf-web/rmf-web
> python3 -m pipenv install -d

/usr/bin/python3: No module named pipenv
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! root@undefined install: `python3 -m pipenv install -d`
```

matiasbavera:
Should we have a mock on the backend now? I mean, mocked endpoints? or we should stick to the proper backend now and on.

koonpeng:
> Should we have a mock on the backend now? I mean, mocked endpoints? or we should stick to the proper backend now and on.

The problem with mock backends is that you are only really testing the frontend aspect, in most cases, it is possible to decouple the frontend from the backend (e.g. receiving trajectory as a prop instead of querying the server). For other cases, I think it's still very useful to use some mocks for tests, but I think running the whole dashboard with a mock is quite a big challenge, it takes quite a lot of effort to make the mock behave similarly to the actual implementation, not to mention the work needed to keep it updated and maintained.

Maybe we could keep the mock env, but not advertise it in the README as it is not updated and maintained? Or maybe we can experiment with using a "mock server" instead of in-browser mocks? The problem though is as mentioned before, the mock server needs to mock most/all functions of rmf, including state updates, receiving commands, performing negotiation, assigning tasks and also non-rmf functions like sending alerts, authentication, db queries etc.

matiasbavera:
> The problem with mock backends is that you are only really testing the frontend aspect, in most cases, it is possible to decouple the frontend from the backend (e.g. receiving trajectory as a prop instead of querying the server). For other cases, I think it's still very useful to use some mocks for tests, but I think running the whole dashboard with a mock is quite a big challenge, it takes quite a lot of effort to make the mock behave similarly to the actual implementation, not to mention the work needed to keep it updated and maintained.
> 
> Maybe we could keep the mock env, but not advertise it in the README as it is not updated and maintained? Or maybe we can experiment with using a "mock server" instead of in-browser mocks? The problem though is as mentioned before, the mock server needs to mock most/all functions of rmf, including state updates, receiving commands, performing negotiation, assigning tasks and also non-rmf functions like sending alerts, authentication, db queries etc.

Yeah, you're right, it's one more thing to maintain, we can rely on the storybook then, to try and test the components there. The time would tell us if we will need it or not, but we will be ok without them.

matiasbavera (approved):
code LGTM, thanks for this.

PR #230, opened 2021-02-23
react-components: refactor omnipanel

koonpeng:
## What's new

There was some omnipanel bugs discovered by @ChawinTan previously, I don't fully understand it but hopefully this refactor means it harder to make mistakes when using the omnipanel.

The main change is that instead of taking a view id as the prop, it takes an array (stack) of view ids. A helper hook `useStackNavigator` is added to make it easier to manage the stack, but even without using the hook, the fact that it takes in an array of view should make it more obvious that the views should be managed "as a set".

Also fix some very minor gui glitches around border clipping and make the css more responsive to flexible layouts.

## Self-checks


matiasbavera (approved):
Ok, yeah it's less error-prone now. LGTM.

PR #234, opened 2021-02-25
api-server: share venvs

koonpeng:
## What's new

monorepo support in python is not as robust as nodejs, in nodejs, we can easily have multiple versions of a dependencies and share dependencies between packages in a monorepo.

python tools on the other hand, does not usually support such use cases, so there are 2 ways we can manage python dependencies,

1. each package includes all of it's dependencies
2. have the root package include dependencies of all packages

Option 1 has the disadvantage that devdependencies like black, isort will have to be duplicated in every package. It also forces us to manage each package's black, isort version etc separately, this is particularly problematic for "beta" packages like black where each release does not guarantee compatibility. We may end up depending on multiple versions of black and it gets extra complicated when we add commit hooks which will now have to determine which version of black a package should be using and run the hooks with the correct versions.

Option 2 solves that problem but it has its own disadvantages. All packages have to share dependencies, we can't have package A depend on foo@1 and package B depend on foo@2.

## Self-checks


matiasbavera:
> Option 2 solves that problem but it has its own disadvantages. All packages have to share dependencies, we can't have package A depend on foo@1 and package B depend on foo@2.

Yeah, you're right, that could be a problem, but still, I think I prefer this option too.


Also, tests are failing 

```
Run . /opt/rmf/setup.bash

> api-server@0.0.0 lint /__w/rmf-web/rmf-web/packages/api-server
> python3 -m pipenv run pylint --ignore=mixins api_server

Creating a virtualenv for this project...
Pipfile: /__w/rmf-web/rmf-web/Pipfile
Using /usr/bin/python3.8 (3.8.5) to create virtualenv...
created virtual environment CPython3.8.5.final.0-64 in 570ms
  creator CPython3Posix(dest=/__w/rmf-web/rmf-web/.venv, clear=False, no_vcs_ignore=False, global=False)
  seeder FromAppData(download=False, pip=bundle, setuptools=bundle, wheel=bundle, via=copy, app_data_dir=/github/home/.local/share/virtualenv)
    added seed packages: pip==21.0.1, setuptools==52.0.0, wheel==0.36.2
  activators BashActivator,CShellActivator,FishActivator,PowerShellActivator,PythonActivator,XonshActivator


Successfully created virtual environment!
Virtualenv location: /__w/rmf-web/rmf-web/.venv
Error: the command pylint could not be found within PATH or Pipfile's [scripts].
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! api-server@0.0.0 lint: `python3 -m pipenv run pylint --ignore=mixins api_server`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the api-server@0.0.0 lint script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.
npm WARN Local package.json exists, but node_modules missing, did you mean to install?
```

matiasbavera (approved):
The new configuration LGTM, there is a conflict with the piplock. After solving that should be ready to go.

QQ: did you merged this PR with the door heartbeat one or did you rebase it?

Issue #237, opened 2021-02-26
Test Error with Robot trajectories overlay


vallq:
Current Test with Assertion causes an error
`robot-trajectories-overlay.test.tsx`
```
it('smoke test', async () => {
    const settings = defaultSettings();
    let root: RenderResult;
    await waitFor(async () => {
      root = render(
        <LMap>
          <SettingsContext.Provider value={settings}>
            <RobotTrajectoriesOverlay
              bounds={mapBound}
              robots={robots}
              trajectories={[trajectoryValue]}
              conflicts={trajectoryConflict}
            />
          </SettingsContext.Provider>
        </LMap>,
      );
      screen.debug();
      expect(root.findAllByTestId('trajMarker')).toBeInTheDocument();
      root.unmount();
    });
  });
```
Start of Error Message
```
console.error
    Error: Uncaught [TypeError: pathAnim.getTotalLength is not a function]
        at reportException (/home/val/dev/rmf_web/src/rmf-web/packages/dashboard/node_modules/jsdom/lib/jsdom/living/helpers/runtime-script-errors.js:62:24)
```

The test will continue to run without stopping (time taken keeps increasing)

screen.debug
```
console.log
    <body>
      <div>
        <div
          class="leaflet-container leaflet-grab leaflet-touch-drag"
          style="position: relative;"
          tabindex="0"
        >
          <div
            class="leaflet-pane leaflet-map-pane"
            style="left: 0px; top: 0px;"
          >
            <div
              class="leaflet-pane leaflet-tile-pane"
            />
            <div
              class="leaflet-pane leaflet-shadow-pane"
            />
            <div
              class="leaflet-pane leaflet-overlay-pane"
            />
            <div
              class="leaflet-pane leaflet-marker-pane"
            />
            <div
              class="leaflet-pane leaflet-tooltip-pane"
            />
            <div
              class="leaflet-pane leaflet-popup-pane"
            />
          </div>
          <div
            class="leaflet-control-container"
          >
            <div
              class="leaflet-top leaflet-left"
            >
              <div
                class="leaflet-control-zoom leaflet-bar leaflet-control"
              >
                <a
                  aria-label="Zoom in"
                  class="leaflet-control-zoom-in"
                  href="#"
                  role="button"
                  title="Zoom in"
                >
                  +
                </a>
                <a
                  aria-label="Zoom out"
                  class="leaflet-control-zoom-out"
                  href="#"
                  role="button"
                  title="Zoom out"
                >
[...]

PR #244, opened 2021-03-05
dashboard: fix tests by using a fake marker

koonpeng:
## What's new

might be related to #237, also added a path to github workflows to run dashboard tests when react-component changes, this should help prevent similar issues in the future.

## Self-checks


koonpeng:
merging this to unblock ci

PR #245, opened 2021-03-05
dashboard: stub auth

koonpeng:
## What's new

Adds a stub authenticator and makes it the default, also change the CI workflow to be hopefully more user friendly.

There are some environment variables that control how the tests.

| Key | Description |
|---|---|
| E2E_DOCKER_NETWORK | The network that services uses, defaults to `rmf-web_default` |
| E2E_NO_AUTH | Do not launch the authentication provider service |
| E2E_NO_DASHBOARD | Do not launch the dashboard server |
| E2E_NO_ROS2_BRIDGE | Do not launch the ros2-bridge server |
| E2E_USER | The user to login with |
| E2E_PASSWORD | The password to login with |
| E2E_DASHBOARD_URL | Base url where the dashboard is hosted |

There are also some environment variables the test sets by default

| Key | Default Value |
|---|---|
| REACT_APP_AUTH_PROVIDER | keycloak |
| REACT_APP_KEYCLOAK_CONFIG | { "realm": "master", "clientId": "romi-dashboard", "url": "http://localhost:8088/auth" } |
| REACT_APP_TRAJECTORY_SERVER | ws://localhost:8006 |
| REACT_APP_ROS2_BRIDGE_SERVER | ws://localhost:50002 |
| E2E_USER | admin |
| E2E_PASSWORD | admin |
| E2E_DASHBOARD_URL | http://localhost:5000 |

You can overwrite them by setting them in your environment variables.

By using the environment variables, you should be able to control how the e2e tests are ran with more granularity. The default settings are choosen to make it behave the same as how the current e2e launches. This pattern should be make it easier to run e2e tests in differrent scenario, e.g. we might want to use a staging keycloak server for futher tests, or if we support other auth providers, especially SaaS providers like auth0 that requires the user to provide their own settings, which may include secrets.

Failing unit tests should be fixed by #244, the number of tries in CI is also reduced from 3 to 2, I figured that should be enough.

## Self-checks


packages/dashboard/e2e/README.md
```diff
@@ -1,4 +1,34 @@
...
+# Settings
+
+There are some environment variables that control how the tests.
+
+| Key | Description |
+|---|---|
+| E2E_DOCKER_NETWORK | The network that services uses, defaults to `rmf-web_default` |
+| E2E_NO_AUTH | Do not launch the authentication provider service |
+| E2E_NO_DASHBOARD | Do not launch the dashboard server |
+| E2E_NO_ROS2_BRIDGE | Do not launch the ros2-bridge server |
```
matiasbavera, line 10: Maybe add the value type?

packages/dashboard/e2e/README.md
```diff
@@ -13,24 +43,6 @@ Below is a diagram representing the flow of commands when running `npm run test:
...
-Broadly speaking, there are 3 major parts of the system:
-
-- Romidashboard container, built from the `docker.pkg.github.com/osrf/rmf-web/e2e` image with the RMF backend compiled into it. It holds the ROS2 bridge server and the React dashboard. Sourcing of RMF is required before running the tests and it can be sourced from `. /opt/rmf/setup.bash`
-- Auth container
-- auth_network, a known network where the Auth container is reachable 
-
-Requests from the romidashboard to the auth container would be routed to the gateway ip address of `auth_network`. Docker would then perform a NAT to route the requests to the auth container and vice versa.
+The key difference between running the tests locally and in github workflows is that in github, the tests are ran from inside a container with docker-beside-docker. So the docker commands connects to the host daemon. The main difference this cause is that "localhost" in github refers to the container, but "localhost" in local runs refers to the host (where the docker daemon is running), as such, the tests wouldn't be able to connect to the auth service because when you "expose" or "publish" a port, it is mapped to the host, not the container.
```
matiasbavera, line 46: nit suggestions 

I changed some wording:

So, the docker commands connect to the host daemon,  this causes that the "localhost" in GitHub refers to the container. "Localhost" in local runs refers to the host (where the docker daemon is running). Tests in local runs wouldn't be able to connect to the auth service because when a port is "exposed" o "published" is mapped to the host and not the container.

HTH,

packages/dashboard/src/app-config.ts
```diff
@@ -23,23 +24,33 @@ export const appConfig: AppConfig = (() => {
...
-  const authConfig: KeycloakConfig = (() => {
-    if (!process.env.REACT_APP_AUTH_CONFIG) {
-      throw new Error('REACT_APP_AUTH_CONFIG env variable is needed but not defined');
+  const authenticator = (() => {
+    if (!process.env.REACT_APP_AUTH_PROVIDER) {
+      return new StubAuthenticator();
+    }
+    // it is important that we do not do any processing on REACT_APP_AUTH_PROVIDER so that webpack
+    // can remove dead code, we DO NOT want the output to have the stub authenticator even if
+    // it is not used.
```
matiasbavera, line 33: Yeah, you're right. Good comment

matiasbavera (commented):
Code looks good to me! It's way cleaner. It's great that the auth provider can be easily configurable. 

Left a couple of comments

matiasbavera (approved):
LGTM!

PR #246, opened 2021-03-08
react-component:reporting:rmf-auth:Dashboard UI for static reports and split auth from Dashboard

### LINKED ISSUE #252: Create reporting interface for statics reports
## Feature request
The idea of this is to create a reporting interface with a static structure to show them to the operators.

### Implementation considerations

Some reports we should consider for the UI design:
* robot states
* doors states
* lift states
* charger states
* workcell states
* negotiations 
* robot motion plans
* robot actions
* user actions in UI
* Tasks
* user logins/logouts/login-failures

### Additional information
We should consider that we already have a UI to search from all logs. That should be included as a report as well.

matiasbavera:

## What's new

fixes #252 and #287

This PR affects the react-component and **adds** a new package call it **reporting**. Also adds a new package called **rmf-auth** which contains the logic of keycloak and stubuser Auth. 

image

I created a UI for static reports. This includes a new `page` and a Multimenu sidebar to select the reports.

This would be useful for the person that it's viewing the logs.

reports

## Self-checks

## Discussion

I added a static report component and I'm showing that component from the stories. I think it'll be best to show those reports (or that screen) when at least one of them is working. As a follow-up to this issue, I'll continue with the doors report, creating an endpoint for the `doors state` on the backend and loading that result to the UI.

 
Is the name `reporting` ok?


matiasbavera:
Hi @koonpeng what do you think of having an interface for the static reports, like the one from the video?

koonpeng:
Thats certainly very useful for the log viewer.

koonpeng:
Can you merge main? Commands panel has been removed so the e2e tests should pass now.

koonpeng (commented):
image

Got an error at door logs.

image

Not sure if it's just storybook, but I'm getting an extra scroll bar for the "outer" div.

Assuming we are going ahead with out of app logging, I think this should be in a different package, it will be deployed separately and served on a different url. Reason is because retrieving logs inherently requires knowledge of the database and schemas, putting this UI in dashboard would lock us into the inhouse logging system.

@codebot What do you think? Does it make sense to have the log viewer in a different application?

matiasbavera:
> Can you merge main? Commands panel has been removed so the e2e tests should pass now.

I rebased onto main

matiasbavera:
> image
> 
> Got an error at door logs.

Yeah, my bad it's not supposed to exist in this PR yet =)

> Not sure if it's just storybook, but I'm getting an extra scroll bar for the "outer" div.

Fixed

> Assuming we are going ahead with out of app logging, I think this should be in a different package, it will be deployed separately and served on a different url. Reason is because retrieving logs inherently requires knowledge of the database and schemas, putting this UI in dashboard would lock us into the inhouse logging system.

It makes sense to me, you mean in the same `rmf-web` repo or as another repo?

matiasbavera:
Hey @koonpeng since this is a different package, and it should be served outside the dashboard we should add the auth logic with Keycloak as well, right?

koonpeng:
> Hey @koonpeng since this is a different package, and it should be served outside the dashboard we should add the auth logic with Keycloak as well, right?

Yeah, let's go with this design.

packages/react-components/lib/reports/multi-level-menu.tsx
```diff
@@ -0,0 +1,103 @@
...
+      </Collapse>
+    </div>
+  );
+};
+
+export interface MultilevelMenuProps {
+  menuStructure: (ExpandableMultilevelMenuProps | MenuItemProps)[];
+}
+
+export const MultiLevelMenu = (props: MultilevelMenuProps): React.ReactElement => {
```
koonpeng, line 74: This re-renders all items when any of them changes, should we memo `ExpandableMenuItem` and `MenuItem` to optimize it? Or it doesn't really matter since they are simple enough that they have minimal performance impact.

packages/react-components/lib/reports/render-selected-component-from-dict.tsx
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+
+interface RenderSelectedComponentProps {
+  selectedKey: string;
+  obj: Record<string, React.ReactElement | null>;
+}
+
+export const RenderSelectedComponentFromDict = (
```
koonpeng, line 8: afaik, it conditionally renders an item, or nothing if the item doesn't exist. Would it be simpler if we just use
```tsx
{ Object.prototype.hasOwnProperty.call(obj, selectedKey) && obj[key] ? obj[key] : null }
```

packages/react-components/lib/reports/render-selected-component-from-dict.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const component = React.useMemo(() => {
+    if (!Object.prototype.hasOwnProperty.call(obj, selectedKey)) {
+      return null;
+    }
+    if (!obj[selectedKey]) {
+      return null;
+    }
+
+    return obj[selectedKey];
+  }, [selectedKey, obj]);
```
koonpeng, line 22: This causes a re-render even if the selected component is unchanged

```tsx
const obj = React.useState({ unchanged: component });
React.useEffect(() => {
  setObj((prev) => { ...prev, changed: newComponent });
}, []);
return <RenderSelectedComponent selectedKey="unchanged" obj={obj} />
```

Even though only `changed` is updated, `RenderSelectedComponent` still re-renders because `obj` itself is changed.

packages/react-components/lib/reports/multi-level-menu.tsx
```diff
@@ -0,0 +1,103 @@
+import React, { useState } from 'react';
```
koonpeng, line 1: Can you add a story for this?

packages/reporting/package.json
```diff
@@ -0,0 +1,82 @@
+{
+  "name": "reporting",
+  "version": "1.0.0",
```
koonpeng, line 3: nit: start from `0.0.1`

packages/reporting/src/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+ReactDOM.render(
+  <React.StrictMode>
+    <App />
+  </React.StrictMode>,
+  document.getElementById('root'),
+);
+
+// NOTE: There is no point to have the app work offline as all the data must come from rmf/romi and
+// cannot be cached.
```
koonpeng, line 15: I guess this doesn't hold true anymore, even though the data comes from rmf, the assets can still be cached. We could experiment with offline first in the future but for now just remove this note.

packages/reporting/src/reportWebVitals.ts
```diff
@@ -0,0 +1,15 @@
+import { ReportHandler } from 'web-vitals';
+
+const reportWebVitals = (onPerfEntry?: ReportHandler) => {
+  if (onPerfEntry && onPerfEntry instanceof Function) {
+    import('web-vitals').then(({ getCLS, getFID, getFCP, getLCP, getTTFB }) => {
+      getCLS(onPerfEntry);
+      getFID(onPerfEntry);
+      getFCP(onPerfEntry);
+      getLCP(onPerfEntry);
+      getTTFB(onPerfEntry);
```
koonpeng, line 10: What do all these acronyms mean? Are all these metrics client side?

packages/reporting/src/stories/0-Welcome.stories.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
```
koonpeng, line 1: We can remove this.

packages/react-components/lib/reports/multi-level-menu.tsx
matiasbavera: Added!

matiasbavera: image

packages/reporting/package.json
matiasbavera: Thanks!

packages/reporting/src/stories/0-Welcome.stories.tsx
matiasbavera: Removed from the dashboard package as well.

packages/reporting/src/index.tsx
matiasbavera: Got it! Thanks

packages/react-components/lib/reports/render-selected-component-from-dict.tsx
matiasbavera: Yeah, you're right, I'll remove this component. I'll add it directly to the ReportDashboard component.

packages/react-components/lib/reports/render-selected-component-from-dict.tsx
matiasbavera: Yeah I removed this component and added that logic to ReportDashboard

packages/react-components/lib/reports/multi-level-menu.tsx
matiasbavera: Some people said that it doesn't matter at this stage of the app. But I can assure you that if in the future we have a troublemaker child component that triggers rerenders every simple re-render of the parents will count. So, I think we should add a memo as you recommend here.

matiasbavera: Without memo
image

With memo
image

packages/reporting/src/reportWebVitals.ts
matiasbavera: You can check that here https://web.dev/vitals/

matiasbavera: Mmm... but now that I'm thinking better it shouldn't affect us much. Since this would be used locally. I'll removed from now.

.github/workflows/reporting.yml
```diff
@@ -0,0 +1,39 @@
...
+        shell: bash
+        working-directory: packages/reporting
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12
+      - name: bootstrap
+        run: |
+          pip3 install pipenv
```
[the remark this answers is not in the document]
matiasbavera: Oh yeah you're right! thanks! I also removed from react-components too.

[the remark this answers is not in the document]
koonpeng: oh, didn't notice you are not using the image.

koonpeng (commented):
To avoid usage of `react-router-dom` in `react-components` and `rmf-auth`, you can move it to devdependencies, and make sure to only import it with `import type`. To get the concrete component, you can follow the current pattern, where you take in a component as a prop, or make use of HOC functions.

packages/reporting/package.json
```diff
@@ -0,0 +1,83 @@
...
+    "@testing-library/user-event": "^12.1.9",
+    "@types/debug": "^4.1.5",
+    "@types/node": "12.12.14",
+    "@types/react": "^16.9.49",
+    "@types/react-router": "^5.1.7",
+    "@types/react-router-dom": "^5.1.5",
+    "axios": "^0.21.1",
+    "clsx": "^1.1.1",
+    "debug": "^4.1.1",
+    "jest-canvas-mock": "^2.3.1",
```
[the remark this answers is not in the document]
matiasbavera: Yeah! I move it debug and react-router-dom

packages/rmf-auth/lib/utils/url.ts
```diff
@@ -0,0 +1,41 @@
+/**
```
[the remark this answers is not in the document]
matiasbavera: If this is the case `rmf-auth` will depend on `react-components`. And dashboard will depend on `react-components` and `rmf-auth`. Maybe an option, but it may be something overkill is to have our own common-utils package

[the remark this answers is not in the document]
koonpeng: Hmm yea, the current dumb dependencies system probably can't handle transitive dependencies. Maybe we should start looking at moving to npm@7 with the new workspaces feature, or move to yarn, or use one of the meta repo utils like lerna.

For now I guess duplicating the code is fine, I agree that a common-utils package might be abit overkill if this is the only code in there.

packages/rmf-auth/lib/utils/url.ts
```diff
@@ -0,0 +1,41 @@
...
+
+/**
+ * Get a qualified url based on the path. The schema and hostname is obtained based on the
+ * current `window.location`.
+ */
+export function getUrl(path: string): string {
+  return `${window.location.origin}${normalizePath(path)}`;
+}
+
+export const LOGIN_ROUTE = normalizePath(`${BASE_PATH}/login`);
```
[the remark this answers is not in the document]
matiasbavera: Still on rmf-auth but LOGIN_ROUTE is on every package

packages/dashboard/src/components/auth/tests/login.test.tsx
```diff
@@ -4,7 +4,7 @@ import userEvent from '@testing-library/user-event';
 import { BrowserRouter } from 'react-router-dom';
 import { AuthenticatorContext, UserContext } from '../../auth/contexts';
 import Login from '../login';
-import FakeAuthenticator from '../__mocks__/authenticator';
+import { FakeAuthenticator } from 'rmf-auth';
```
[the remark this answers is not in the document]
matiasbavera: Yeah, I'll implement that on the dashboard and in monitoring

matiasbavera:
I tested to remove the `react-route-dom` from all packages, clean up the node_modules, and only install in the root package and added `react-route-dom` as a peer dependency on all the other packages.  But still not working

image


Here are some solutions that I found:

https://stackoverflow.com/questions/58949438/how-do-i-use-reacts-link-component-in-an-external-npm-package-without-error-in/65773452#65773452

https://stackoverflow.com/questions/55552147/invariant-failed-you-should-not-use-route-outside-a-router

matiasbavera:
Ready for review again @koonpeng  Thanks!

koonpeng (approved):

PR #247, opened 2021-03-10
api-server: more health monitoring

koonpeng:
## What's new

This mainly adds health monitoring for door, lifts, dispensers and robots, also adds various endpoints related to them.

Currently all the components are monitored by their heartbeat, doors are additionally monitored by its mode, if the mode is `MODE_OFFLINE` or `MODE_UNKNOWN` or unknown (because we never received any states), it is considered UNHEALTHY, other modes are considered HEALTHY.

## Self-checks

## Discussion

How should we determine the health of an item that has a mode? For example door state has 

```
uint32 MODE_CLOSED=0
uint32 MODE_MOVING=1
uint32 MODE_OPEN=2
uint32 MODE_OFFLINE=3
uint32 MODE_UNKNOWN=4
```

* Should `MODE_OFFLINE` be considered `UNHEALTHY` or `DEAD`?
* Should we make a distinction between "offline" and "no heartbeart", I suppose "offline" means the door is still "working" but unable to accept requests, "no heartbeat" means we did not receive any states from the door adapter so it may have crashed and is completely uncontactable.
* How about for lifts, what should be the health status be for `MODE_FIRE` or `MODE_EMERGENCY`.


matiasbavera:
> How should we determine the health of an item that has a mode? For example door state has
> 
> ```
> uint32 MODE_CLOSED=0
> uint32 MODE_MOVING=1
> uint32 MODE_OPEN=2
> uint32 MODE_OFFLINE=3
> uint32 MODE_UNKNOWN=4
> ```
> 
> * Should `MODE_OFFLINE` be considered `UNHEALTHY` or `DEAD`?
> * Should we make a distinction between "offline" and "no heartbeart", I suppose "offline" means the door is still "working" but unable to accept requests, "no heartbeat" means we did not receive any states from the door adapter so it may have crashed and is completely uncontactable.
> * How about for lifts, what should be the health status be for `MODE_FIRE` or `MODE_EMERGENCY`.

Well, those are good questions, let's ask @codebot about it.

koonpeng:
> I would also suggest the same reasoning for lifts; if they are in `FIRE` / `EMERGENCY` mode, they are unavailable for RMF usage, so they should be considered `UNHEALTHY` because they cannot be used for navigation.

The argument on the other side is that fire/emergency still means the lift is "working normally", one could say that if emergency alarm is activated but the lift is not in emergency mode could be considered unhealthy, and vice versa.

Should we make the distinction of health status based on its ability to response to rmf, or based on if it is working in the correct mode given a certain situation? I'm inclined to go with the former just because its much easier, the latter may also raise some very complex scenario based on site requirements, or the condition may be different for each actor
 (e.g. imagine if we have a fire fighting robot).

matiasbavera:
> Should we make the distinction of health status based on its ability to response to rmf, or based on if it is working in the correct mode given a certain situation? I'm inclined to go with the former just because its much easier, the latter may also raise some very complex scenario based on site requirements, or the condition may be different for each actor
> (e.g. imagine if we have a fire fighting robot).

I vote to go with the first option too, because we wouldn't know all the possible scenarios for the second one.

packages/api-server/api_server/rmf_io/health_watchdog.py
```diff
@@ -24,41 +31,239 @@ def __init__(
...
+        sub = heartbeat_health.pipe(
+            op.combine_latest(door_mode_health),
+            most_critical(),
+        ).subscribe(self._report_health(self.rmf.door_health), scheduler=self.scheduler)
+        self.watchers.append(sub)
+
+    def _watch_lift_health(self, building_map):
+        lifts: Sequence[Lift] = building_map.lifts if building_map else []
+
+        def to_lift_health(data: Tuple[str, bool]):
```
matiasbavera, line 162: I think that if we follow what you mentioned

> Should we make the distinction of health status based on its ability to response to rmf

The only thing that it's missing (i think) is to handle the     `static readonly MODE_OFFLINE = 4;` state of the lift

matiasbavera (commented):
Hi KP, thanks for this, I left a couple of comments.

HTH,

matiasbavera (approved):
Thanks for the changes! code LGTM

PR #248, opened 2021-03-10
Feature/emergency panel

vallq:
## What's new
I created a barebones UI for the emergency panel and included a test alarm button to simulate code red and code blue emergencies. Right now it's just an interface with placeholder data. More definitely needs to be done. 
I'm not sure what would be important information a user would want to see, or how an emergency would be 'registered' by the dashboard. Opening this up for discussion and input since it is just structural for now.

final_6048a2a41b6bbd0084316459_576894

## Self-checks

## Discussion
- How an emergency will be published to the dashboard
- Critical information and actions on the panel


vallq:
Hi @codebot @matiasbavera , has there been any discussion or update regarding how the dashboard would receive an emergency 'state' or message? is this something within `rmf_core` or it will be some sort of external system?

koonpeng:
I don't think we should hide the emergency status in a panel, you might want to look at integrating it into the new wip main menu. The bell at the appbar is useful, but I think we should share it with other notifications. Instead of a sidebar, I think it's better to present a pop up on the center of the screen.

matiasbavera:
> Hi @codebot @matiasbavera , has there been any discussion or update regarding how the dashboard would receive an emergency 'state' or message? is this something within `rmf_core` or it will be some sort of external system?

Hi @vallq it's like @codebot said, also I tagged you in Asana and older PRs related to this topic, for you to get the full context of this. In those tasks/PR is the name of the topic.  

The thing about this, that you will find in one of the conversations, is that we "agreed" in the past on having a backend (in this case could be the api-server) that gets the message from the rmf emergency topic and publish through a web socket constantly to all of the clients (browsers) the state of the alarm. @koonpeng  please correct me if I'm wrong

matiasbavera:
> I think it's better to present a pop up on the center of the screen.

+1 to this idea

vallq:
Thanks all for the input/feedback! @codebot @matiasbavera @koonpeng I'll look through the older discussions and see what else I can come up with. And also the wip main menu!

I think I did not explain it clearly but the sidebar was meant to be a drawer just for 'testing' the alarm? To show that the dialog/pop up will appear when the emergency state is set to true / a specific emergency condition is received.

Initially, I thought to make the dialog have an 'Open Emergency Panel' button so that the user can be directed there instead of manually opening it to check the details. But as KP had answered my question regarding manipulating the dashboard actions out of the dashboard, it seems to not be a good idea.

Another idea I'm having is to have a separate page/widget specifically for emergency-related things, such as actions that the user can take or a 'log' of messages to include details and updates on the emergency. But again, it also depends on the use case and if it is relevant for the app to have such details.

I'll look into this more and ping on the SG web channel (:

matiasbavera:
> I think I did not explain it clearly but the sidebar was meant to be a drawer just for 'testing' the alarm? To show that the dialog/pop up will appear when the emergency state is set to true / a specific emergency condition is received.
Oh, I get now =) I did a similar thing to test the alarm a while ago.

> Another idea I'm having is to have a separate page/widget specifically for emergency-related things, such as actions that the user can take or a 'log' of messages to include details and updates on the emergency. But again, it also depends on the use case and if it is relevant for the app to have such details.

This sounds good, but as you said, it depends on the use case

PR #249, opened 2021-03-11
build: remove backend dependencies from dashboard

koonpeng:
## What's new

I noticed a preference to develop components directly in dashboard instead of react-components, so this PR removes the backend depdendencies on the dashboard so that you can run storybook without having RMF installed/sourced. Running the actual dashboard still requries RMF so it is now the dev's responsibility to bootstrap the backend packages (`ros2-bridge` for now) prior to running `npm start` on the dashboard.

If you have been using `npm run bootstrap`, it shouldn't change anything, it only affects if you run `npm run bootstrap -- packages/dashboard`, now the latter no longer automatically bootstraps the backend.

## Self-checks

## Discussion

One thing to note about this change is that in the future, `api-client` will generate it's model from `api-server`, which generate it's models from rmf, so this means that `api-client` will require rmf and will not automatically bootstrapped. More importantly, it also WILL NOT be in package.json so the build process must somehow find `api-client`, for the current react-scripts build, I don't think it allows relative imports outside `src`, so we will need to do some tricks like symlinking `api-client` into node_modules through some hooks. Devs must also manually bootstrap `api-client` before building `dashboard`, they must also make sure that all stories they write must work in the absent of `api-client`. We will need CI tests to check that all the above things are working.

The alternative is to keep it as it is now, we enforce the backend and `api-client` into package.json, downside is that you will need rmf to `npm install` `dashboard`, and so you need rmf to run storybook.


koonpeng:
after considering the downsides, i don't think this change is worth it, it's much easier to enforce requirement of rmf in order to do development on `dashboard`.

Please chip in if anyone thinks removing requirement on rmf is better.

PR #265, opened 2021-03-15
Feature/search and filter

ChawinTan:
## What's new

Added a search bar on top of the omni-panel for each item.

Added a new `SimpleSearch` component, a story for it as well as a unit test.

Dashboard:
search_and_filtering

Story
simple_search_story

## Self-checks


koonpeng:
I think we can streamline it by removing the "X" button. An empty search term means show all items. We can also probably remove the `disabled` prop, I don't think we need it if the length is 0, are there any use case you can think of that uses it?

One thing we need to account for is spotlighting an item when a user click on a marker in the map. If the current filter does not include the item, what do you think is the best approach?
  * Do nothing, item won't be spotlighted
  * Automatically reset the filter
  * Show a prompt asking the user if they want to reset the filter

For the UI side, is it possible to make the separator between the search bar and the item a little more prominent?

ChawinTan:
> I don't think we need it if the length is 0, are there any use case you can think of that uses it?

My original intention was to disable the search bar if there are no equipment in that category, for example, there are no lifts/dispensers in this map, at least for now.

> One thing we need to account for is spotlighting an item when a user click on a marker in the map. If the current filter does not include the item, what do you think is the best approach

Thanks for the reminder, I think option 2 would suffice for this simple search. I feel that allowing user to select "no" for option 3 might not be needed because if the user clicks on an item, they most probably already want to see some details for that item and would mostly select "yes" to reset the search bar. At least that's what I think.

ChawinTan:
- Removed the X button and the disabled feature
- added a divider between search bar and accordion
- reset the search bar every time a new panel is selected
- reset the search bar and spotlight the item after the item is clicked
- updated wip-main-menu-plus-plus with the changes
updated_search

packages/dashboard/src/components/dashboard/dashboard.tsx
```diff
@@ -308,55 +323,69 @@ export default function Dashboard(_props: {}): React.ReactElement {
...
-                ref={doorAccordionRefs[door.name].ref}
-                door={door}
-                doorState={doorStates[door.name]}
-                onDoorControlClick={handleOnDoorControlClick}
-                data-name={door.name}
-              />
-            ))}
+            <SimpleSearch onChange={onChange} value={search} />
+            {doors.map((door) => {
+              return door.name.includes(search) ? (
```
koonpeng, line 331: recommend for this to be case insensitive

packages/dashboard/src/components/dashboard/main-menu.tsx
```diff
@@ -16,37 +16,44 @@ const useStyles = makeStyles((theme) => ({
 
 export interface MainMenuProps {
   pushView(view: OmniPanelViewIndex): void;
+  setSearch?: () => void;
```
koonpeng, line 19: I think we should have some docs for this prop.

On another note, this brings up some design consideration I have:

Should `MainMenu` manage the search state? Would it be better to expose on click handlers for each button and let `Dashboard` handle it? This also goes for `pushView`. I feel doing so is more "react-ive". The counter argument is that since there is no plan to use `MainMenu` outside of `Dashboard`, the usual react best practices does not offer any advantage, and doing it this way reduces the load/bloat on `Dashboard` and avoid it becoming somewhat of a "god class".

reference: https://reactjs.org/docs/thinking-in-react.html#step-4-identify-where-your-state-should-live

@matiasbavera Which do you think is the better approach?

packages/react-components/lib/simple-search.tsx
```diff
@@ -0,0 +1,47 @@
...
+  },
+  searchBar: {
+    width: '100%',
+  },
+  divider: {
+    margin: '1.5rem 0',
+  },
+}));
+
+export const SimpleSearch = (props: SimpleSearchProps): JSX.Element => {
```
koonpeng, line 29: Do you think changing "Search" to "Filter" across the PR makes it more user friendly? Technically it is filtering instead of searching, having a text input with "Filter" might be better UX.

koonpeng (commented):
I think it's useful to add a e2e test to test the scenario where a filter is applied and user clicks on a marker.

ChawinTan:
I have left out the coverage of `dasbboard.tsx` for now. I have created a separate task for the mocking of schedule visualizer as that seem to be quite a big task on its own. I think we can also test for `setFilter` on the dashboard now since that part of the code is not affected, but the increment in coverage would be minimal and I feel that it is not worth it and would be better addressed when we successfully mock the `scheduleVisualizer`. For now, I have added some tests to `main-menu` to test for `pushViews` and `setFilters`.

- Added an end to end test
- updated Search to Filter
- updated wip-main-menu to keep in sync
- set filter term and equipment to lower caps to make it case-insensitive
- added test to main menu and wip main menu and updated it to use RTL.
- add some docs to `setFilter`

packages/dashboard/e2e/tests/filtering.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    // set value to filter input
+    $('[data-component=simple-filter]').$('input').setValue('value');
+
+    // get door marker and click
+    const door = $('[data-component=DoorMarker]');
+    door.waitForClickable();
+    door.click();
+
+    // check that the filter is empty after clicking
+    expect($('[data-component=simple-filter]').$('input').getValue()).toEqual('');
```
koonpeng, line 29: we should check if the door accordion for the clicked door is expanded as well.

koonpeng (approved):

PR #267, opened 2021-03-22
api-server: task summary

koonpeng:
## What's new

Add task summaries to api-server endpoints; fix api-client typings.

## Self-checks


packages/api-client/lib/index.ts
```diff
@@ -12,6 +12,7 @@ export type Topic =
   | 'ingestor_health'
   | 'fleet_states'
   | 'robot_health'
+  | 'task_summaries'
```
matiasbavera, line 15: Do you think that we can use an Emun here? I think it'd be ideal, but knowing about the type checking downside I'm not so sure. =|

matiasbavera: Mmm.. but it'll be easier for the user that imports that type and see the options as it is now.

packages/api-server/api_server/rmf_io/test_book_keeper.py
```diff
@@ -217,3 +218,67 @@ async def test_write_robot_health(self):
         health = await models.RobotHealth.get(id_="test_robot")
         self.assertIsNotNone(health)
         self.assertEqual(health.health_status, models.HealthStatus.UNHEALTHY)
+
+    async def test_write_task_summary(self):
```
matiasbavera, line 222: Do you know if we now have support to pause/resume tasks as well?

matiasbavera (approved):
Left a couple of questions but LGTM

packages/api-client/lib/index.ts
koonpeng: We need string literal types because we need to be compatible with the types of event emitter, not sure if enum can work.

packages/api-server/api_server/rmf_io/test_book_keeper.py
koonpeng: I think it is wip.

Issue #284, opened 2021-03-22
backend: auth


codebot:
needs to work on air-gapped LAN's (so, 3rd-party web-based auth is not possible). Continue iterating on our Keycloak-based system as needed, adding auth to backend API calls, etc.

codebot:
https://github.com/open-rmf/rmf-web/pull/288

PR #288, opened 2021-03-23
api-server: auth

koonpeng:
## What's new

Adds simple jwt auth, addresses #284 

## Self-checks


packages/api-server/api_server/rmf_io/test_rmf_io.py
```diff
@@ -255,3 +259,72 @@ def on_building_map(building_map: dict):
...
+        await self.runner.setup()
+        self.site = aiohttp.web.TCPSite(self.runner, "127.0.0.1", 0)
+        await self.site.start()
+        self.server_port = self.runner.addresses[0][1]
+
+    async def asyncTearDown(self):
+        await self.client.disconnect()
+        await self.runner.cleanup()
+
+    async def test_fail_with_no_auth(self):
```
matiasbavera, line 294: Maybe I'm not getting it right, but aren't those tests always going to pass?

matiasbavera (commented):
Hi, good idea about the stub user, same as the front end. Left a comment

packages/api-server/api_server/rmf_io/test_rmf_io.py
koonpeng: turns out the tests are broken because the python client doesn't support auth, now using nodejs client via subprocess

matiasbavera (approved):
Thanks for this. LGTM, left a minor comment.

packages/api-server/api_server/rmf_io/test_data/connect.js
```diff
@@ -0,0 +1,22 @@
...
+(async () => {
+  const client = io(uri, { auth: { token: token } });
+  const success = await new Promise((res) => {
+    client.on('connect_error', () => res(false));
+    client.on('connect', () => res(true));
+  });
+  client.disconnect();
+
+  if (!success) {
+    console.log('fail');
```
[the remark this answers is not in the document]
koonpeng: I don't think it matters, this is just a minimal script to report if it can connect to the socketio server.

PR #289, opened 2021-03-24
Fix/make api client avail

ChawinTan:
## Self-checks

## Discussion

Made a symlink and a script to build `api-client` so as to allow the front end to subscribe to states from the dashboard.

@codebot  and @koonpeng Do you think these changes are needed for the time being or are we sticking to `ros2-bridge` for April? If not then I will leave it as a draft request for now


koonpeng:
tests are failing because of lint errors, you can ignore the rule for the generated files or change the template in https://github.com/open-rmf/ts_ros

ChawinTan:
I'm getting a bunch of errors locally because of missing `.srv` files. Are they needed?
srv_errors

packages/api-server/api_server/app.py
```diff
@@ -166,7 +166,8 @@ async def on_shutdown():
     logger.info("shutdown app")
 
 
-sio = socketio.AsyncServer(async_mode="asgi")
+# TODO - change cors_allowed_origin to more specific origin. Putting a wild card for now
```
koonpeng, line 169: I think this would depend on the use case of api-server. In order for it to be consumed by any web page, we need cors *. If we only want our own pages to consume the apis, then we need a config to specify the allowed origins. In either case, apps can always consume the apis, this only affects pages in browser.

It is hard to say how * cors would impact security, we are not using cookies so that limits the scope of xsrf attacks, but of course limiting the origin would always be more secure, at the expense of blocking off third party pages (if we have any intention for third party pages to make use of our apis).

koonpeng (approved):

packages/api-server/api_server/app.py
ChawinTan: I see! I am still making the assumption that everything is air-gapped and there won't be any other use case anytime soon

cnboonhan (approved):
LTGM! :+1:  thanks for pushing this through so swiftly!

packages/api-server/api_server/app.py
koonpeng: Note that being airgap doesn't actually affect this attack vector. It reduces the chance of a malicious actor taking advantage of such vulnerability, but someone that has access to the network, but doesn't have access to certain admin functions may make use of security vulnerabililties like these to do things they should not be doing.

I don't think a * here opens any vulnerabilities, but I could be wrong, in any case, there should be a security audit to sniff out vulnerabilities.

PR #291, opened 2021-03-26
Feature/task management

ChawinTan:
## What's new

- Removed socketIo connection to `rmf_api_server` since there doesn't seem to be a need for it
- Rest endpoints for submit_task, get_tasks, and cancel_task

## Self-checks


koonpeng:
this seems to contain the commits for #289 as well, now that it has been merge, i would recommend to rebase it before things get out of hand.

packages/api-server/rest_server/tasks/dispatcher.py
```diff
@@ -0,0 +1,216 @@
...
+
+    def get_task_status(self):
+        """
+        Get all task status - This fn will trigger a ros srv call to acquire
+        all submitted tasks to dispatcher node. Fn returns an object of tasks
+        """
+        req = GetTaskList.Request()
+        try:
+            future = self.get_tasks_srv.call_async(req)
+            rclpy.spin_until_future_complete(self, future, timeout_sec=0.5)
```
koonpeng, line 105: Some things to test:

Because we will need to publish stuffs, test if `publish` can work if rclpy is not spinning, or if there is a way to publish with `spin_until_future_complete`. If that didn't work out, then we can't use this pattern. We will need to start a new thread and do `rclpy.spin` there, then we need some way to synchronize the event loop. We could use `call()` here, but that is a blocking call and would destroy performance.

In any case, `spin_until_future_complete` is blocking and would destroy perf so we should probably move away from this pattern.

packages/api-server/rest_server/tasks/dispatcher.py
```diff
@@ -0,0 +1,216 @@
...
+            response = future.result()
+            if response is None:
+                return self.active_tasks_cache + self.terminated_tasks_cache
```
koonpeng, line 108: I don't think we should cache results, if the call fails then just return an error. Caching should be done by a reverse proxy.

packages/api-server/rest_server/tasks/dispatcher.py
```diff
@@ -0,0 +1,216 @@
...
+            else:
+                active_tasks = self.__convert_task_status_msg(
+                    response.active_tasks, False
+                )
+                terminated_tasks = self.__convert_task_status_msg(
+                    response.terminated_tasks, True
+                )
+                self.active_tasks_cache = active_tasks
+                self.terminated_tasks_cache = terminated_tasks
+                return active_tasks + terminated_tasks
```
koonpeng, line 118: we should probably follow the original structure, aka `{ "active_tasks": [tasks], "terminated_tasks": [tasks] }` instead of flattening them.

packages/api-server/rest_server/tasks/dispatcher.py
ChawinTan: So there would be 2 threads, one to spin for the result of the service (task details) and the other to spin the publisher? The publisher is going to publish other things but if a task is returned, then we would need to synchronize the 2 threads to publish a `task event` or something like that?

Pardon me for my ignorance, but what would we be publishing? I thought the `rest server` is just to handle a one time request and the change in robot states and trajectories would be handled by the `api-server`, where we can subscribe to it from the dashboard?

koonpeng: There would be 1 spinning thread that takes care of all rclpy operations. Door and lift requests are an example where we need to publish stuffs.

I'm working on the infrastructure apis, along with some refactoring, so I guess you can just ignore this, let's merge this first, then I will refactor this in my PR.

ChawinTan:
I'm going to take a look as to why the CI is failing all of a sudden in the mean time.

koonpeng:
see #293 for the e2e failures

packages/api-server/rest_server/tasks/dispatcher.py
```diff
@@ -0,0 +1,214 @@
...
+                    + " --> "
+                    + desc.delivery.dropoff_place_name
+                )
+            elif status["task_type"] == "Charging":
+                status["description"] = "Back to Charging Station"
+
+            # Generate a progress percentage
+            duration = abs(task.end_time.sec - task.start_time.sec)
+            # check if is completed
+            if is_done or task.state == 3:
```
koonpeng, line 200: dont use magic literals, use the constants defined in rmf_task_msgs

packages/api-server/rest_server/tasks/dispatcher.py
```diff
@@ -0,0 +1,214 @@
...
+            4: "Clean",
+            5: "Patrol",
+        }
+
+        status_list = []
+        rclpy.spin_once(self, timeout_sec=0.0)
+        now = self.get_clock().now().to_msg().sec  # only use sec
+        for task in task_summaries:
+            desc = task.task_profile.description
+            status = {}
```
koonpeng, line 166: create a model for this?

packages/api-server/rest_server/tasks/dispatcher.py
```diff
@@ -0,0 +1,214 @@
...
+                    + str(desc.loop.num_loops)
+                )
+            elif status["task_type"] == "Delivery":
+                status["description"] = (
+                    desc.delivery.pickup_place_name
+                    + " --> "
+                    + desc.delivery.dropoff_place_name
+                )
+            elif status["task_type"] == "Charging":
+                status["description"] = "Back to Charging Station"
```
koonpeng, line 195: Add a final else case for unknown task type?

packages/api-server/rest_server/tasks/dispatcher.py
```diff
@@ -0,0 +1,214 @@
...
+            5: "Pending",
+        }
+        type_enum = {
+            0: "Station",
+            1: "Loop",
+            2: "Delivery",
+            3: "Charging",
+            4: "Clean",
+            5: "Patrol",
+        }
```
koonpeng, line 159: don't use magic literals, use constants in rmf_task_msgs instead

packages/api-server/rest_server/tasks/dispatcher.py
```diff
@@ -0,0 +1,214 @@
...
+
+        status_list = []
+        rclpy.spin_once(self, timeout_sec=0.0)
+        now = self.get_clock().now().to_msg().sec  # only use sec
+        for task in task_summaries:
+            desc = task.task_profile.description
+            status = {}
+            status["task_id"] = task.task_id
+            status["state"] = states_enum[task.state]
+            status["done"] = is_done
```
koonpeng, line 169: doesn't the task state already contains information if it's done?

packages/api-server/rest_server/tasks/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+
+from .dispatcher import DispatcherClient
+
+router = APIRouter(prefix="/tasks", tags=["tasks"])
+
+rclpy.init(args=None)
+dispatcher_client = DispatcherClient()
+
+
+class Description(BaseModel):
```
koonpeng, line 17: `Description` is too vague to describe what it contains.

packages/api-server/rest_server/tasks/test_dispatcher.py
```diff
@@ -0,0 +1,135 @@
...
+                "dropoff_ingestor": "coke_ingestor",
+                "dropoff_place_name": "supplies",
+            },
+        }
+        req_msg_7, err_msg_7 = dispatcher_client.convert_task_request(mock_task_json_7)
+        self.assertEqual(err_msg_7, "")
+        self.assertIsNotNone(req_msg_7)
+
+        mock_task_json_8 = {
+            "task_type": "wrong_task",
```
koonpeng, line 80: @Yadunund Who should be verifying the input? Should dispatcher node response with a error message if it can't process the task?

Yadunund: The result of the service call to the dispatcher will be [unsuccessful](https://github.com/open-rmf/rmf_internal_msgs/blob/b6a583250948dcf6fd32fe32a1d8ba6a3d7a3787/rmf_task_msgs/srv/SubmitTask.srv#L18) if a task type that is not enumerated in [TaskType.msg](https://github.com/open-rmf/rmf_internal_msgs/blob/b6a583250948dcf6fd32fe32a1d8ba6a3d7a3787/rmf_task_msgs/msg/TaskType.msg#L2-L7) is received. But the `unsuccessful` result will also be received in situations where the request fields are incomplete. So it may be worthwhile for the UI to verify the input as well so that the user can be informed the exact reason the task cannot be processed (incomplete field, invalid task type, etc)

koonpeng: Can the dispatcher node give the error message as well? It's going to be very hard for the api server to verify more complex inputs. e.g. Checking start and end actually exists, checking certain field is required for certain task types. This would eventually lead to problems like the UI not allowing to submit tasks that should be possible (because new updates in dispatcher node has new rules), or the UI allowing tasks the should not be possible to be submitted, and the dispatcher node assumes that the input are verified correctly when it is not, leading to undefined behaviors.

koonpeng: There is also the UX problem that, for rules the UI doesn't/cannot check, and the dispatcher node returns with an error, all it can do is tell the user an unknown error occured. Since the dispatcher node already does the checking, it would really help if it can return an error message so the user will know what is wrong.

Yadunund: That is a fair point. @tanyouliang95 , do you think we can add a `string` field to the response of `SubmitTask.srv` that contains an error message when the request is unsuccessful?

tanyouliang95: Sure, we can add an additional field for this. Similarly, this will also apply to `CancelTask.srv`. Tbh, currently the possible error during submit task srv can only be related to incorrect or incomplete field in `TaskDescription`. More variation of Error will occur after submission.

On top of the verbose string, one future proposal is to add a new msg in  `rmf_task_msgs`, which could be named as `TaskErrorCode.msg`. This will consist of an enumeration of possible task-related error in RMF, which can also part of `TaskStatus.msg` for verbose informing of error.

packages/api-server/rest_server/tasks/dispatcher.py
ChawinTan: I guess we can remove it if there is no need for the dashboard to explicitly display if the task is `done` or `not done` base on the definition above.

ChawinTan:
- Remove the magic numbers from enums and statements
- created a model for status object
- remove is_done from response
- change name of Description to TaskTypeDescription

PR #298, opened 2021-04-01
refactor(api-server): unified server, ros perf improvements

koonpeng:
## What's new

* Merges the socketio server with the rest server, there is now a single unified server that serves both endpoints. The advantage of this is that it is now easier to share code between the 2 endpoints.
  * The previous problem with fastapi and socketio has been fixed, making this now possible.
  * Regarding scaling, a side effect is that the rest portion is no longer scalable. imo it shouldn't be too much of an issue, for apis that are essentially frontends to rmf, rmf will be the bottleneck anyway. For apis that is bottlenecked by computation, we could always add another layer of proxy / compute cluster, and they should be written in a more performant language.
* The current task apis has performance problems because it uses blocking calls in an async context. They are now refactored to spin ros node in a background thread, and using futures to synchronise work.
* Refactor task data validation to use pydantic, this makes validation simpler and catches more edge cases. Instead of a `TaskTypeDescription` that contain fields for all task types, using pydantic allows us to separate them out to 3 models for clean, loop and delivery tasks and easily validate it such that if `task_type == "clean"`, the task description should contain all the fields for the clean task etc.
* All tasks are now recorded in the db, previously it deletes tasks that are completed or failed. `RmfGateway.current_tasks` still only contains tasks that are active/pending/queued.
* Added authentication to the rest api.

## Self-checks


packages/api-server/package.json
```diff
@@ -4,12 +4,11 @@
   "description": "dummy package",
   "private": true,
   "scripts": {
-    "prepare": "cmake -B build && cmake --build build",
-    "test": "python3 -m pipenv run python -m unittest",
-    "test:cov": "python3 -m pipenv run python -m coverage run -m unittest && TEST_COVERAGE=1 npm run test:integration",
-    "test:integration": "python3 -m pipenv run python -m unittest discover -s tests",
+    "prepare": "",
```
matiasbavera, line 7: Should we remove this empty prepare?

matiasbavera (approved):
Code LGTM left a couple of comments


> Regarding scaling, a side effect is that the rest portion is no longer scalable. imo it shouldn't be too much of an issue, for apis that are essentially frontends to rmf, rmf will be the bottleneck anyway. For apis that is bottlenecked by computation, we could always add another layer of proxy / compute cluster, and they should be written in a more performant language

Makes sense. rmf will be the bottleneck. Thanks for this explanation.

packages/api-server/package.json
koonpeng: i think we can leave it here, npm looks at the prepare script to decide if it should install devdependencies

packages/api-server/CMakeLists.txt
```diff
@@ -1,48 +0,0 @@
...
-append_rosidl_interfaces(rmf_interfaces rmf_charger_msgs)
-append_rosidl_interfaces(rmf_interfaces rmf_dispenser_msgs)
-append_rosidl_interfaces(rmf_interfaces rmf_door_msgs)
-append_rosidl_interfaces(rmf_interfaces rmf_fleet_msgs)
-append_rosidl_interfaces(rmf_interfaces rmf_ingestor_msgs)
-append_rosidl_interfaces(rmf_interfaces rmf_lift_msgs)
-append_rosidl_interfaces(rmf_interfaces rmf_task_msgs)
-append_rosidl_interfaces(rmf_interfaces rmf_traffic_msgs)
-append_rosidl_interfaces(rmf_interfaces rmf_workcell_msgs)
-append_rosidl_interfaces(rmf_interfaces rmf_building_map_msgs)
```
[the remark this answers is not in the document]
koonpeng: This is to generate the mixins, it isn't actually used so I removed it.

Issue #274, opened 2021-03-22
Front end: implement the navigation bar


codebot:
Implement the navigation bar wireframe. This will also need to be done in a way that can be configured for each deployment, since the number and names of the panel collections will likely differ across sites.

codebot:
Can prototype with "stock" material-ui tab component: https://material-ui.com/components/tabs/

PR #302, opened 2021-04-13
React-Components / Dashboard: Title Bar and Navigation Bar

### LINKED ISSUE #273: Front end: implement the title bar
Implement the title bar wireframe in a way that can be used across many deployment sites. Also includes the user drop-down menu.

vallq:
## What's new

fixes #273 #274

Changes:
https://user-images.githubusercontent.com/28668944/114524897-32f72b80-9c78-11eb-92f6-d51ea1bda9d1.mp4

I've added Title Bar and Navigation Bar components
The title bar takes in a path to an image file to render the logo on the top left and I've kept the previous toolbar from the Appbar component.
The navigation bar has tabs rendered as elaborated in Discussion and the routes will lead to new panels (to be implemented).

## Self-checks

## Discussion
For now the tabs are 'hardcoded' as such in the `utils/url.ts` file
`export const TAB_NAMES = ['Overview', 'Building', 'Robots', 'Tasks', 'Logs'];`
in subsequent iterations, these should come in some sort of config file that also determines the layout?


koonpeng:
e2e test is failing, it's trying to run ros2 bridge, which should not be used anymore. You probably need to update the workflow yamls.

The omnipanel in the overview tab is cutting into the navigation bar.

Instead of putting placeholders for the various, how about just have a single overview tab for now, and add in the other tabs as we have a ui for them.

vallq:
@koonpeng thanks for your feedback! which files in particular should I be looking at for the workflow update?

> Instead of putting placeholders for the various, how about just have a single overview tab for now, and add in the other tabs as we have a ui for them.
Okay, got it

> Would it be better if we combine the appbar and navbar? There is quite abit of free space in the appbar, and imo having 2 layers of headers looks abit bad.
yea I think if we can be sure there will not be too many tabs, this would work too?
I'll do this and post a side by side comparison

matiasbavera:
For a follow-up issue I think it'll be great if we can switch tabs with hotkeys

matiasbavera:
> @koonpeng thanks for your feedback! which files in particular should I be looking at for the workflow update?

I think he is talking about e2e.yaml which uses the e2e package inside the Dashboard (you may want to look into that too).

matiasbavera:
> Instead of putting placeholders for the various, how about just have a single overview tab for now, and add in the other tabs as we have a ui for them.

+1 to this comment. Looks good Val

vallq:
Update on single vs double nav headers
photo6323339118947445450

koonpeng (commented):
I guess we can ignore the e2e tests for now, personnally i prefer the combine appbar, a separator between the tabs should make it look better, and it seems like the tabs are slightly aligned to the top, centering it and maybe increasing the font size should make it look better.

koonpeng:
actually, i think the whole appbar is not aligned vertically, "powered by openRMF" and all the icons are slightly on top as well. Also I think there should be some effects for the currently opened tab, the material ui example has a underline which is not here.

Do we actually have a building panel now? What about just putting the current dashboard view as the "Building" tab for now (the wireframe doesn't have a "Overview" tab if we are following that).

vallq:
Ok let me add in the divider and also check the alignment! Will provide a better screenshot

vallq:
divider
Hi all, the default `Divider` component shows up like this if we put it at the start and at the end of the tabs (the dark grey lines!), and the light variant is also a shade of grey lol
The white lines are implemented with CSS on the tabs component by setting the borderRight.

v3_navigation
I think going ahead with the current white lines should work for now, right? Anyway, this can be easily removed should we decide to go without dividers.

koonpeng:
The underline should align with the bottom of the app bar, it also needs some margins with the logo. I would say the divider is up to your preference, either on the ends, between the tabs or other arrangements.

koonpeng:
rmf-models branch has been merged, can you rebase it to main?

PR #305, opened 2021-04-16
react-components: task panel

koonpeng:
## What's new

UI for the new task panel based on the wireframe.

https://user-images.githubusercontent.com/45189696/114971870-d716e780-9eaf-11eb-84a9-f6fa8409d5cf.mp4

## Self-checks


packages/react-components/lib/task-summary/create-task.tsx
```diff
@@ -0,0 +1,460 @@
...
+        fullWidth
+        margin="normal"
+        value={value.startName}
+        onChange={(ev) =>
+          onChange({
+            ...value,
+            startName: ev.target.value,
+          })
+        }
+      />
```
matiasbavera, line 178: Maybe I'm missing something, but shouldn't we provide available places for the robot to go?

Same with the delivery request and the dispenser-ingestor

matiasbavera (commented):
Great! Thanks for this! Code LGTM and I also did some testing and it's working correctly, but I left some comments. HTH

scripts/symlink-packages.js
```diff
@@ -0,0 +1,16 @@
...
+  const packageJson = JSON.parse(fs.readFileSync(`${pkg}/package.json`));
+  const pkgName = packageJson.name;
+  try {
+    fs.unlinkSync(`node_modules/${pkgName}`);
+  } catch (e) {
+    if (e.code !== 'ENOENT') {
+      throw e;
+    }
+  }
+  fs.symlinkSync(`../${pkg}`, `node_modules/${pkgName}`);
```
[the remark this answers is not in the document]
koonpeng: It might not work on other OS, all these are really just temp workarounds until we can migrate to monorepo tools.

packages/react-components/lib/task-summary/create-task.tsx
koonpeng: I left that feature as todo for this initial implementation. Particularly because there is no formal way to get other infos like cleaning zones, (and future task types like patrol zones, charging stations etc).

matiasbavera (approved):

PR #310, opened 2021-04-23
react-components: header bar (#273 #274)

### LINKED ISSUE #273: Front end: implement the title bar

vallq:
## What's new

fixes #273 #274

header-bar

My previous commits in the other PR were not verified and signed off, hence I cherry-picked the commits and pushed them to this new branch instead.

This addresses previous comments about the alignment of the content in the bar.
I will proceed to remove the previous branch.

## Self-checks


packages/react-components/lib/navigation-bar.tsx
```diff
@@ -0,0 +1,45 @@
+import React from 'react';
```
koonpeng, line 1: Is this still being used?

packages/react-components/lib/title-bar.tsx
```diff
@@ -0,0 +1,78 @@
...
+        scrollButtons="auto"
+        value={value}
+        onChange={handleChange}
+        aria-label="navigation-tabs"
+      >
+        <Tab
+          label={
+            <Badge badgeContent={<img src={logoPath} alt="logo" className={classes.logo} />} />
+          }
+        />
```
koonpeng, line 65: This makes the logo clickable, which is pretty nice. Do you have a plan on what clicking it would do?

packages/react-components/lib/title-bar.tsx
```diff
@@ -0,0 +1,78 @@
...
+import { createStyles, makeStyles, AppBar, Theme, Toolbar, Typography } from '@material-ui/core';
+import Tabs from '@material-ui/core/Tabs';
+import Tab from '@material-ui/core/Tab';
+import Badge from '@material-ui/core/Badge';
+
+export interface TitleBarProps {
+  logoPath: string;
+  tabNames: string[];
+  children?: React.ReactNode;
+}
```
koonpeng, line 11: Maybe I'm missing something, how can the consuming component handle tab change events?

packages/react-components/stories/title-bar.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+import TitleBar from '../lib/title-bar';
+import AccountCircle from '@material-ui/icons/AccountCircle';
+import IconButton from '@material-ui/core/IconButton';
+
+export default {
+  title: 'Title Bar',
+  component: TitleBar,
+} as Meta;
+
+export const ExampleTitleBar: Story = (args) => {
```
koonpeng, line 12: Would be good to make it support storybook controls and actions, see the tasks stories for some examples. Mainly, you need to let `{...args}` take care of the props, and not pass fixed props like `tabNames={tabNames}` explicitly. Some things like `logoPath` can be fixed, but try to make some of the props configurable from storybook.

koonpeng (changes requested):

packages/react-components/lib/title-bar.tsx
vallq: Previously I thought having the action to be clicking on the logo would bring it back to the 'homepage', which is the first building tab for now, but I'm not sure if it's necessary or it will be confusing. Although this is conventional in most web pages!

packages/react-components/lib/title-bar.tsx
vallq: My understanding is that the current tab change handler si currently 'overridden' by the dashboard URLs, so if we choose not to use URLs, then clicking on another tab will lead to an empty page with only just the header bar, and if we include a `<TabPanel/>`, the contents there will be displayed. like in the `material-ui` [demo](https://material-ui.com/components/tabs/#tabs)  Right now it just keeps the props in check, by changing the 'value' according to the individual tab values.

packages/react-components/lib/navigation-bar.tsx
vallq: nope, but I was thinking to just leave it if it would be useful for building other dashboards? can remove if you think it's cluttering the storybook!

vallq:
@koonpeng thanks for the comments / feedback!!

packages/react-components/lib/title-bar.tsx
koonpeng: The `children` are inside the `Toolbar`, so we can't add `TabPanel` this way, it would end up in part of the `Toolbar`. Also this creates some confusing API where the consumer must somehow know that the `TabPanel` values is the index of the `TabNames` props and that the first tab name must be the homepage link. Also, how can the consumer overwrite the tab behavior?

The most flexible would be a "NavigationBar" and "Toolbar" and let the consumer control it freely, an alternative is to take in a tab names + values props and a "onTabChange" props to let the consumer handle tab change behavior.

packages/react-components/lib/header-bar.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const [value, setValue] = React.useState(0);
+
+  const handleChange = (event: React.ChangeEvent<unknown>, newValue: number) => {
+    setValue(newValue);
+  };
+
+  const populateTabs = () => {
+    const tabs = tabNames.map((tabName, index) => (
+      <Tab key={index} label={tabName} value={index} aria-label={`tab-${index}`} />
```
koonpeng, line 51: The aria label should be a human readable representation of the "subject" of the element, so something like "tab-0" doesn't make any sense, you should put the tab name instead. On another note, does the "label" prop not already provide the aria label?

packages/react-components/lib/header-bar.tsx
```diff
@@ -0,0 +1,106 @@
...
+    ));
+    return tabs;
+  };
+
+  const populateTabPanels = (data: TabPanelProps[]): JSX.Element[] => {
+    const tabPanels = data.map((item, idx) => {
+      return (
+        <TabPanel key={idx} value={value} index={idx}>
+          {item.children}
```
koonpeng, line 60: I don't think this works, at least not in the way react code is normally written. It takes in `TabPanelProps[]`, but only uses the `children`, the other props like `value` and `index` are ignored, it may cause a lot of confusion for the consumer.

Instead of doing it this way, you can just let the consumer create it's own `TabPanel`. e.g.

```tsx
const [tabValue, setTabValue] = React.useState(...);
return (
  <>
    <HeaderBar onTabChange={setTabValue} tabNamesAndValues={...}>...</HeaderBar>
    <TabPanel value={tabValue} index={...}>...</TabPanel>
    <TabPanel value={tabValue} index={...}>...</TabPanel>
  </>
)
```

The downside of the above example is that the consumer need to "magically" know that it should create it's own `TabPanel`.

There are some ways to improve upon it, e.g.

```tsx
const [tabValue, setTabValue] = React.useState(...);
return (
  <>
    <HeaderBar>
      <NavigationBar onChange={...}>
        <Tab .../>
        <Tab .../>
        ...
      </NavigationBar>
      <Toolbar>
        ...
      </Toolbar>
    </HeaderBar>
    <TabPanel value={tabValue} index={...}>...</TabPanel>
    <TabPanel value={tabValue} index={...}>...</TabPanel>
  </>
)
```

This would be pretty close to how the material api is designed. With this, the consumer controls both the `Tab` and the `TabPanel`, so it doesn't need any "magical" knowledge. But of course, the downside is that this is pushing more work onto the consumer, so the `NavigationBar` and `Toolbar` (and `HeaderBar` as well I guess) are going to be mostly just css styles "containers". e.g. In this case, the `NavigationBar` would be acting like the `Tabs` component (with our own styling) in material ui. One point to note is that since the homepage/banner is also a tab and that it probably has a different styling, you should create a `BannerTab` component as well.

packages/react-components/lib/tab-panel.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+import Box from '@material-ui/core/Box';
+
+export interface TabPanelProps {
+  children?: React.ReactNode;
+  index: number;
+  value: number;
+}
+
+export function TabPanel(props: TabPanelProps): JSX.Element {
```
koonpeng, line 10: Can't we just use material's `TabPanel` it doesn't seem to be doing anything much different.

packages/ros2-bridge/src/authenticator.d.ts
```diff
@@ -0,0 +1,6 @@
+import { WebSocketMiddleware } from './websocket-connect';
```
koonpeng, line 1: this should be removed, it's probably due to stale commit, can you rebase main? `git rebase main next ui/header-bar`, "next" is the commit that you initially based this PR on.

koonpeng (changes requested):

packages/react-components/lib/header-bar.tsx
vallq: I went to check just to be sure, the `label` prop only refers to the output rendered by the tab! there are no aria labels [implemented](https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Tab/Tab.js) in the button base of the component.

packages/react-components/lib/tab-panel.tsx
vallq: yea, it's the same implementation I believe. Tab Panel is still 'under development' because it's in the lab dir (not core) of the material-ui library

packages/react-components/lib/header-bar.tsx
koonpeng: Yeah, it actually just renders the `label` prop "as is".

packages/react-components/lib/tab-panel.tsx
koonpeng: :+1: I see, looks like the material tutorial creates it's own `TabPanel` as well.

packages/ros2-bridge/src/authenticator.d.ts
vallq: I wasnt able to run this command; it didnt give any changes or output.. but I managed to cherry-pick the changes, so I think it should be resolved..

vallq:
Getting this error in the CI, is this something that is being fixed? because I have pulled in changes from `main`!
```
> reporting-server@0.0.0 lint /__w/rmf-web/rmf-web/packages/reporting-server
> python3 -m pipenv run pylint rest_server

************* Module rest_server.repositories.test_log_creation_handler
rest_server/repositories/test_log_creation_handler.py:6:0: C0411: third party import "from tortoise import Tortoise" should be placed before "from models import DispenserState, DoorState, RawLog" (wrong-import-order)
```

Also the force push is because I had duplicate commits when trying to resolve the DCO failure..
Updated the files to reflect the changes that you mentioned above! Let me know if there are other changes I should make.

Thanks @koonpeng. Sorry about the slow progress as well.

packages/dashboard/src/components/appbar.tsx
```diff
@@ -42,83 +60,82 @@ export const AppBar = React.memo(
...
-              onClick={() => setShowSettings(true)}
+      <div>
+        <HeaderBar>
+          <BannerTab logoPath={'/roshealth-logo-white.png'} />
+          <NavigationBar onTabChange={onTabChange} value={value}>
+            <Tab
+              key={'building-tab'}
+              label={'Building'}
+              value={'building'}
+              aria-label={`building-tab`}
```
koonpeng, line 71: nit: you can pass string directly in jsx like this

```jsx
<Tab key="building-tab">
```

packages/react-components/lib/banner-tab.tsx
```diff
@@ -0,0 +1,43 @@
...
+const useStyles = makeStyles(() =>
+  createStyles({
+    logo: {
+      maxWidth: 120,
+      opacity: 1,
+    },
+  }),
+);
+
+interface BannerTabProps {
```
koonpeng, line 16: good to export the interface this as it might be useful.

packages/react-components/lib/banner-tab.tsx
```diff
@@ -0,0 +1,43 @@
...
+  logoPath?: string;
```
koonpeng, line 17: At least expose the onClick event so consumers is listen on it.

packages/react-components/lib/banner-tab.tsx
```diff
@@ -0,0 +1,43 @@
...
+}
+
+export const BannerTab = (props: BannerTabProps): JSX.Element => {
+  const { logoPath } = props;
+  const classes = useStyles();
+
+  return (
+    <Tabs
```
koonpeng, line 25: In the previous design, the banner is one of the tabs in the navigation bar, but now in this it's a separate tab bar in itself. In that case, it makes sense to change this to a image or button (or whatever material's best component for this use case). While you are at it, also rename this to avoid confusion.

This simplifies stuff as you no longer have to deal with all the complexity of a tab.

packages/react-components/stories/header-bar.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+export const NavBar: Story = () => {
+  const [value, setValue] = React.useState('building');
+
+  const onTabChange = (event: React.ChangeEvent<unknown>, newValue: string) => {
+    setValue(newValue);
+  };
+
+  return (
+    <>
+      <TabContext value={value}>
```
koonpeng, line 26: Not exactly sure how this works, but I don't think we should be using the TabContext ourselves, the new `NavigationBar` already contains a tab bar and iirc it will provide it's own TabContext.

koonpeng (commented):
@matiasbavera Can you check if the changes to reporting-server is correct?

matiasbavera:
> @matiasbavera Can you check if the changes to reporting-server is correct? 

Hey @vallq can you rebase this into main? Because if not this will replace all the changes I made on the report server

matiasbavera (changes requested):
> @matiasbavera Can you check if the changes to reporting-server is correct? 

Hey @vallq can you rebase this into main? Because if not this will replace all the changes I made on the report server

matiasbavera (changes requested):
Thanks Val, it seems that all */package-lock.json are modified which I think it's not the case. 

I'll recommend to do: 
`git checkout  origin/main  -- package-lock.json`
`git checkout  origin/main  -- packages/api-client/package-lock.json`
`git checkout  origin/main  -- packages/api-server/package-lock.json`
`git checkout  origin/main  -- packages/dashboard/package-lock.json`
`git checkout  origin/main  -- packages/dashboard/e2e/package-lock.json`
`git checkout  origin/main  -- packages/reporting/package-lock.json`
`git checkout  origin/main  -- packages/rmf-auth/package-lock.json`
`git checkout  origin/main  -- packages/rmf-models/package-lock.json`
`git checkout  origin/main  -- packages/reporting-server/package-lock.json`

and then run a `npm install` (on the root) and a `npm run bootstrap`


HTH

vallq:
Ok! Thanks @matiasbavera

koonpeng (approved):

PR #311, opened 2021-04-26
Feature/jwt authentication

ChawinTan:
## What's new

Related to [this](https://github.com/open-rmf/rmf_visualization/pull/15) pr

- United the 2 different websocket instances of default trajectory manager and negotiation status manager
- added the token parameter to the websocket request parameters
- enable dashboard to use jwt authentication
- handling of token expiry/token error
- resolve some warnings during npm build/install

There are currently no methods in `keycloak-js` to refresh id tokens. The work around I did was to log the user out automatically once the token expires and login again to start a new session. Below is a video demonstration of how it would work. (Happens at ~37 second mark in the video)

https://user-images.githubusercontent.com/29136440/116034631-2e7a3c00-a696-11eb-8530-06b4c20750a3.mp4

## Self-checks


koonpeng (changes requested):
Based on the video, I don't think we can use this, it would be horrible for the UX to require re-login every X seconds, note that this is even when the user is "active".

ChawinTan:
> Based on the video, I don't think we can use this, it would be horrible for the UX to require re-login every X seconds, note that this is even when the user is "active".

Right now the life span of token in one minute since we are still in development. I suppose in production the token lifespan would probably be a lot longer?

matiasbavera:
I'll like to point another thing too. 

I have a question about that, now we refresh our access token every 30 seconds (`const refreshed = await this._inst.updateToken(30);`), should we keep that 30 seconds on prod?

Thinking in an edge case (and maybe I'm wrong) but here goes an example:
id_token: 10:00 min.
access_token: 6 min
refresh_token_time: 3 min

1. User logs in. Got id and access
1. minute 3: refresh access and checks id on the server
1. minute 6: refresh access and checks id on the server
1. minute 9: refresh access and checks id on the server
1. minute 10: id token expires, but we are not refreshing the token on minute 10. So the user can still be on the app (let say changing the background color of his app in the browser and getting a lot of unauthorized requests).
1. minute 12: try to refresh the access token and logs out.  


If that the case, we have two ways of solving this, reduce the refresh token time or store on the browser persistent storage the expiration time of the id token and redirect to logout when `new Date() >=  expiration_date_id_token`

HTH

ChawinTan:
Actually, I was wondering if we can simply update the access token in exchange for a new refresh token? Is that good practice? Would the user have to be notified of the expiration of the id token somehow?

ChawinTan:
- Use access tokens instead of id tokens
- use a setInterval to refresh the tokens 5 seconds before the token expires. I chose not to use `onTokenExpired` in the [docs](https://github.com/keycloak/keycloak-documentation/blob/master/securing_apps/topics/oidc/javascript-adapter.adoc) as we are sending requests in intervals and that will result in some situations where an expired token is still in use which the token gets refreshed upon expiry.
- change the "client-id" of server-config.py to make e2e pass

One thing I realize is that if I navigate away from the dashboard for longer than 5 refreshes or around 5 minutes, the dashboard does not seem to detect the updated token and ends up throwing an error. I logged the user out in this scenario as it is better than showing a bunch of errors.

koonpeng (commented):
> use a setInterval to refresh the tokens 5 seconds before the token expires. I chose not to use onTokenExpired in the docs as we are sending requests in intervals and that will result in some situations where an expired token is still in use which the token gets refreshed upon expiry.

Doesn't the keycloak adapter support refreshing the token before it expires? On a sidenote the expiry for each token can be different so you should not use a fixed interval, just chain setTimeout based on the expiry of the new token.

> change the "client-id" of server-config.py to make e2e pass

I think it's normal for keycloak's access token to not include the client id by default, instead of "hacking" the server to accept the token, you should change keycloak's config to include the api url in the aud claim, while you are at it, might also want to change the name of the client id config to aud to avoid confusion. See https://auth0.com/docs/tokens/access-tokens/get-access-tokens, https://community.auth0.com/t/is-api-identifier-the-same-as-url-to-api-server/7535. We should make the default aud something like `localhost`. You might also want to look into if we should verify the `iss` claim based on the oauth2 spec, and include a config for that if we need to do so.

ChawinTan:
Added a new client scope config `rmf-dashboard`:
rmf_scope

Under the new scope, added a new mapper called `rmf-audience`, where the included audience include the existing `rmf-dashboard` client
rmf-audience

Added this scope to the `rmf-dashboard` client
add_rmf_dashboard_scope

All these changes are updated in the new `keycloak.mv.db` file under e2e. Material [referenced](https://stackoverflow.com/questions/53550321/keycloak-gatekeeper-aud-claim-and-client-id-do-not-match)

Other changes included:

- added a new function in authenticator to check if the existing access token needs to be refreshed.
- change the `client_id` parameter to `aud`
- added issuer check as it seems like [this a check needed](https://auth0.com/docs/tokens/json-web-tokens/validate-json-web-tokens) for jwt as a whole
- some other unit test updates

packages/reporting-server/rest_server/default_config.py
```diff
@@ -19,5 +19,8 @@
     "oidc_url": None,
     # client id registered with the authentication provider, this will be used to verify the
     # "aud" claim.
-    "client_id": "reporting",
+    "aud": "reporting",
+    # url or string that identifies the entity that issued the jwt token
+    # "iss" claim
```
koonpeng, line 24: explain what happens if issuer is `None`?

packages/api-server/api_server/default_config.py
```diff
@@ -18,5 +18,8 @@
     "oidc_url": None,
     # client id registered with the authentication provider, this will be used to verify the
```
koonpeng, line 19: This comment probably should be updated, the aud is no longer same as the client_id.

packages/dashboard/e2e/README.md
```diff
@@ -28,6 +28,8 @@ There are also some environment variables the test runner sets by default
 
 You can overwrite them by setting them in your environment variables.
 
+To run the e2e test locally, you need to replace the `iss` config value with this url: `config["iss"] = "http://localhost:8088/auth/realms/master"`
```
koonpeng, line 31: Add this automatically to the config?

packages/dashboard/src/app-config.ts
```diff
@@ -7,6 +7,7 @@ export interface AppConfig {
   authenticator: Authenticator;
   trajServerUrl: string;
   rmfServerUrl: string;
+  isKeycloak: boolean;
```
koonpeng, line 10: We can just use the `REACT_APP_AUTH_PROVIDER` env, suppose we have N auth providers, it wouldn't be nice to have `isProvider1`, `isProvider2`...

packages/dashboard/src/managers/robot-trajectory-manager.ts
```diff
@@ -62,37 +68,30 @@ interface Request {
...
+export class DefaultTrajectoryManager extends TrajectorySocketManager {
+  constructor(ws: WebSocket, authenticator?: Authenticator) {
+    super();
+    if (ws) this._webSocket = ws;
+    if (authenticator) this._authenticator = authenticator;
   }
 
   async latestTrajectory(request: TrajectoryRequest): Promise<TrajectoryResponse> {
-    const event = await this._send(JSON.stringify(request));
+    if (appConfig.isKeycloak) await this._authenticator?.refreshToken();
```
koonpeng, line 82: It should refresh the token regardless of the authenticator, the stub authenticator can just be a no-op.

packages/dashboard/src/managers/trajectory-socket-manager.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  /**
+   * Sends a message and waits for response from the server.
+   *
+   * @remarks This is an alternative to the old implementation of creating a promise, storing the
+   * resolver and processing each message in an event loop. Advantage of this is that each message
+   * processing logic can be self-contained without a need for a switch or if elses.
+   */
+  async _send(
```
koonpeng, line 28: public methods shouldn't be prefixed with _

packages/rmf-auth/lib/keycloak.ts
```diff
@@ -57,17 +57,35 @@ export default class KeycloakAuthenticator
...
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          username: (this._inst.idTokenParsed as any).preferred_username,
+          token: this._inst.token || '',
+        };
+        this.emit('tokenRefresh', null);
+      } else {
+        debug('token not refreshed');
+      }
+    }
+    return;
```
koonpeng, line 86: throw an error?

packages/rmf-auth/lib/keycloak.ts
```diff
@@ -57,17 +57,35 @@ export default class KeycloakAuthenticator
...
+      const refreshed = await this._inst.updateToken(5);
+      if (refreshed) {
+        this._user = {
+          // eslint-disable-next-line @typescript-eslint/no-explicit-any
+          username: (this._inst.idTokenParsed as any).preferred_username,
+          token: this._inst.token || '',
+        };
+        this.emit('tokenRefresh', null);
+      } else {
+        debug('token not refreshed');
```
koonpeng, line 83: throw an error?

koonpeng (changes requested):
There was a problem with trajectory and negotation server previously where by it couldn't detect if a response is meant for trajectory or negotiation, so sometimes it misses messages/throws exceptions because it thinks that it received invalid request/responses.

Has that been fixed and so now it's ok to mix the servers in the same websocket connection?

packages/dashboard/src/managers/trajectory-socket-manager.ts
ChawinTan: If this method is only intended for use in the child class, do you think marking it with an `_` as a reminder would be a good idea? I think I read an article before and the author does this, but I think it is his preference

packages/rmf-auth/lib/keycloak.ts
ChawinTan: Is this the same as above?

ChawinTan:
> There was a problem with trajectory and negotation server previously where by it couldn't detect if a response is meant for trajectory or negotiation, so sometimes it misses messages/throws exceptions because it thinks that it received invalid request/responses.
> 
> Has that been fixed and so now it's ok to mix the servers in the same websocket connection?

Tested the whole setup by running a few simulations via the rmf-panel. So far the responses are going to the right places. There is previously an error thrown in `_checkResponse` in robot-trajectory-manager but that has been fixed by removing the error thrown to returning a boolean and passing a response that is compatible to the `TrajectoryResponse` type

packages/rmf-auth/lib/keycloak.ts
ChawinTan: I think it is not good to throw an error here. If the token is not ready for refreshing because it is not `min validity` seconds prior to expiry, the error will break the whole system when it is functioning perfectly well.

packages/dashboard/src/managers/trajectory-socket-manager.ts
koonpeng: You can make it `protected` in this case. Also `_websocket` shouldn't have a _ prefix as well.

packages/rmf-auth/lib/keycloak.ts
koonpeng: yeah

ChawinTan:
Remove _ from websocket and added protected keyword.

packages/reporting-server/rest_server/default_config.py
```diff
@@ -18,7 +18,11 @@
...
     "oidc_url": None,
-    # client id registered with the authentication provider, this will be used to verify the
-    # "aud" claim.
-    "client_id": "reporting",
+    # Audience the access token is meant for. Can also be an array.
+    # Used to verify the "aud" claim.
+    "aud": "localhost",
+    # url or string that identifies the entity that issued the jwt token
+    # Used to verify the "iss" claim
+    # If iss is set to None, it means that keycloak authentication should be disabled
```
koonpeng, line 26: nit: in a way, keycloak is an implementation detail so should not be mentioned explicitly

koonpeng (approved):

PR #312, opened 2021-04-28
Fix reporting auth problem

matiasbavera:
## What's new

FluentD only supports Basic auth, so the two endpoints that FluentD uses to send data to the reporting server has BasicAuth, the rest of the endpoints (that are accessible from the UI) are working with tokens.

* Created a new BasicAutorization class. I'm using secrets as suggested here (https://fastapi.tiangolo.com/advanced/security/http-basic-auth/#fix-it-with-secretscompare_digest).
* Added tokens to reports

## Self-checks

### DIFF
--- a/example-deployment/k8s/fluentd-configmap.yaml
+++ b/example-deployment/k8s/fluentd-configmap.yaml
@@ -112,6 +112,11 @@ data:
       <buffer>
         flush_interval 10s
       </buffer>
+      <auth>
+        method basic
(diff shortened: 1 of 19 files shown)


koonpeng:
Is the reporting server acting like the "fluentd aggregator" in this diagram?


I wonder what is the best way to do this, if we are deploying using kubernetes, we can listen on a separate port that is not exposed to the outside, this would remove the need for ssl/auth. But this is only possible if we are using some container based deployment, it won't work on bare metal.

PR #313, opened 2021-04-29
reporting-sever:  hide fluentd ports from external services

matiasbavera:
## What's new
If this solution looks good to you, you can disregard and close https://github.com/open-rmf/rmf-web/pull/312 and merge this directly.

The two endpoints that FluentD uses to send data to the reporting server are now hidden unlike the rest of the endpoints (that are accessible from the UI).

We are running two instances of the app, one on port 8002 that only exposes the `report/` endpoint and the other on port 8003 that exposes the `log/` endpoint. 8003 can only be accessed from the Kubetenes cluster through a service and 8002 (reports) can be accessed from outside the cluster.

Exist 3 modes of running the all ALL: for development, expose all the endpoints, FLUENTD: to expose "/log/*" endpoints, and REPORT:  to expose only the "report/*" endpoints 

## Self-checks


packages/reporting-server/rest_server/__main__.py
```diff
@@ -1,18 +1,48 @@
...
 
+from .app import get_app
+from .app_config import SystemMode, app_config
 
-def main():
-    uvicorn.run(
-        app,
+
+class MyServer(Server):
+    # pylint: disable=invalid-overridden-method
```
koonpeng, line 10: Can you add a comment explaining why this is ok/needed?

packages/reporting-server/rest_server/__main__.py
```diff
@@ -1,18 +1,48 @@
...
+
+    apps.append(server1.run())
+    apps.append(server2.run())
+
+    return await asyncio.gather(*apps)
+
+
+def main():
+    loop = asyncio.get_event_loop()
+    loop.run_until_complete(run())
```
koonpeng, line 44: Downside of this is that uvicorn auto reload no longer works? I'm not sure if there is any way to get the best of both worlds except totally separating the 2 components, but then you can't run both components at once like this.

koonpeng (approved):
Does this supersedes #312 ?

packages/reporting-server/rest_server/__main__.py
matiasbavera: Yeah, if you run with `reporting_server` command it will not reload, but if we use this command `uvicorn --reload rest_server.app:get_app` for development, it will reload the app without any problems, because it'll run one instance of uvicorn (and you can configure what you want to expose in that one instance, all, report or fluentd).

packages/reporting-server/rest_server/__main__.py
matiasbavera: Sure thing! Thanks!

PR #314, opened 2021-04-30
reporting: Added more information to the rawlogs report

matiasbavera:

## What's new
* Added a endpoint to filter logs by containers 
* Containers are dynamically added because the data can be pruned at some point (7 days)
* Added kubernetes information to logs
* Fix to postgres  parallel writing on new database (FUP PR add a migration lib to handle db change)

image

## Self-checks

### DIFF
--- a/packages/react-components/lib/reports/log-management/log-table.tsx
+++ b/packages/react-components/lib/reports/log-management/log-table.tsx
@@ -5,7 +5,12 @@ import { makeStyles, Typography } from '@material-ui/core';
 import moment from 'moment';
 import { materialTableIcons } from '../../material-table-icons';
 
-export type LogRowsType = { level: string; message: string; created: string }[];
+export type LogRowsType = {
+  level: string;
+  message: string;
+  created: string;
+  payload: any;
+}[];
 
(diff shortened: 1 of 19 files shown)


packages/react-components/lib/reports/log-management/log-table.tsx
```diff
@@ -5,7 +5,12 @@ import { makeStyles, Typography } from '@material-ui/core';
 import moment from 'moment';
 import { materialTableIcons } from '../../material-table-icons';
 
-export type LogRowsType = { level: string; message: string; created: string }[];
+export type LogRowsType = {
+  level: string;
+  message: string;
+  created: string;
+  container_name: string;
```
koonpeng, line 12: should this be camelCase?

koonpeng (approved):

packages/react-components/lib/reports/log-management/log-table.tsx
matiasbavera: This is data coming from the backend so maybe it is a good idea to keep it as it is to differentiate from the FE pure data.

PR #303, opened 2021-04-14
Feature/theming

ChawinTan:
## What's new

#272 

Added the custom theme colors provided in figma.

Theme colors as shown in storybook:
theme_colors

We can add on more custom theming as more specifications come in.

- Trying to make the e2e work in the meantime

## Self-checks

- [x] I'm familiar with and follow this [Typescript guideline](https://basarat.gitbook.io/typescript/styleguide)
- [ ] I added unit-tests for new components
- [x] I tried testing edge cases
- [ ] I tested the behavior of the components that interact with the backend, with an e2e test

PR #321, opened 2021-05-07
Feature/theming and thememode

ChawinTan:
## What's new

Refer to [this pr](https://github.com/open-rmf/rmf-web/pull/303) for previous conversations.

- Added theme color
- Added settings to toggle between dark and light mode
- Implement dark mode

The image color inversion is done using a combination of different css filters. The filter parameters are derived from [this](https://gist.github.com/ChawinTan/cb5feaa6f4f837cc317e96b887668808) formula. Refer to [here](https://codepen.io/sosuke/pen/Pjoqqp) for a demonstration of the implementation. 

The inversion requires a combination of invert, sepia, saturate, hue-rotate, brightness, and contrast. I have tested and these operations are not very computationally intensive and does not really affect the performance of the app.

Photos for reference:
dark_mode_with_traj
dark_theme_leaflet
dark_mode_2
light_mode_1
light_mode_2

Organization:
Theme_color_organization

## Self-checks


packages/dashboard/src/util/theme.ts
```diff
@@ -0,0 +1,63 @@
...
+  components: {
+    backgroundColor: theme.palette.primary.dark,
+    color: theme.fontColors.darkTheme,
+  },
+  background: {
+    backgroundColor: theme.palette.secondary.dark,
+  },
+  map: {
+    filter:
+      'invert(90%) sepia(12%) saturate(5773%) hue-rotate(193deg) brightness(92%) contrast(92%)',
```
koonpeng, line 14: I guess these filters are based off the office map, can you test it with other maps in rmf_demos?

packages/dashboard/src/util/theme.ts
```diff
@@ -0,0 +1,63 @@
...
+  font: {
+    color: theme.fontColors.lightTheme,
+  },
+  leafletControl: {
+    color: `${theme.fontColors.lightTheme} !important`,
+    backgroundColor: `${theme.palette.primary.main} !important`,
+  },
+}));
+
+export const decideThemeStyle = (themeContext: ThemeMode) => {
```
koonpeng, line 61: This causes a lot of prop drilling, theming is universal enough that you should use a context to pass these.

Also not sure if these are a good idea, each component should set their own styles following the theme engine. If a component is not following the theme, it should be fixed by making it use the theme properly instead of applying a fixed class on top of it.

koonpeng (changes requested):

koonpeng:
> dark_theme_leaflet

Noticed there is a scrollbar at the bottom, not sure if this is a bug caused by the theming or if it's already there.

packages/dashboard/src/util/theme.ts
ChawinTan: I guess passing in the current themeMode would be a better idea instead of passing the css, then let each component decide what style they are going to have

matiasbavera (commented):
Thanks for this, looking good. As KP mentioned it will be good for you to try with the others maps as well.

ChawinTan:
Created 2 different MuiTheme instead, one for dark theme and one for light theme. The theme provider is shifted into appbase so that a change in the settings can trigger the toggling of the 2 different themes.

```
preferDarkMode ? darkTheme : lightTheme
```
This eliminates the need for conditional statements and minimal prop drilling to toggle the styles, apart from the maps and leaflet elements on the schedule visualizer.

Apart from this, I changed the refs on the `ImageOverlay` to an array of refs so that it is possible to handle demos with multiple maps.

Airpot dark mode:
airpot_dark

Clinic dark mode:
clinic

packages/dashboard/src/components/dashboard/tour/tests/__snapshots__/tour.test.tsx.snap
```diff
@@ -31,7 +31,7 @@ exports[`Dashboard Tour renders without crashing 1`] = `
         "content": [Function],
```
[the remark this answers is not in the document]
ChawinTan: I think lets just keep it for now. Dun think there are any concrete plans yet

koonpeng (changes requested):
It appears this has broken the theming for material. I think this is because there are several points where it set some css like `backgroundColor: theme.palette.secondary.main,`. material has palette for things like background, doing so overwrites those colors and instead use the secondary color as the background, in most cases, the secondary color is not meant for such use cases so it breaks the default theme.

Instead of setting css like the above, properly define background color in the theme palette, this probably also goes for font colors etc.

I added a global storybook toolbar that changes the theme of the components in storybook, it will be useful to test out the themes, look for the circle near the top of storybook.

pic

koonpeng:
> Created 2 different MuiTheme instead, one for dark theme and one for light theme. The theme provider is shifted into appbase so that a change in the settings can trigger the toggling of the 2 different themes.
> 
> ```
> preferDarkMode ? darkTheme : lightTheme
> ```
> 
> This eliminates the need for conditional statements and minimal prop drilling to toggle the styles, apart from the maps and leaflet elements on the schedule visualizer.
> 
> Apart from this, I changed the refs on the `ImageOverlay` to an array of refs so that it is possible to handle demos with multiple maps.
> 
> Airpot dark mode:
> airpot_dark
> 
> Clinic dark mode:
> clinic
> 
> `I guess these filters are based off the office map, can you test it with other maps in rmf_demos?`
> 
> Actually, it is generated based on the primary color of the dark theme, starting with black as the base color for iteration. Basically, the algorithm iterates till the difference between the actual color and the generated color is less than an amount.

The airport map turns out to be like some dystopian cyber punk map, hopefully we can get more control/rules of the original colors of the maps used. Worst case I guess we could just serve 2 maps from the api-server, and manually process the maps for better colors.

packages/react-components/lib/logo-button.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { Button, createStyles, makeStyles } from '@material-ui/core';
 
-const useStyles = makeStyles((theme) =>
+const useStyles = makeStyles(() =>
   createStyles({
     root: {
       '&$disabled': {
-        backgroundColor: theme.palette.primary.dark,
-        color: theme.palette.primary.dark,
+        backgroundColor: '#1F396B',
```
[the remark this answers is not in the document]
ChawinTan: The logo would be too white to see in the light theme, so I set its background to the dark theme color. I forgot to bring this up, but should we also make a logo that is suitable for the light theme that is darker in the font?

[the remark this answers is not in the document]
koonpeng: Sounds good

ChawinTan:
- Removed the primary and secondary override from the palette. The default is working properly now.
- Added a dark blue logo (thanks @vallq for the logo!)

resolve_default_mui_theme_issues

koonpeng (changes requested):
The border colors and other parts still look wrong,

Looking at the material ui source code, https://github.com/mui-org/material-ui/blob/6658323b5e2ba366691829bbef3f53c00008497e/packages/material-ui/src/Button/Button.js#L73.

The border color is hardcoded according to the palette "type", there is no border color palette, I'm guessing the reason for that is either because of an oversight, or it is not practical as each component needs to have a slightly different border color.

We should keep compatibility with the default material theme. We can either tune the theme to work well with the hardcoded colors, or extend any of the material components with our own theming extensions, but if we do that we should keep compatibility with the existing themes, so instead of setting the border color to `palette.success`, introduce a new palette and use that color only if it is in the palette. If a theme without the palette is defined, just use the default color, aka set `borderColor: undefined`, or do not add the css.

Or better yet, instead of extending each of the material components, you may try this https://material-ui.com/customization/components/#global-css-override. The advantage should be that the styles are transparent to the components, so you can continue to use `import Button from @material/core/Button` instead of `import { Button } from 'react-components`.

ChawinTan:
- add global css for dividers and ExpandMoreIcon
- update robot and task panels with the latest changes
- decided to leave buttons in door accordian as it is since the `open` and `close` should be indicative enough

koonpeng (changes requested):
The global css should be in `react-components`.

In the storybook preview, the divider colors are still wrong.

Many components are not following the theme:
* systems summary components
* "Main Level Menu"
* logging components
* notification dialog
* omnipanel background

All the components should automatically follow the theme without having to set per theme special stylings. For example, `DoorAccordion` sets 
```tsx
  button: {
    backgroundColor: theme.secondaryBackground,
    color: theme.fontColors,
  },
  root: {
    backgroundColor: theme.secondaryBackground,
    color: theme.fontColors,
  },
```
These styling should not be set for the default material theme because they mess up the theme colors. I'm guessing they are set because rmf themes does not properly sets the palette, so instead of hacking the theme for each component, properly set the theme palette for all components to follow.

The way we are determining trajectory colors does not work well with different themes. tinyRobot for example has a blue trajectory which does not go well with the rmf-dark theme. @vallq @codebot @matiasbavera  Do you see this as a problem? Should we change the way we determine trajectory colors so they won't have colors that blend into the theme?

ChawinTan:
> These styling should not be set for the default material theme because they mess up the theme colors. I'm guessing they are set because rmf themes does not properly sets the palette, so instead of hacking the theme for each component, properly set the theme palette for all components to follow.

I still don't really understand this part. By `theme palette`, do you mean override material ui's original components in `global css` using, for eg: `theme.fontColors` for buttons as the default theme? When the default theme does not look good then manually override the default theme in `react-components`?

Logging components and "multi-level-menu" are not using the themes currently. Do they need to be included as well apart from inside the storybook?

koonpeng:
material theme palette contains options like primary, secondary colors, font colors, background colors etc. Setting these correctly should have the components use the colors without having to set css stylings manually.

In some cases like border colors where material theme does not have a palette for, introduce a new field for that and modify the global css to use those colors. But keep the modification backwards compatible, so only set the border color if the theme defines it, meaning if the theme in use is the default material theme, there should be no changes from the modifications.

Ultimately we want to be able to either material's default theme or rmf theme in the theme provider context without either of them breaking. The default material theme should also look exactly the same as what material ui would normally look like.

ChawinTan:
Thanks for clarifying! Didn't realize there was already a background palette that material already provided.

vallq:
> The way we are determining trajectory colors does not work well with different themes. tinyRobot for example has a blue trajectory which does not go well with the rmf-dark theme. @vallq @codebot @matiasbavera Do you see this as a problem? Should we change the way we determine trajectory colors so they won't have colors that blend into the theme?

For trajectory colours, is it possible to have 2 "themes" for them as well? A lighter version to use on the dark mode and darker colours on the normal mode. To avoid the abovementioned accidental blending into the background colour.

I think this can be a separate task as well so that this PR doesn't get too complicated?

koonpeng:
> For trajectory colours, is it possible to have 2 "themes" for them as well? A lighter version to use on the dark mode and darker colours on the normal mode. To avoid the abovementioned accidental blending into the background colour.

Yeah, it can be an extension to the material theme, the hardest part is coming up with an algorithm that can generate good colors for each theme.

packages/react-components/lib/navigation-bar.tsx
```diff
@@ -2,15 +2,12 @@ import { createStyles, makeStyles } from '@material-ui/core';
 import Tabs from '@material-ui/core/Tabs';
 import React from 'react';
 
-const useStyles = makeStyles((theme) =>
+const useStyles = makeStyles(() =>
   createStyles({
     tabsContainer: {
+      borderRight: '0.25px solid rgba(169,169,169, 0.5)',
+      borderLeft: '0.25px solid rgba(169,169,169, 0.5)',
```
koonpeng, line 9: I think we should use theme values here.

packages/react-components/lib/omni-panel.tsx
```diff
@@ -24,8 +24,7 @@ const useStyles = makeStyles((theme) => ({
   },
   navigationButton: {
     borderRadius: 'inherit',
-    backgroundColor: theme.mainBackground,
-    color: theme.fontColors,
+    borderColor: theme.palette.divider,
```
koonpeng: I think the default border color is not the same as the divider color, so doing this would mess up the default material theming.

packages/react-components/lib/robots/robot-info.tsx
```diff
@@ -20,8 +20,7 @@ const useStyles = makeStyles((theme) =>
   createStyles({
     root: {
       '&$disabled': {
-        color: theme.fontColors,
-        borderColor: theme.fontColors,
+        borderColor: theme.palette.divider,
```
koonpeng, line 23: I think the default border color is not the same as the divider color, so doing this would mess up the default material theming.

packages/react-components/lib/simple-info.tsx
```diff
@@ -54,7 +56,9 @@ const useStyles = makeStyles((theme) => ({
     display: 'table-cell',
     textAlign: 'end',
     borderBottom: '1px solid',
+    borderBottomColor: theme.palette.divider,
     borderTop: '1px solid',
+    borderTopColor: theme.palette.divider,
```
koonpeng: I think the default border color is not the same as the divider color, so doing this would mess up the default material theming.

packages/react-components/stories/theme.stories.tsx
```diff
@@ -1,8 +1,8 @@
 import { makeStyles, Paper, Typography } from '@material-ui/core';
-import { ThemeProvider } from '@material-ui/styles';
+// import { ThemeProvider } from '@material-ui/styles';
```
koonpeng, line 2: leftovers?

koonpeng (commented):
Regarding the border colors, I believe they are hardcoded for certain components like buttons. Setting the border color to divider color can mess up the button look. Try and see if you can dynamicly set global css based on the theme. Use the theme border color if it specifies it, else do not change the border color so material defaults to the hard coded values.

packages/react-components/stories/trashbin-control-button-group.stories.tsx
```diff
@@ -11,5 +11,7 @@ export default {
 } as Meta;
 
 export const SimpleTrashBinControlButtonGroup: Story = (args) => (
-  <TrashBinControlButtonGroup {...args} />
+  <div style={{ backgroundColor: '#A8A8A8', padding: '1rem' }}>
```
koonpeng, line 14: Why hardcoded colors?

packages/dashboard/src/settings.ts
```diff
@@ -29,5 +35,6 @@ export function defaultSettings(): Settings {
   return {
     trajectoryAnimation: TrajectoryAnimation.Follow,
     themeMode: ThemeMode.Light,
+    useTheme: UseTheme.False,
```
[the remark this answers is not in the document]
ChawinTan: This is give users the option to switch back to the original material theme we are using. Since this feature is optional and not really complete yet, I thought it would be good to let users have this option.

packages/react-components/lib/themes/global-theme.tsx
```diff
@@ -2,22 +2,24 @@ import { withStyles } from '@material-ui/core/styles';
 
 export const GlobalCss = withStyles((theme) => ({
   '@global': {
+    // override accordian background to distinguish it from onipanel
```
koonpeng, line 5: Sounds like we should change the palette of omnipanel or the accordions used in it instead. The global css changes every instance of the component, it should not depend specific components.

packages/react-components/lib/themes/global-theme.tsx
```diff
@@ -2,22 +2,24 @@ import { withStyles } from '@material-ui/core/styles';
...
     '.MuiAccordion-root': {
       backgroundColor: theme.palette.background.default,
     },
+    // override text color of disabled button to make it visible
```
koonpeng, line 9: I don't really agree with this change. Disabled text have their own color so that they look disabled.

packages/react-components/lib/themes/global-theme.tsx
```diff
@@ -2,22 +2,24 @@ import { withStyles } from '@material-ui/core/styles';
...
     '.MuiButtonBase-root.Mui-disabled': {
       color: theme.palette.text.primary,
     },
-    '.MuiAppBar-root': {
-      backgroundColor: theme.palette.primary.main,
-    },
+    // highlight table header in panels
```
koonpeng, line 13: Whats the reasoning to change this? It doesn't seem to help to make the table head more prominent in rmf-dark.

packages/react-components/lib/themes/global-theme.tsx
```diff
@@ -2,22 +2,24 @@ import { withStyles } from '@material-ui/core/styles';
...
     '.MuiTableCell-head': {
       background: 'rgba(0, 0, 0, 0.1)',
       borderBottom: 'none',
     },
+    // override text color of forms to be visible in theme background colors
```
koonpeng, line 18: This seems more like a problem with the theme. Is it possible to fix the theme so it has better colors instead?

packages/react-components/lib/themes/global-theme.tsx
```diff
@@ -2,22 +2,24 @@ import { withStyles } from '@material-ui/core/styles';
...
     '.MuiFormLabel-root': {
       color: theme.palette.text.primary,
     },
+    // text color of forms to distinguish them when focus
```
koonpeng, line 22: This and the above form css breaks how form looks.

Without global css
image

With global css
image

It's no longer able to differentiate between the hint text and the actual text.

Even with the global css, forms still look broken in both rmf-light and rmf-dark

image
Buttons color same as background, bad text colors.

image
Buttons color same as background, wrong colors some elements (date selectors, text field underlines etc).

The "cancel" and "ok" buttons in date selector becomes completely invisible
image

packages/react-components/lib/themes/rmf-light.tsx
```diff
@@ -1,6 +1,15 @@
...
 
+declare module '@material-ui/core/styles/createMuiTheme' {
+  interface Theme {
+    mapClass: string;
+  }
+  interface ThemeOptions {
+    mapClass: string;
+  }
+}
+
```
koonpeng, line 12: This is duplicated in rmf-dark.

packages/react-components/stories/omni-panel.stories.tsx
```diff
@@ -79,10 +79,12 @@ const SimpleOmniPanel = (props: SimpleOmniPanelProps): JSX.Element => {
 
 export const SimplePanelFixedSize: Story = (args) => {
   return (
+    // add a background to visualize different theme colors properly
```
koonpeng, line 82: `rmf-dark` is "broken" with this hardcoded background color.

image

Without it, it is even more broken
image

I think there is some problem with the theme itself.

packages/react-components/stories/trashbin-control-button-group.stories.tsx
```diff
@@ -11,6 +11,7 @@ export default {
 } as Meta;
 
 export const SimpleTrashBinControlButtonGroup: Story = (args) => (
+  // add a background to visualize different theme colors properly
```
koonpeng, line 14: Same as omnipanel problem.

koonpeng (changes requested):

packages/dashboard/src/settings.ts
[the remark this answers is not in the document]
koonpeng: The problem is that even with `useTheme === false`, there are still some changes in this PR to the material theme. It is also confusing to have such an option, better to just make light theme be the material default and dark be rmf-dark in that case. Or have the theme be a dropdown with 3 options.

packages/dashboard/src/components/app-base.tsx
```diff
@@ -19,6 +19,8 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
+export const defaultTheme = createMuiTheme();
```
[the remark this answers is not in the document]
ChawinTan: I guess I missed that. Will remove the default theme

packages/react-components/stories/omni-panel.stories.tsx
ChawinTan: That's because the omnipanel itself does not have any background color as it is wrap in a div. The background color is passed down from classes in dashboard. For now, I set the background to the same value of what is defined in dashboard

packages/react-components/stories/trashbin-control-button-group.stories.tsx
ChawinTan: same as omnipanel

packages/react-components/lib/themes/global-theme.tsx
ChawinTan: The header looks like it was blending into the background and I thought this would make it look better. Removed it for now.

packages/react-components/stories/omni-panel.stories.tsx
koonpeng: Can we fix it by giving it a background color or making it wrapped with a `Paper` instead?

koonpeng (changes requested):
image

The underlines and borders blends in too much. Buttons are using the default colors which looks off.

image

Headers are using default colors, which look off, some text like the days (mon, tue) blends into the background and is unreadable.

image

Separators became very distinct, messing up material's default theme.

image

This don't work well with the new theme, the text are not very readable.

image

This might be the same problem as the omnipanel previously.

image

Text are unreadable

---

Haven't tested all the components yet, but it seems the theme is still pretty broken.

ChawinTan:
- Modified the theme using material ui's `type: light/dark` option. Went through all the components and this should take care of the inconsistent font and border colors.
- Made some changes to the admin panel
- Gave the dark theme a new primary color that is complemetary
- Made some changes to the theme colors

I remove the color override for the headers and I think it looks better without overriding it with the new background colors. Should we override it with the background colors or should we stick the the default?

[For personal reference, link to color system: https://material.io/design/color/the-color-system.html#tools-for-picking-colors]

koonpeng (commented):
image

By default, material should have a border/separator between the rows, maybe some css changes caused them to be removed.

@vallq Can you look at the color palette? Are these the expected colors we are aiming for?

ChawinTan:
Just to confirm, should the theme toggle be in the admin panel now instead of settings?

vallq:
Screenshot at 2021-08-25 17-52-54
image

I think the font colors for the info panel should be white? it looks like grey on my screen so it looks faded out and hard to view. I think the table should also include the default table lines, but I'm not sure if it will be easy to see on dark mode (hence on the ui it's a shadow to the bottom border)

the other thing would be to highlight the different states with different colours, but I don't think that is "urgent" because it's not on the main branch as well, and we could revisit whether this is a good idea or not

thanks for all the work put into this @ChawinTan

koonpeng:
Can you merge main and fix the conflicts?

ChawinTan:
Fix conflicts and make some minor style fixes.

packages/react-components/lib/error-overlay.stories.tsx
```diff
@@ -11,13 +11,14 @@ export default {
...
   // override style with userSelect disabled
-  const useStyles = makeStyles(() => ({
+  const useStyles = makeStyles((theme) => ({
     container: {
       display: 'table',
       borderCollapse: 'collapse',
       width: '100%',
       overflowX: 'auto',
       userSelect: 'none',
+      backgroundColor: theme.palette.background.paper,
```
koonpeng, line 21: We probably shouldn't be specifying this in a story, if this is needed it should be in the component.

packages/react-components/lib/robots/robot-info.tsx
```diff
@@ -134,8 +115,8 @@ export function RobotInfo({ robot }: RobotInfoProps): JSX.Element {
           <Button
             disableElevation
             variant="outlined"
-            classes={{ root: classes.root, disabled: classes.disabled }}
-            disabled
+            className={classes.button}
+            disableRipple={true}
```
koonpeng, line 119: Why disable the ripple effect?

packages/react-components/lib/simple-filter.tsx
```diff
@@ -12,9 +12,36 @@ export interface SimpleFilterProps {
   value: string;
 }
 
-const useStyles = makeStyles(() => ({
+const CustomTextField = withStyles((theme) => ({
```
koonpeng, line 15: Why can't we use the normal text field?

packages/dashboard/src/components/tasks/task-panel.tsx
```diff
@@ -164,17 +167,17 @@ export function TaskPanel({
                 }}
                 aria-label={`${autoRefreshTooltipPrefix} auto refresh`}
               >
-                <AutorenewIcon />
+                <AutorenewIcon className={classes.icons} />
```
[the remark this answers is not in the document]
ChawinTan: I think this is a relic of the old and bad design. Removed them already

packages/react-components/lib/robots/robot-info.tsx
ChawinTan: So as to mimic a disabled button without having to modify the classes to change colors to make it more visible.

koonpeng: In that case, you should set the `component` prop to `div` so as to correctly let the accessibility engine know that it is not a button.

PR #323, opened 2021-05-07
send only when websocket is ready

ChawinTan:

## What's new

See linked issue for the screen shot of the error that happens when toggling between panels. In short, the websocket seems to try to connect to the server again when toggling between panels. 

The fix is to check if websocket is ready before sending.

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/app.tsx
+++ b/packages/dashboard/src/components/app.tsx
@@ -44,7 +44,7 @@ export default function App(): JSX.Element | null {
   const [authInitialized, setAuthInitialized] = React.useState(!!appConfig.authenticator.user);
   const [authenticator, setAuthenticator] = React.useState(appConfig.authenticator);
   const [user, setUser] = React.useState<User | null>(appConfig.authenticator.user || null);
-  const [ws, setWs] = React.useState<WebSocket | undefined>(undefined);
+  const [ws, setWs] = React.useState<WebSocket>(new WebSocket(appConfig.trajServerUrl));
   const appRoutes = [DASHBOARD_ROUTE, TASKS_ROUTE];
   const [tabValue, setTabValue] = React.useState<TabValue | null>(() =>
     locationToTabValue(window.location.pathname),
--- a/packages/dashboard/src/components/app.tsx
+++ b/packages/dashboard/src/components/app.tsx
@@ -44,6 +44,7 @@ export default function App(): JSX.Element | null {
   const [authInitialized, setAuthInitialized] = React.useState(!!appConfig.authenticator.user);
   const [authenticator, setAuthenticator] = React.useState(appConfig.authenticator);
   const [user, setUser] = React.useState<User | null>(appConfig.authenticator.user || null);
+  const [ws, setWs] = React.useState<WebSocket | undefined>(undefined);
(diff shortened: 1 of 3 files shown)


packages/dashboard/src/managers/trajectory-socket-manager.ts
```diff
@@ -32,7 +32,7 @@ export default class TrajectorySocketManager extends EventEmitter<Events> {
     if (!webSocket) throw Error('Websocket not initialized!');
     // response should come in the order that requests are sent, this should allow multiple messages
     // in-flight while processing the responses in the order they are sent.
-    webSocket.send(payload);
+    if (webSocket.readyState === WebSocket.OPEN) webSocket.send(payload);
```
koonpeng, line 35: wouldn't this cause an infinite wait as no request is actually sent? You probably should throw an error in this case.

ChawinTan: I thought the connection to the backend is made here?

```
<UserContext.Provider value={user}>
    <TrajectorySocketContext.Provider value={new WebSocket(appConfig.trajServerUrl)}>
        <ThemeProvider theme={theme}>
```

Once the connection to the backend is successful, then the websocket's ready state will be marked as `open` and the request would be sent.

Anyway, I also realized that switching between the tabs causes multiple client connections to be made to the dashboard, make it very laggy after several switches. I'm also going to solve this problem in this pr too

koonpeng: This PR adds a condition so that it only sends a request if the websocket is opened, if it is not, it continues execution without sending anything, so then later in this part

```tsx
    this._ongoingRequest = new Promise((res) => {
      this._listenOnce('message', webSocket, (e) => {
        this._ongoingRequest = null;
        res(e);
      });
    });
```

We create a promise that is resolved when we get a `message` event. since we never sent any requests, it's possible that there will never be a response so the promise would never get resolved.

ChawinTan:
Removed the if statement for `ready state` since storing websocket in a state does not cause it to make a new connection everytime it toggles between tabs.

Remove `useMemo` in schedule-visualizer for `negoTrajs` as it does not render when `negotiationTrajStore`'s value changes.

packages/dashboard/src/components/app.tsx
```diff
@@ -44,6 +44,7 @@ export default function App(): JSX.Element | null {
   const [authInitialized, setAuthInitialized] = React.useState(!!appConfig.authenticator.user);
   const [authenticator, setAuthenticator] = React.useState(appConfig.authenticator);
   const [user, setUser] = React.useState<User | null>(appConfig.authenticator.user || null);
+  const [ws, setWs] = React.useState<WebSocket | undefined>(undefined);
```
koonpeng, line 47:
```suggestion
  const [ws, setWs] = React.useState<WebSocket>(() => new WebSocket(appConfig.trajServerUrl));
```

Would this do the same thing as the effect?

ChawinTan: Tested and it seems to be fine.

koonpeng (approved):

PR #335, opened 2021-05-17
react-components: robot panel

vallq:
## What's new
Robot Panel component based off from Task Panel
robot panel ui
- will be removing the refresh button

## Self-checks

## Discussion
1. I'm still not sure how to use the `api_server`'s implementation of task progress in `react-components` because what react-components use is technically just dummy data. should this be taken as a prop and only implemented in the dashboard?

~2. The original idea was to also have 'Est End Time' be a clock value in 12hour format, and likewise, I'm not sure how this can be done. using `rosTimeToJs` in `lib/utils.ts`?~


packages/react-components/lib/robots/robot-panel.tsx
```diff
@@ -0,0 +1,73 @@
...
+              rowsPerPageOptions: [10],
+              page,
+              onChangePage: (_ev, newPage) => setPage(newPage),
+            }}
+            onRobotClick={(_ev, robot) => setSelectedRobot(robot)}
+            onRefreshClick={onRefreshClick}
+          />
+        </Grid>
+        <Paper className={classes.detailPanelContainer}>
+          {selectedRobot ? <RobotInfo robot={selectedRobot} /> : <NoSelectedRobot />}
```
matiasbavera, line 68: Does it improve the performance if you store this on a `React.useMemo` or using `React.memo` on `RobotInfo ` component?

matiasbavera (changes requested):
Hi Val, left a couple of comments, HTH

koonpeng (commented):
I'm going to push some changes to this branch that adds some features to make query the tasks/robots easier.

When using the new task query api, you can pass the list of tasks as an object, you can get the task_id with RobotState.task_id. On the dashboard side you can

1. Use `/fleets` to get a list of fleets.
2. Use `/fleets/{name}/state` to get the state of each fleet (and robots), this is inefficient as it means sending a request for each fleet, but it is the best we have now.
3. Use `/tasks?task_id=a,b,c` to get the task summary of each task.

koonpeng:
from the other pr:
> oh okay!
other than the current tasks, is it possible to retrieve all assigned and active tasks? (completed/failed tasks are not needed). because there is an 'assigned tasks' section in the robot-info.tsx

oh I see, the information in robot state is not enough, I think disregard the above guide. Just use the task query api the find the list of assigned tasks for a robot. The new additions allow you to query by multiple robot/fleet/states so you should be able to get the required data.

packages/react-components/lib/robots/utils.ts
```diff
@@ -29,3 +29,53 @@ export function robotModeToString(robotMode: RmfModels.RobotMode): string {
...
+  robots: RmfModels.RobotState[],
+  tasks: RmfModels.TaskSummary[],
+): VerboseRobot[] {
+  const removableTaskStates = [
+    RmfModels.TaskSummary.STATE_ACTIVE,
+    RmfModels.TaskSummary.STATE_PENDING,
+    RmfModels.TaskSummary.STATE_QUEUED,
+  ];
+  const robotsWithAssignedTasks = robots.map((robot) => {
+    const assignedTasks = tasks.filter((task) => {
```
koonpeng, line 73: I think we should add a TODO to move this to the server. The server will have a robots endpoints returning the equivalent of `VerboseRobot`.

packages/react-components/lib/robots/robot-table.tsx
```diff
@@ -0,0 +1,190 @@
...
+  }
+}
+
+export interface RobotTableProps extends PaperProps {
+  /**
+   * The current list of robots to display, when pagination is enabled, this should only
+   * contain the robots for the current page.
+   */
+  tasks: RmfModels.TaskSummary[];
+  robots: RmfModels.RobotState[];
```
koonpeng, line 128: Could this take in a `VerboseRobot` instead?

koonpeng:
I realised the robots endpoints will need more work to get the required info, so I think I will revert the api-server changes and work on it on another branch.

vallq:
updated branch: `dashboard/robot-page`

PR #338, opened 2021-05-18
E2E package migration

ChawinTan:

## What's new

- Shifted e2e package from dashboard package into it's own package.
- Combine door-request test, filtering test and tasks test into ui-interaction test

## Self-checks


koonpeng (changes requested):
Instead of `e2e` can you rename it to `dashboard-e2e`, because there might be other e2e tests from reporting etc.

packages/dashboard/scripts/setup/get-icons.js
```diff
@@ -4,10 +4,8 @@ const fs = require('fs');
 const chalk = require('chalk');
 const path = require('path');
 
-const curWorkingDir = process.cwd();
-let file = curWorkingDir.includes('dashboard')
-  ? './.resources.json'
-  : './dashboard/.resources.json';
+const basePath = __dirname.slice(0, __dirname.length - 'scripts/setup'.length);
```
koonpeng, line 7: nit: This might not work in windows (not that we support it), using the `path` api, or using `..` to go back one level is more cross platform friendly.

PR #340, opened 2021-05-19
dashboard/react-components: robot-page

vallq:
## What's new
Robot Panel component based off from Task Panel
robot panel ui
https://user-images.githubusercontent.com/28668944/118747937-866b2380-b88d-11eb-9fae-ccff90669be0.mp4
addition of robot page to the dashboard

## Self-checks

## Discussion
1. I'm still not sure how to use the `api_server`'s implementation of task progress in `react-components` because what react-components use is technically just dummy data. should this be taken as a prop and only implemented in the dashboard?


vallq:
@koonpeng @matiasbavera hello, this is the implementation of the robots page. my current question is as described above, but I think the rest API should be able to resolve that? not entirely sure.

also I originally wanted to use antd's progress bar component to indicate task progress status because mui does not have such a component. but it would be "overkill" to include this, right? was thinking that a number would suffice.
tasks_v1

matiasbavera:
The truth is that this is neat but it seems asymmetrical. For testing purposes, what do you think about putting `Battery` and` Assigned Tasks` centered to see how it looks

image

koonpeng:
> I'm still not sure how to use the api_server's implementation of task progress in react-components because what react-components use is technically just dummy data. should this be taken as a prop and only implemented in the dashboard?

1 way is to `import type {} from 'api-client`, another is to just take in a string, the dashboard will handle extracting the correct data from the api response.

vallq:
@matiasbavera I'm not sure about the GH link you shared because it uses the original material UI library and it also uses `styled-components` which is something that we do want to remove, right?
I did find another: https://github.com/martyan/react-customizable-progressbar, which is also more updated. just that there's some webpack version incompatibility if I run the install line. will try this out and see!

@koonpeng ooo okay, will try that, thank you

packages/dashboard/src/components/robots/robot-page.tsx
```diff
@@ -0,0 +1,76 @@
...
+    fleetNames.current = newFleetNames;
+  }
+  const [taskSummaries, setTaskSummaries] = React.useState<RmfModels.TaskSummary[]>([]);
+  const [robotStates, setRobotStates] = React.useState<RmfModels.RobotState[]>([]);
+
+  const handleRefresh = React.useCallback(async () => {
+    if (!tasksApi) {
+      return;
+    }
+    const tasks = await tasksApi.getTasksTasksGetTasksGet();
```
koonpeng, line 51: In order to save bandwidth, we need to add a robot query api to the server. The returned result would include the current task of each robot. It will be similar to the new tasks query api added in #341.

I can help add the new api, is there any other data that will be needed?

vallq: oh okay!
other than the current tasks, is it possible to retrieve all assigned and active tasks? (completed/failed tasks are not needed). because there is an 'assigned tasks' section in the robot-info.tsx

vallq:
Original left alignment
left align
Centered battery and assigned tasks
centered
The battery bar is already centered but it looks otherwise because there is white space after the '%'
alignment
Another option I can think of, but might get a little messy?
image

matiasbavera:
> Another option I can think of, but might get a little messy?
> image

Would it lose context if we do something like this?
``` 
               Battery

           Task progress

Assigned Task     Est. End time

Location              Destination
```

vallq:
I've addressed the comments made in the previous branch pr (`ui/robot-panel`). Thanks @matiasbavera and @koonpeng 
Some of the comments with 'VerboseRobot' updates, I think those can be made in another PR after this and the robot endpoints are merged in?

packages/react-components/lib/robots/utils.ts
```diff
@@ -29,3 +30,58 @@ export function robotModeToString(robotMode: RmfModels.RobotMode): string {
       return `Unknown (${robotMode.mode})`;
   }
 }
+
+//TODO: endpoint should return the equivalent of this
+export interface VerboseRobot {
+  name: string;
+  model: string;
+  task_id: string;
```
koonpeng, line 38: should be `camelCase`.

packages/react-components/lib/robots/utils.ts
```diff
@@ -29,3 +30,58 @@ export function robotModeToString(robotMode: RmfModels.RobotMode): string {
...
+  seq: number;
+  mode: RmfModels.RobotMode;
+  battery_percent: number;
```
koonpeng, line 41: should be `camelCase`.

packages/react-components/lib/robots/utils.ts
```diff
@@ -29,3 +30,58 @@ export function robotModeToString(robotMode: RmfModels.RobotMode): string {
...
+  location: RmfModels.Location;
+  path: RmfModels.Location[];
+  assigned_tasks: TaskProgress[];
```
koonpeng, line 44: should be `camelCase`.

packages/react-components/lib/robots/utils.ts
```diff
@@ -29,3 +30,58 @@ export function robotModeToString(robotMode: RmfModels.RobotMode): string {
       return `Unknown (${robotMode.mode})`;
   }
 }
+
+//TODO: endpoint should return the equivalent of this
+export interface VerboseRobot {
```
koonpeng, line 35: If you mean to extend the base `RobotState` you can do
```tsx
export interface VerboseRobot extends RmfModels.RobotState {
```

packages/react-components/stories/circular-progress-bar.stories.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Story, Meta } from '@storybook/react';
+import { CircularProgressBar, CircularProgressBarProps } from '../lib/circular-progress-bar';
+import { Typography } from '@material-ui/core';
+
+export default {
+  title: 'Circular Progress Bar',
```
koonpeng, line 7: Move this to a `Robots` section if `ProgressBar` and `CircularProgressBar` is moved to be next to `RobotInfo`.

packages/react-components/stories/progressbar.stories.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { Story, Meta } from '@storybook/react';
+import { ProgressBar as ProgressBar_ } from '../lib/progressbar';
+import { LinearProgressProps } from '@material-ui/core';
+
+export default {
+  title: 'Progress Bar',
```
koonpeng, line 7: Move this to a `Robots` section if `ProgressBar` and `CircularProgressBar` is moved to be next to `RobotInfo`.

packages/react-components/stories/progressbar.stories.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export const ProgressBar: Story<LinearProgressProps & { value: number }> = (args) => {
+  return <ProgressBar_ {...args} />;
+};
+
+ProgressBar.args = {
+  value: 100,
+};
+
+export const SecondaryProgressBar: Story<LinearProgressProps & { value: number }> = (args) => {
```
koonpeng, line 19: This is probably not needed as the values and colors can easily be changed with storybook settings.

packages/react-components/stories/circular-progress-bar.stories.tsx
```diff
@@ -0,0 +1,37 @@
...
+    </>
+  );
+};
+
+ActiveProgressBar.args = {
+  progress: 70,
+  strokeColor: 'green',
+};
+
+export const StalledProgressBar: Story<CircularProgressBarProps> = (args) => {
```
koonpeng, line 26: This is probably not needed as the values and colors can easily be changed with storybook settings.

packages/react-components/tests/robots/robot-panel.spec.tsx
```diff
@@ -0,0 +1,32 @@
...
+    userEvent.click(root.getByText('test_robot1'));
+    expect(root.getByRole('heading', { name: 'test_robot1' })).toBeTruthy();
+    expect(root.getAllByRole('button', { name: '-' }).length).toBe(5);
+  });
+
+  it('shows detailed information when robot is clicked', () => {
+    const tasks = [makeDefinedTask('Loop', 'test_robot1', 'task_1', 3, 3)];
+    const robots = [makeRandomRobot('test_robot1', 'test_fleet', 2)];
+    const root = render(<RobotPanel robots={robots} fetchTasks={makeFetchTasks(tasks)} />);
+    root.debug();
```
koonpeng, line 28: I think this is leftover from testing?

packages/react-components/tests/test-data/tasks.ts
```diff
@@ -24,3 +25,98 @@ export function makeTask(
     robot_name: 'test_robot',
   });
 }
+
+export function makeDefinedTask(
+  type: string,
+  robot_name: string,
```
koonpeng, line 31: camelCase

matiasbavera:
Hi, @vallq I'll approve it for you to merge it "today", so we can create a branch from it. But can you please address the minor changes pointed out from @koonpeng, test all the new features again, and create a follow-up GH issue of the things that we need to do?

matiasbavera (approved):

packages/react-components/lib/robots/robot-info.tsx
```diff
@@ -0,0 +1,222 @@
...
+  function assignedTasksToStr(robot: VerboseRobot): string {
+    return robot.assigned_tasks
+      .map((task, index) => {
+        if (index != robot.assigned_tasks.length - 1) {
+          return task.task_summary.task_id + ' → ';
+        } else {
+          return task.task_summary.task_id;
+        }
+      })
+      .join('');
```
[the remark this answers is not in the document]
vallq: oh but this means that the last ID will also have an arrow
`1002 → 1004 → `
compared to `1002 → 1004`

PR #341, opened 2021-05-19
feat(api-server): add task querying, sorting and pagination

koonpeng:
## What's new

The get tasks endpoint now supports querying, results are returned by pages. This allows the webapp to scale properly when there are a large amount of tasks, previously it would have to download every task at once. It is also helpful when we add more specialised view, e.g. a robot view that shows the robot's current task (old api would need to download every task and do client side filtering).

image

## Self-checks


packages/api-server/api_server/routes/test_base_route.py
```diff
@@ -0,0 +1,73 @@
+import asyncio
+import concurrent.futures
+
+from ..models import TaskSummary
+from ..models import tortoise_models as ttm
+from .test_fixtures import RouteFixture
+
+
+class TestBaseQuery(RouteFixture):
```
matiasbavera, line 9: You are testing the Pagination as well here, can we add a comment or add that reference to the test name?

packages/dashboard/src/components/tasks/task-page.tsx
```diff
@@ -14,59 +13,58 @@ const useStyles = makeStyles((theme) => ({
...
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        undefined,
+        limit,
+        offset,
+        'priority,-start_time',
+      );
```
matiasbavera, line 42: Just to check is it possible to change `getTasksTasksGet` to an object? Or wrap it? 

e.g: await tasksApi.getTasksTasksGet({limit, offset, options: 'priority,-start_time'}

matiasbavera (approved):
LGTM, left a couple of comments. 

Once this is merged, maybe @vallq will have a few conflicts to solve on her PR.

packages/dashboard/src/components/tasks/task-page.tsx
koonpeng: unfortunately that is generated from swagger codegen, i havent found a way to change the args yet.

packages/api-server/api_server/routes/test_base_route.py
koonpeng: I was intending the "base query" to include pagination and sorting together. It was meant to test the "base query" dependency.

PR #342, opened 2021-05-19
add a timeout of 10minutes

cnboonhan:
Signed-off-by: Boon Han <charayaphan.nakorn.boon.han@gmail.com>

Add a timeout of 10 minutes to the example-deployment deploy script. The default value is 30 seconds, which is not always sufficient for keycloak to launch.

### DIFF
--- a/example-deployment/deploy.sh
+++ b/example-deployment/deploy.sh
@@ -38,7 +38,7 @@ cd $(dirname $0)
 
 kubectl apply -f k8s/keycloak.yaml
 echo 'waiting for keycloak to be ready...'
-kubectl wait --for=condition=available deployment/keycloak
+kubectl wait --for=condition=available deployment/keycloak --timeout=2m
 
 echo 'creating jwt configmap...'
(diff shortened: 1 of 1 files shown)


example-deployment/deploy.sh
```diff
@@ -38,7 +38,7 @@ cd $(dirname $0)
 
 kubectl apply -f k8s/keycloak.yaml
 echo 'waiting for keycloak to be ready...'
-kubectl wait --for=condition=available deployment/keycloak
+kubectl wait --for=condition=available deployment/keycloak --timeout=10m
```
koonpeng, line 41: It is rare for keycloak to take so long to set up, longer timeout means longer wait before if decide keycloak is :boom: . As you mentioned, the a small aws instance takes slightly longer than 30s, so if it is still not up in 2-3min its pretty safe to say that keycloak is :boom:.

cnboonhan:
changed timeout from 10m to 2m

PR #351, opened 2021-05-24
Added aerich migration and configured deploy

### LINKED ISSUE #320: reporting-server: Add a migration tool for Tortoise
## Feature request
Add a migration tool for Tortoise to handle model changes. 

Aerich seems to be a good option `https://github.com/tortoise/aerich`.

matiasbavera:
pre-requisite PR #430 

fixes #320 

## What's new

Tortoise doesn't have a migration tool included but it's one of the midterms objectives. https://tortoise-orm.readthedocs.io/en/latest/roadmap.html#mid-term

* I Added Aerich (from the author of tortoise) as our migration tool. https://github.com/tortoise/aerich
*  Added "Initial" migration with the initial structure of our database (this is generated by aerich).
* I created a k8s job that runs the migration. This job should be executed manually when the `reporting-server-db` pod is up and running.  
* Added migration tests (downgrade from top to bottom and bottom to top)
* Updated documentation of the reporting server
* Add migration test to the pipeline
* Updated deploy script
* Every time we run the `npm run start:dev`, before the server starts it'll check for unapplied migrations and it'll ask you for running those migrations.
* Added and docker build and push GH actions for the `reporting-server` and its migrations. 

image

## Self-checks


example-deployment/docker/reporting-server/reporting-server-base.dockerfile
```diff
@@ -11,6 +11,8 @@ RUN . /opt/rmf/setup.bash && \
 
 FROM rmf-web/builder
 
+COPY --from=0 /root/rmf-web/packages/reporting-server/ /root/reporting-server/
```
koonpeng, line 14: Doesn't this supersedes the next line,
`COPY --from=0 /root/rmf-web/packages/reporting-server/dist/ .`

example-deployment/k8s/jobs.yaml
```diff
@@ -0,0 +1,41 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: reporting-server-migrations-job
+spec:
+  # This is an Alpha feature to remove completed jobs automatically.
```
koonpeng, line 6: Is there any alternative to using this alpha feature?

example-deployment/docker/reporting-server/reporting-server.dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM rmf-web/reporting-server-base
```
koonpeng, line 1: There seems to be a dockerfile at `packages/reporting-server` as well. Would it be possible for the example deployment to use the "official image" as is?

In the case of dashboard, it is not possible because the dashboard requires build time configuration, so  each deployment will need a different dockerfile. For api-server, it is not possible because it depends on rmf, and we haven't had any compatibility guarantee to various rmf versions, even if we have, the release of rmf is still quite slow and we often need in-development features so each deployment will need to build their own rmf.

packages/reporting-server/migrations/models/0_20210524111029_init.sql
```diff
@@ -0,0 +1,90 @@
+-- upgrade --
+CREATE TABLE IF NOT EXISTS "authevents" (
```
koonpeng, line 2: Is this generated or hand coded?

packages/reporting-server/rest_server/app.py
```diff
@@ -51,7 +62,7 @@ def get_app(run_config=SystemMode.ALL):
         app,
         db_url=app_config.db_url,
         modules={"models": ["models"]},
-        generate_schemas=run_config in (SystemMode.ALL, SystemMode.FLUENTD),
+        generate_schemas=False,
```
koonpeng, line 65: I have some confusion when I was working with tortoise-orm. The docs says to not run `generate_schemas` as "part of the application". But in a fresh start without any existing database, we need to run `generate_schemas` to create the tables.

What is your approach to handling table creation? Is it always done by aerich in the pre-startup script?

packages/reporting-server/rest_server/test_orm_config.py
```diff
@@ -0,0 +1,12 @@
+# conflicts with isort because of local non-relative import
+# pylint: disable=wrong-import-order
```
koonpeng, line 2: There doesn't seem to be any imports here

packages/reporting-server/setup.py
```diff
@@ -22,6 +22,7 @@
         "tortoise-orm~=0.16.21",
         "pyjwt[crypto]~=2.0",
         "pydantic~=1.8",
+        "aerich~=0.5.3",
```
koonpeng, line 25: I'm not sure if I got this right, aerich is only ran as a pre-startup script, so technically, the server doesn't really need aerich as part of the application?

Or does the server run aerich during it's startup process?

If it is the former, perhaps aerich should be a dev dependency. On 2nd thought, if we use the latter, would it be possible to do the one container deployment? Since aerich is ran as part of the startup process, the database must be already up. We can either use kubernetes health probe to wait for the db to be started before starting the server pod, or simply have the startup process try connecting to the database several times with a timeout.

example-deployment/k8s/jobs.yaml
matiasbavera: The most common solution is to create a crontab job that cleans up the completed jobs every X hours.

Off topic: We also need to create a crontab job that cleans up the database table if logs are > 7 days.

packages/reporting-server/migrations/models/0_20210524111029_init.sql
matiasbavera: This is generated by aerich, every time that we update a model we should create a new migration with the `aerich migrate` command and then apply it with `aerich upgrade`.

matiasbavera: I'll add a recommender workflow on the Readme

packages/reporting-server/rest_server/app.py
matiasbavera: Yeah, we are now dropping the `generate_schemas` which is a fast built-in migration tool from Tortoise and we are moving completely to aerich to have more control over the migration and versions. So we need to create a database manually and then run the migrations `aerich upgrade` (because we already have the migration sql files)

packages/reporting-server/rest_server/test_orm_config.py
matiasbavera: Sorry this is a leftover

packages/reporting-server/setup.py
matiasbavera: > I'm not sure if I got this right, aerich is only ran as a pre-startup script, so technically, the server doesn't really need aerich as part of the application?

Yes, you're right, just to check I should put this in the `extras_require`? 

```
    extras_require={
        "postgres": ["asyncpg~=0.22.0"],
        "mysql": ["aiomysql~=0.0.21"],
        "maria": ["aiomysql~=0.0.21"],
        "aerich": ["aerich~=0.5.3"],
    }
```

example-deployment/docker/reporting-server/reporting-server.dockerfile
matiasbavera: I'll try to make it work

packages/reporting-server/setup.py
koonpeng: Hmm, I think this is pretty complicated. I understand the dependencies as the packages required to run the application. In this case, since aerich is technically not part of the application, you can `pip install api-server` and run `rmf-api-server` without aerich installed. But however, aerich is needed to do a proper deployment with migrations.

I think if we can build aerich into the application, then this problem can be solved, along with the database and migration startup race condition.

example-deployment/docker/reporting-server/reporting-server.dockerfile
matiasbavera: > There seems to be a dockerfile at `packages/reporting-server` as well. Would it be possible for the example deployment to use the "official image" as is?
So I tested this, and it's working but our "official" image weights 5.14GB so our reporting server ends up having 5.9 gb

example-deployment/k8s/jobs.yaml
matiasbavera: Removed alpha feature.  I didn't find an elegant way to kill a finished task if this task is not found within a cronjob. Since migration does not fall into the category of a cronjob (because it should be run on every deployment and not every certain amount of time), we should run this task as a deploy script. And we should kill that task manually.

matiasbavera: Removed alpha feature. I didn't find an elegant way to kill a finished task if this task is not found within a cronjob. Since migration does not fall into the category of a cronjob (because it should be run on every deployment and not every certain amount of time), we should run this task as a deploy script. And we should kill that task manually.

example-deployment/docker/reporting-server/reporting-server-base.dockerfile
matiasbavera: No, it's copying to another folder, this is used only for migration purposes.

packages/reporting-server/setup.py
matiasbavera: So you mean to run it wvwry time before the reporting-server starts?

matiasbavera:
@koonpeng I'll like your feedback on this =)

packages/reporting-server/README.md
```diff
@@ -39,7 +36,7 @@ RMF_REPORT_REST_SERVER_CONFIG='my_config.py' reporting_server
 * MySQL
 * MariaDB
 
-by default it uses a in-memory sqlite instance, to use other databases, install rmf-server with the relevalent extras
+by default it uses a PostgreSQL instance, to use other databases, install rmf-server with the relevalent extras
```
koonpeng, line 39:
```suggestion
by default it uses a PostgreSQL instance, to use other databases, install reporting-server with the relevalent extras
```

Whats the reason to move from sqlite by default to postgres?

packages/reporting-server/README.md
```diff
@@ -65,8 +62,49 @@ postgres://<user>:<password>@<host>/<database>
 
 for more information, see https://tortoise-orm.readthedocs.io/en/latest/databases.html.
 
+## Run the application
+
+Now that you have already elected, configured and created a database and set you configurations. You need to run the migrations to create the database tables. For that we use `Aerich`:
```
koonpeng, line 67: It sounds like it is not possible to run `reporting-server` without `aerich`, so instead of an extra, it should be part of the core dependencies.

packages/reporting-server/README.md
```diff
@@ -65,8 +62,49 @@ postgres://<user>:<password>@<host>/<database>
...
+```bash
+reporting_server
+
+# Or for server reload
+uvicorn --reload rest_server.app:get_app
+```
 
 # Developers
+## Migration
+We are using [aerich](https://github.com/tortoise/aerich) as our database migration tool. That means that the changes to made to the model will not be reflected in the database automatically. You must run the aerich migration command to create a new migration with the changes:
```
koonpeng, line 92: What happens if we fail to run the migrate command? I'm guessing the db will still be using the old schemas, and we can expect the app to fail. If that is the case, would it be good to automatically run this as part of the app startup? Maybe add a config to enable auto migrate. If it is enabled, the app will perform the migration automatically, if it is disabled, it will try to detect if the schemas are using the correct version, if not, it will exit with a failure immediately.

.github/workflows/e2e.yml
```diff
@@ -1,6 +1,16 @@
...
+   paths:
+    - 'packages/api-client/**'
+    - 'packages/api-server/**'
+    - 'packages/dashboard-e2e/**'
+    - 'packages/dashboard/**'
+    - 'packages/react-components/**'
+    - 'packages/rmf-auth/**'
+    - 'packages/rmf-models/**'
+    - 'packages/ros-translator/**'
+    - 'packages/ros2-bridge/**'
```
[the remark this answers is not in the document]
matiasbavera: 'package/reporting' and 'package/reporting-server/' I think they should have their own e2e test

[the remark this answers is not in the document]
matiasbavera: > `ros2-bridge` is no longer part of this repo.

Removing it

example-deployment/deploy.sh
```diff
@@ -100,4 +112,13 @@ echo 'Applying FluentD configmap ...'
 echo 'deploying FluentD daemonset...'
 .bin/minikube kubectl -- apply -f k8s/fluentd.yaml
 
+# if the migration is finished kill the job
+until kubectl get jobs reporting-server-migrations-job -o jsonpath='{.status.conditions[?(@.type=="Complete")].status}' | grep True ;
```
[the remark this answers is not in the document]
matiasbavera: Yeah, makes sense. Moving this up.

example-deployment/k8s/reporting-server.yaml
```diff
@@ -114,6 +114,10 @@ spec:
           env:
             - name: RMF_REPORT_REST_SERVER_CONFIG
               value: /reporting-server-config/reporting_server_config.py
+          
+          command: ["/bin/sh", "-c"]
+          args: ['exec reporting_server']
```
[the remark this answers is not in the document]
matiasbavera: I added the `reporting_server` as a default command to the docker file. So I already removed this.

example-deployment/k8s/jobs.yaml
```diff
@@ -0,0 +1,33 @@
...
+              name: reporting-server-config
+      volumes:
+        - name: jwt-pub-key
+          configMap:
+            name: jwt-pub-key
+        - name: reporting-server-config
+          configMap:
+            name: reporting-server-config
+
+
```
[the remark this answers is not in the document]
matiasbavera: Removed thanks!

example-deployment/docker/reporting-server.dockerfile
```diff
@@ -11,19 +11,17 @@ RUN . /opt/rmf/setup.bash && \
...
 # cleanup
 RUN rm -rf /var/lib/apt/lists && \
   npm cache clean --force
 
-RUN echo -e '#!/bin/bash\n\
-  exec reporting_server "$@"\n\
-  ' > /docker-entry-point.sh && chmod +x /docker-entry-point.sh
-
-ENTRYPOINT ["/docker-entry-point.sh"]
 CMD ["--help"]
```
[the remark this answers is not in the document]
matiasbavera: I added a default cmd so we can be override it later in `jobs.yaml`

.github/workflows/reporting-server-migration.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Install python dependencies
+        run: pip3 install tortoise-orm pydantic aerich asyncpg 
+      - name: Copy migrations 
+        run: cp -r packages/reporting-server/migrations/ packages/reporting-server/test_migration/migrations/
+
+      - name: Connect to PostgreSQL and run migrations
+        # Runs a script that creates a PostgreSQL table, populates
+        # the table with data, and then retrieves the data.
+        run: cd packages/reporting-server/test_migration/ && aerich upgrade && aerich downgrade --yes -v 0
```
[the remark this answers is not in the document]
matiasbavera: After this test, the app will not work because I'm downgrading all the migrations with this command `aerich downgrade --yes -v 0`. My idea is to have 3 different workflows for this:
* The unit test running on the `reporting-server.yml` file
* `reporting-server-migration.yml` to test if I can upgrade all the migrations from scratch and if I can downgrade them without conflics. This will help us in case of a rollback.
* (In the future) ` reporting-server-e2e` where we will run all the current migrations, run the server and the UI to test the reports. I think this is covering your case.

example-deployment/docker/reporting-server.dockerfile
```diff
@@ -11,19 +11,17 @@ RUN . /opt/rmf/setup.bash && \
...
 
+COPY --from=0 /root/rmf-web/packages/reporting-server/ /root/reporting-server/
+
 COPY --from=0 /root/rmf-web/packages/reporting-server/dist/ .
 
 SHELL ["bash", "-c"]
 RUN . /opt/rmf/setup.bash && \
-  pip3 install $(ls -1 | grep '.*.whl')[postgres]
+  pip3 install $(ls -1 | grep '.*.whl')[postgres] &&\
+  pip3 install $(ls -1 | grep '.*.whl')[aerich]
```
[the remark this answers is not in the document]
matiasbavera: Yeah, makes sense!

packages/reporting-server/README.md
matiasbavera: Yeah, we can do that, having the two options seems more dynamic. Most of the time we will want to have the last migrations. The only case that we don't want to do that is in case we want to test our code for supporting older migrations for a time (here it'll be useful to handle the migrations manually with `aerich` on development).  Maybe this is an edge case but let's say you have 5 reporting-severs instances that point to the same database, but you can update just one instance every hour because we have end-users using the app, so you will have 4 outdated and 1 updated, so you cannot update your database until the 5 instances of the reporting-server are up to date, but your new version should support old APIs and structure as well.

example-deployment/docker/reporting-server.dockerfile
```diff
@@ -11,19 +11,17 @@ RUN . /opt/rmf/setup.bash && \
 
 FROM rmf-web/builder
 
+COPY --from=0 /root/rmf-web/packages/reporting-server/ /root/reporting-server/
```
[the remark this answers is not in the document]
matiasbavera: Ok, I'll try to package all and run the migration script from there.

packages/reporting-server/README.md
matiasbavera: Yeah, moved to core dependencies

packages/reporting-server/README.md
matiasbavera: We can use the persistent SQLite as well, I have no problems at all. 

Let's suppose I want to try a GET request to get data, every time that the system reloads for any change, all the data is wiped out (because it's a DB on memory). So you need to run either a post to fill the DB or a script. With the default Tortoise's migration system, all the tables are created automatically on a fresh DB (every time we reloaded). With aerich, it's different because you need to run a command for the tables to be created, so that's why an in-memory wasn't a good option. Also with the default Tortoise's migration system, I had a lot of problems updating the tables with new fields, and I do not have it anymore with aerich (but I need to create a migration and apply it to the database).

example-deployment/docker/reporting-server.dockerfile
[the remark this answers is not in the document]
matiasbavera: So, I created an `entry_point` `console_script` named `migration` to run it from the `whl` reporting-server python package, but for that to work I needed it to run Aerich embedded to the app but it seems that it cannot be done right now. Aerich doesn't expose a way to actually use it embedded in the app now. Other folks already had that problem and already exists PRs to solve it (check https://github.com/tortoise/aerich/pulls). So either we wait for that feature to be merge in or we maintain our own custom version. 

In the meantime, I split the `reporting-server` from their migrations. We are creating now two containers, one for the app and the other one just for the migrations. This last version it's working correctly and it'll be easy for us to replace it in the future.

.github/workflows/reporting-server-migration.yml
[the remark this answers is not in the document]
koonpeng: Does the downgrade command require both versions of the schema? For example I have a v1 schema and I want to upgrade to v2, aerich generates the .sql file for both v1 and v2. Now if I want to downgrade, do I need the v2 schema? Is it possible for aerich to perform the downgrade with just the v1 schema?

Asking this because I'm wondering how a rollback would work in practice. Suppose we push a bad version to production, the easiest way to rollback is to revert the commit, but doing so will lose the v2 schema.

packages/reporting-server/README.md
koonpeng: In the case of updating the database while one or more instance of the server is still running, I think the server's would definitely need some way to determine the current version of the database? I sounds like it could get very complicated if the database could change schema with every transaction, so I'm think the server would need to somehow lock the database from any schema changes and detect the current schema version before it makes any queries.

In any case, multi schema support must be built into the app, so I think it makes more sense to build schema migrations into it as well. Instead of a boolean, the auto migrate option could be a string or int, which specify the schema version that it should migrate to.

example-deployment/docker/reporting-server/reporting-server.dockerfile
koonpeng: > So I tested this, and it's working but our "official" image weights 5.14GB

Do you know whats filling up all the space? I thought reporting server does not depend on ros or rmf so it should only need python and its dependencies.

packages/reporting-server/setup.py
koonpeng: I think in the worst case, you can spawn a sub-process and wait for it to complete before proceeding with the rest of the init phase.

packages/reporting-server/rest_server/default_config.py
```diff
@@ -3,7 +3,7 @@
     "host": "127.0.0.1",  # ip or hostname to bind the socket to
     "port": 8002,
     "port_fluentd": 8003,
-    "db_url": "sqlite://:memory:",
+    "db_url": "postgres://postgres:postgres@localhost:5432/reporting",
```
koonpeng, line 6: Does this mean that it is no longer possible to `npm start` the server without a running postgres instance?

packages/reporting-server/README.md
```diff
@@ -51,7 +48,7 @@ by default it uses a in-memory sqlite instance, to use other databases, install
 pip3 install reporting-server[postgres]
 ```
 
-Then in your config, set the `db_url` accordingly, the url should be in the form
+Create the database instance on the engine that you chose. Then in your config, set the `db_url` accordingly, the url should be in the form
```
koonpeng, line 51: Now that it is no longer possible to `npm start` without a postgres instance, it might be a good idea to add a section to help people set up the development environment.

example-deployment/docker/reporting-server/reporting-server.dockerfile
matiasbavera: Mmmh... sorry I think I got this msg in the wrong way and I'm confused now. By "official" image I understood: e2e [docker.pkg.github.com/open-rmf/rmf-web/e2e](docker.pkg.github.com/open-rmf/rmf-web/e2e) or even `docker.pkg.github.com/open-rmf/rmf-web/rmf:nightly`. Those weights around 5.14GB. 

Are you saying that I need to create a new official image for the `reporting-server`? and use that for the `reporting-server-migrations.dockerfile`? And use the official `ros:foxy` as our base instead of  `rmf-builder`.

 Sorry for my confusion

.github/workflows/reporting-server-migration.yml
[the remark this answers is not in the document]
matiasbavera: Aerich creates a table on the database where stores all the tables and the migration order. So you need to downgrade them in the correct order.

Answering your questions:
1. You will need V2 to go to V1.
2. You will have to downgrade in two steps. Apply downgrade on the database first and then revert the commit.

packages/reporting-server/README.md
matiasbavera: > In the case of updating the database while one or more instance of the server is still running, I think the server's would definitely need some way to determine the current version of the database? I sounds like it could get very complicated if the database could change schema with every transaction, so I'm think the server would need to somehow lock the database from any schema changes and detect the current schema version before it makes any queries.

But if that the case, and we will block the app until they are on the same version, maybe we should just turn off all the reporting-server instances, upgrade and migrate the app, and turn it on again. Because as you realized it's not an easy path to follow when you're there. 

> In any case, multi schema support must be built into the app, so I think it makes more sense to build schema migrations into it as well. Instead of a boolean, the auto migrate option could be a string or int, which specify the schema version that it should migrate to.

Ok, I'll create a script to do that.

packages/reporting-server/rest_server/default_config.py
matiasbavera: Yes or any other persistent database

packages/reporting-server/README.md
matiasbavera: Makes sense. I'll add that. Thanks

packages/reporting-server/setup.py
matiasbavera: We can do this for development, but it's not the best practice to do it on deployment.  Every instance of the application will attempt to migrate the database. The application has permission to perform destructive updates to the database (this can change in the future). Unless we have only one instance deployed. 

Here is an interesting article related to that https://andrewlock.net/deploying-asp-net-core-applications-to-kubernetes-part-7-running-database-migrations/

example-deployment/docker/reporting-server/reporting-server.dockerfile
koonpeng: Yeah, in fact we might not even need to base it on `ros:foxy`. If the reporting server does not use ros, we could probably base it on `python` or `ubuntu:20.04`, those images would be much smaller.

packages/reporting-server/setup.py
koonpeng: Thanks, thats a great article. I guess kubernetes deployment is still quite new and there isn't an established best practice yet, another web app I worked on had database migrations builtin but I guess it also detects if a migration is ongoing so only 1 instance of the app performs it and the others wait.

There is one part I'm not quite sure on the example, it starts performing the migration while v1 is still running, wouldn't it cause the transactions to fail? Seems like what should happen is that we deploy v2 before the migration, and v2 will have to support both the old and new schema, it will try the new schema first and if it fails it will fallback to the old schema. Then when the migration is complete we can deploy a v3 with the old schema handlers removed.

If we follow this convention (migration in an init container), is it still possible to support in memory db? We can't really migrate an in memory db before starting the app. I guess this also makes it fundamentally different from a traditional app where one would be able to start it by running an executable.

@codebot Would this compromise be acceptable? Or should we make the extra effort to still let it work with in memory db and single executable by implementing some intelligent migrations detection logic into the app?

.github/workflows/reporting-server-migration.yml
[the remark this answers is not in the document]
koonpeng: Thanks, so this makes rolling back more complicated.

packages/reporting-server/README.md
koonpeng: > We can do this for development, but it's not the best practice to do it on deployment. Every instance of the application will attempt to migrate the database. The application has permission to perform destructive updates to the database (this can change in the future). Unless we have only one instance deployed.

> Here is an interesting article related to that https://andrewlock.net/deploying-asp-net-core-applications-to-kubernetes-part-7-running-database-migrations/

Related to this, out-of-process migration looks like the better approach for now, so maybe auto migrate doesn't sound like a good idea? Maybe to work better with the recommended deployment strategy, we need an option to perform migration only and immediately exit. We also need a migrate + run app if we want to support starting with a single executable. Then if we want to go further we need a migration detection logic if running multiple instances in single executable mode, or just put in a doc that multiple instances is not supported in single executable mode.

matiasbavera:
Aerich will always apply all the way to the latest migrations 
image


We cannot upgrade to a specific version, but we can downgrade to a specific version

matiasbavera:
image

example-deployment/deploy.sh
```diff
@@ -88,6 +88,32 @@ kubectl create configmap reporting-server-config --from-file=reporting_server_co
...
+
+# if the migration is finished kill the job
+until kubectl get jobs reporting-server-migrations-job -o jsonpath='{.status.conditions[?(@.type=="Complete")].status}' | grep True ;
+do 
+  echo "wait for migration job to finish"
+  sleep 1; 
+done
+
+echo 'killing reporting-server-migrations-job...'
+.bin/minikube kubectl  -- delete -f k8s/reporting-server-migrations-job.yaml
```
koonpeng, line 115: Why do we need to manually kill the jobs, can't we let it run to completion?

example-deployment/docker/reporting-server.dockerfile
```diff
@@ -1,29 +1,32 @@
-FROM rmf-web/builder
+FROM ubuntu:20.04
+ENV TZ=Asia/Singapore
```
koonpeng, line 2: I'm not sure if we should hardcode this.

---

Can this image be published?

example-deployment/run-reporting-sever-migrations.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
koonpeng, line 1: parts of this seems to be duplicated in `deploy.sh`

packages/reporting-server/README.md
```diff
@@ -88,7 +88,65 @@ postgres://<user>:<password>@<host>/<database>
...
+
+``` bash
+sudo -u postgres createdb <database>
+# E.g: sudo -u postgres createdb reporting
+```
+
+For more information you can check [this tutorial](https://www.digitalocean.com/community/tutorials/how-to-install-postgresql-on-ubuntu-20-04-quickstart)
+
+
+## Run the application
```
koonpeng, line 111: I think this is missing the instructions to set the configs for the db url, user and password.

packages/reporting-server/README.md
```diff
@@ -88,7 +88,65 @@ postgres://<user>:<password>@<host>/<database>
...
+```
+
+If you want to run the migrations and start the server you could also run:
+
+```bash
+npm run start:dev
+```
+
 # Developers
+## Migration
```
koonpeng, line 132: How do you change the settings for the db url, user and password?

packages/reporting-server/README.md
```diff
@@ -98,6 +156,33 @@ for more information, see https://tortoise-orm.readthedocs.io/en/latest/database
 npm run test
 ```
 
+### Running migration tests locally
```
koonpeng, line 159: Would it be possible to add a npm script to run migration tests?

packages/reporting-server/rest_server/__dev__.py
```diff
@@ -0,0 +1,19 @@
+import uvicorn
```
koonpeng, line 1: Why do we need a different startup script for dev?

packages/reporting-server/setup.py
```diff
@@ -31,6 +33,7 @@
     entry_points={
         "console_scripts": [
             "reporting_server=rest_server.__main__:main",
+            "reporting_server_dev=rest_server.__dev__:main",
```
koonpeng, line 36: I don't think we should be exporting the dev scripts.

example-deployment/deploy.sh
matiasbavera: So, this job creates a pod which runs the migrations, when the migrations are applied correctly, the job change to a completed state and the migration pod remains between the pods.  

The job and its pod are only deleted if the task was completed successfully.

To answer your question, sure, we can leave that job and pod as completed and no delete them.

example-deployment/docker/reporting-server.dockerfile
matiasbavera: > I'm not sure if we should hardcode this.

I think I can send it on build time. I need to do this because if not, on built time a prompt appears asking to pick a TZ, which is not ideal.

> Can this image be published?
Do you mean to our docker registration in github?

koonpeng: > Do you mean to our docker registration in github?

I mean make the image in a way that is "deployment agnostic". For example, the hardcoded region info would not make it suitable to be published. If it is possible, then we can move this into the reporting-server package, if we publish it, then the example deployment could download the image instead of building it. If not then the example deployment will refer to the "official" dockerfile to build the image.

example-deployment/deploy.sh
koonpeng: > So, this job creates a pod which runs the migrations, when the migrations are applied correctly, the job change to a completed state and the migration pod remains between the pods.

> The job and its pod are only deleted if the task was completed successfully.

Sorry, I got lost in these. The job and its pod is deleted when it is completed successfully, and when the migration is completed successfully, the task is completed so the job and pods should be deleted?

If I understand it correctly, the job and pods are left in kubernetes when the migration/job fails? Was this by kubernetes design or some side effect of the migration scripts? Maybe kubernetes designed it this way to give admins a chance to check the logs etc?

packages/reporting-server/setup.py
matiasbavera: Yeah, you're right!

packages/reporting-server/rest_server/__dev__.py
matiasbavera: We can use the other one with uvicorn directly.

packages/reporting-server/README.md
matiasbavera: Sure

matiasbavera: I created a script to build the docker container, run it, and apply migrations. 
QQ: I need to wait for Postgres to be up running before running the migrations. 
I tried (all trying to connect from my terminal/bash to docker container):
* There is something called Health in docker, but with this command, I can only check is the container is running but not if Postgres is running. 
* I tried to curl `dockerIp:port`
* pg_isready -d <db_name> -h <host_name> -p <port_number> -U <db_user>                      

What do you think is the best way?

There is an alternative that I did not try yet, and is to make the script inside the container to check is the database is up and running before migration. 

PS: Now it is just with a sleep (it's working but is not ideal.)

packages/reporting-server/README.md
matiasbavera: Thanks for this

example-deployment/deploy.sh
matiasbavera: Oh sorry,


Yeah, this is the default behavior of Kubernetes: If you run a job in Kubernetes, depending on the job a pod is created or not. 

When that job is completed (successfully), that job and its pod are reachable. 
If that job fails, you can configure Kubernetes to retry to run that job (one new job and one new pod).

Those pods and jobs are left on purpose in case you want to see the logs

--------------------------------------------------------------------------------
Our script only deletes the completed jobs and their pod. But we can remove behavior



Is this ok?

packages/reporting-server/README.md
koonpeng: I believe the example deployment has some way to determine if postgres is ready, could we use that here?

example-deployment/deploy.sh
koonpeng: I'm sure kubernetes has good reasons to keep the pods after completion, I don't think we should try to change that behavior. I assume kubernetes is clever enough to clean up these pods eventually so they don't take up all the disk space?

packages/reporting-server/README.md
matiasbavera: Yes, but that Postgres DB is running in a k8s pod. I'll keep digging.

example-deployment/run-reporting-sever-migrations.sh
matiasbavera: Removed this, because it is useless after having the conversations above

example-deployment/docker/reporting-server.dockerfile
matiasbavera: Ok, now I'm using python as our base image. I think that will be "deployment agnostic", I'm adding now a step to push to the docker registry the image of the reporting server.

example-deployment/docker/reporting-server.dockerfile
```diff
@@ -1,29 +1,30 @@
...
   cd packages/reporting-server && \
   npm run prepack
 
-FROM rmf-web/builder
+FROM python:3.8.11-buster
+
+RUN apt-get update && apt-get install -y python3-pip
+RUN pip3 install pipenv
 
 COPY --from=0 /root/rmf-web/packages/reporting-server/dist/ .
```
koonpeng, line 22: I don't think we need this, or we shouldn't have this, `pip3 install` should install whatever is necessary to run `reporting-server`.

packages/reporting-e2e/docker/reporting-server-migrations/Dockerfile
```diff
@@ -0,0 +1,8 @@
+ARG BASE_IMAGE=docker.pkg.github.com/open-rmf/rmf-web/reporting-server
+FROM $BASE_IMAGE
+
+SHELL ["bash", "-c"]
+
+COPY packages/reporting-server/migrations/ /root/migrations/
```
koonpeng, line 6: This shouldn't be needed as `reporting-server` should install whatever is necessary.

packages/reporting-e2e/docker/reporting-server/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM python:3.8.11-buster
```
koonpeng, line 1: This appears to be duplicated in `example-deployment`. Maybe have the example deployment base off this image?

packages/reporting-server/README.md
```diff
@@ -88,8 +88,148 @@ postgres://<user>:<password>@<host>/<database>
...
+
+``` bash
+\q
+```
+
+
+
+### PostgresSQL URL
+
+Once the above steps are finished you can build your bd_url:
```
koonpeng, line 175: typo "bd_url" -> "db_url"

packages/reporting-server/README.md
```diff
@@ -88,8 +88,148 @@ postgres://<user>:<password>@<host>/<database>
...
+
+``` 
+postgres://<user>:<password>@<host>/<database>
```
koonpeng, line 178: Where do I put this url?

packages/reporting-server/README.md
```diff
@@ -88,8 +88,148 @@ postgres://<user>:<password>@<host>/<database>
...
+After running this command your database should have all the tables defined in the code. Now you can run the application
+
+```bash
+reporting_server
+```
+
+If you want to run the migrations and start the server you could also run:
+
+```bash
+npm run start:dev
```
koonpeng, line 201: This uses the default config right? So it wouldn't work if the user sets up the database differently. How can one change the config so that it can work with the way they setup postgres?

packages/reporting-server/README.md
```diff
@@ -88,8 +88,148 @@ postgres://<user>:<password>@<host>/<database>
...
+```
+
+If there are some changes in the Tortoise models, the previous command will create a migration (sql) file that has your changes. To apply these changes to the database you need to run
+
+``` bash
+npm run apply:migrations 
+```
+
+![aerich](https://user-images.githubusercontent.com/11761240/122826198-d97f2e80-d2b0-11eb-813f-384f4ae61d6b.png)
+
```
koonpeng, line 220: How do you run migrations if you installed `reporting-server` from "binary" (`pip3 install reporting-server`)?

packages/reporting-server/README.md
```diff
@@ -88,8 +88,148 @@ postgres://<user>:<password>@<host>/<database>
...
+To change the settings for the url, user and password you can open the [default_config file](https://github.com/open-rmf/rmf-web/blob/main/packages/reporting-server/rest_server/default_config.py) and change the value of the `db_url` with the following format (E.g PostgreSQL):
```
koonpeng, line 221: Changing the source doesn't look like a good way to change the configs, if you installed from `pip3`, then you need to change system files, and it will be overwritten when you update to a new version.

packages/reporting-server/README.md
```diff
@@ -88,8 +88,148 @@ postgres://<user>:<password>@<host>/<database>
...
+
+```bash
+postgres://<user>:<password>@<host>/<database>
+```
+
+You can also add new Tortoise's model folders and other databases on the file `rmf-web/packages/reporting-server/rest_server/aerich_config.py`. You can read more about the configuration of Aerich [here](https://tortoise-orm.readthedocs.io/en/latest/migration.html)
```
koonpeng, line 227: Is there some scenarios where one would need to change the models?

packages/reporting-server/README.md
```diff
@@ -88,8 +88,148 @@ postgres://<user>:<password>@<host>/<database>
...
+
+## PR migration rules
+
+It should only be one migration file per PR unless is strictly necessary to have multiple migrations for clarity.
```
koonpeng, line 231: Would it be better to put this in a "contributing" section (or file)?

packages/reporting-server/README.md
```diff
@@ -98,6 +238,42 @@ for more information, see https://tortoise-orm.readthedocs.io/en/latest/database
 npm run test
 ```
 
+### Running migration tests locally
+
+**Short Version:**  
+To test the migrations on your local machine you can run the following command:
+
+```bash
+npm run test:migrations
```
koonpeng, line 247: Don't know if it is possible, can we use the same commands in CI as well?

koonpeng (changes requested):

packages/reporting-e2e/docker/reporting-server-migrations/Dockerfile
matiasbavera: Ok! So I prepacked the project first and then I copied the `.whl` to the dockerfile, from there I installed the requirements. 7352ab1db7752bf0d53a55daccb51cc98e7acc0c

I just don't want to install Nodejs and Lerna for this pure python-aerich related test. I think I'll take more time and it'll use more resources. But if that is the correct way of doing it, I will. 

Let me know what you think

example-deployment/docker/reporting-server.dockerfile
matiasbavera: Yeah, you're right I removed those.

packages/reporting-server/README.md
matiasbavera: Thanks!

packages/reporting-server/README.md
matiasbavera: Added a contributing file. Added some additional text on the contributing file as well.

packages/reporting-e2e/docker/reporting-server/Dockerfile
matiasbavera: I'll delete the `dockerfile` from the `example-deployment` and I'll point directly to the dockerfile from the e2e-dashboard. So we can have just one. I'll update the documentation and the deploy script as well.

.github/workflows/reporting-server-migration.yml
```diff
@@ -0,0 +1,57 @@
...
+      # Downloads a copy of the code in your repository before running CI tests
+      - name: Check out repository code
+        uses: actions/checkout@v2
+      
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.x'
+
+      - name: Install python dependencies
+        run: pip3 install tortoise-orm pydantic aerich asyncpg
```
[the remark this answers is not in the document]
matiasbavera: Yes, but correct me if I'm wrong, I think we need to have the whl to install the dependencies. So I need to build the project first. To do that I need from `lerna` and `node`. So maybe I can use the reporting-server docker image on the docker registry to get dependencies I need?

packages/reporting-server/README.md
matiasbavera: No, we cannot because in the localhost we are building a container and the ci is the container itself

packages/reporting-server/README.md
matiasbavera: You're right, thanks!

packages/reporting-server/README.md
matiasbavera: Actually, it will work with any database, you just need to have the database connector installed

```
 extras_require={
        "postgres": ["asyncpg~=0.22.0"],
        "mysql": ["aiomysql~=0.0.21"],
        "maria": ["aiomysql~=0.0.21"],
    },
```
We are working directly with the APIs of Aerich and tortoise. So we should change the db_url and install the correct connector

packages/reporting-server/README.md
matiasbavera: We cannot do that. Because we need to have the sql file generated and aerich.ini file to apply the migrations. 

If we copy the migrations folder and if we install the dependencies from binary we can run it.

packages/reporting-server/README.md
matiasbavera: If we refer to the tortoise models that represent database tables, absolutely. We will always want to add attributes, change attribute names, etc. All of this is going to be reflected in migrations.

For example, I just made a table refactor to try to solve the database size issue.

packages/reporting-server/README.md
matiasbavera: Yeah, but we can override that, as we do it in the example-deployment, right?

packages/reporting-server/README.md
koonpeng: Are all these meant to be done by the sysadmin? This seems like a pretty complicated config, might be worth it to have docs showing a situation where you need to modify the tortoise models and explaining how to do it.

packages/reporting-server/README.md
koonpeng: It's possible to do it but there may be some problems

* Someone that uses `pip3 install reporting-server` will have to find directory where the package is installed (this could be a system dir owned by root), go into the source and change the config.
* The default config will be overwritten when the user upgrades the package, so you have to do the patching again.
* For a docker image, you need to mount a volume on the config file. This might not be obvious so might be good to have some docs explaining how to do it and the location of the config file in the image.
  * For kubernetes, you have to use a configmap and mount it in the correct directory, so it might be good to explain this as well.

packages/reporting-server/README.md
koonpeng: I guess the primary way to run this is in a docker container, so it is not designed to be usable with the "old school pip3 install"?

@codebot Is this a limitation a concern? Should we make the effort to ensure that it still works as a "traditional" software?

packages/reporting-e2e/docker/reporting-server-migrations/Dockerfile
koonpeng: referencing this

> I guess the primary way to run this is in a docker container, so it is not designed to be usable with the "old school pip3 install"?

I guess if there is no plans to make this work with `pip3 install`. Then the whole process of building a wheel is unneeded complexity, we can just copy the source into the image.

But on the other hand, if we still want to support `pip3 install`, then we should not be doing any extra copying of sources into the image. We should just run `pip3 install` and everything needed should be installed, i.e. the wheel should contain the migration scripts.

.github/workflows/reporting-server-migration.yml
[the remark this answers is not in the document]
koonpeng: Yeah, as a alternative, we can `pipenv install`, but since we share the same `Pipfile` for all the projects, this would install other unneeded dependencies as well.

If reporting-server is "pip3 installable", then we can install it from source as "editable" e.g. `pip3 install .`, this should bring in only the needed runtime dependencies.

packages/reporting-server/README.md
matiasbavera: Well, that's a valid point, I thought that a developer would be in charge of creating new reports and modify the database, there is a couple of things to consider here
Do we want the sysadmin to modify the database? If yes, most of the frameworks that work with ORMs have this flow + complexity. If the sysadmin wants to modify the database it should have basic knowledge of databases. One example where we want to modify a table would be the following: 

Let's say we just register the start time and the ETA of a task. But after some time we are able to get the real end time of the task. So we want to add the `end_time` attribute to the table. In the following PR, we should add that `end_time` attribute to the `task` model, generate the migration (sql file with the change) and apply the migrations to the database (run the generated sql file with the changes).

If we want to modify the tables the less number possible times I think we should just work with Postgres JSON fields, but it'll be more complicated, we wouldn't be able to use FK, Tortoise filters will be useless because we need to query a JSON field within a table and since the tortoise ORM does not support a JSON query syntax we have to implement the query syntax of each database on the code and that's when we hit that point where the ORM lose it essence, this will be a good scenario for MongoDB but we cannot use it. Other things to consider we empirically saw few time ago, is more expensive to work with JSON fields, the logs occupied more space and queries are more expensive as well due to the number of fields that we manage.

packages/reporting-server/README.md
matiasbavera: Well, I think I depend on this answer to continue working on this feature. I'll switch to my other task in the meanwhile.

packages/reporting-server/README.md
koonpeng: I agree that these config should be managed by the developer, in that case, I think this part could be in a separate developer section/doc, or put in somewhere in the source code as comments/docstrings.

aaronchongth:
Closing in favor of https://github.com/open-rmf/rmf-web/pull/538.

Issue #364, opened 2021-05-28
Submitting loop request did not return the correct number of loops


ChawinTan:
## Bug report

When loop request is submitted with a loop count of  > 1, it will occasionally return a loop count of 1 instead.

### Required information:

- Operating system and version:
  - 
- OpenRMF installation type:
  - 
- OpenRMF version or commit hash
  - 
- ROS distribution and version:
  - 
- ROS installation type:
  - 
- Package or library, if applicable:
  -

koonpeng:
fixed

PR #367, opened 2021-05-31
use onChange instead of onblur

ChawinTan:

## What's new

Related to: #364 

`onBlur` api does not behave the same way in firefox and chrome. Clicking on the submit button triggers the onblur api in chrome but does not in firefox. 

In order to be more consistent, the number of loops would be set during `onChange` instead.

## Self-checks


koonpeng (approved):
thanks for the fix.

PR #348, opened 2021-05-24
api-server: get robots

koonpeng:
## What's new

Adds new endpoints/more useful data in existing fleets endpoint

* `/fleets` now supports querying, the returned data now includes the current state.
* new `/fleets/robots` endpoint, the endpoint supports querying, the return data includes robot's fleet, robot state, and currently active, pending and queued tasks assigned to the robot.

## Self-checks

- [ ] I'm familiar with and follow this [Typescript guideline](https://basarat.gitbook.io/typescript/styleguide)
- [ ] I added unit-tests for new components
- [ ] I tried testing edge cases
- [ ] I tested the behavior of the components that interact with the backend, with an e2e test

PR #369, opened 2021-05-31
dashboard: add cancel task

koonpeng:
## What's new

image

Adds a button that sends a post request to the cancel task endpoint.

## Self-checks


Yadunund (commented):
Tested with hardware and the task cancellation works as expected! A couple of thing to highlight:
1. The State of the task does not get updated to "Cancelled". It remains "Queued" 
2. Tasks that have already been cancelled should have their "Cancel Task" buttons inactive.
3. The sequence of tasks in the Robots tab does not update to reflect the reassignment of tasks after the cancellation (not sure if this is out of the scope of this PR, cc: @ChawinTan )

tanyouliang95:
Thanks for this! For my side, I have 2 comments:
1) In regards to the UI pop-up notification. 
It will show "Successfully canceled task' even if the `rmf_dispatcher` returns a failed response. A failed cancel req happens when we are trying to cancel an active task/charging task. Is it possible to let the api-server return the response from the [cancel srv](https://github.com/open-rmf/rmf_internal_msgs/blob/main/rmf_task_msgs/srv/CancelTask.srv)? If due to potential async latency, maybe we can change the notification text to "Sent Cancel Task Request'?

2) Task assignment queue
@ChawinTan  This is not so relevant to cancel task. But to follow up on yadu's comment. Other than reverse queue, I notice that one is missing from the assignment here:
Screenshot from 2021-05-31 17-17-26

This is the "ground truth":
Screenshot from 2021-05-31 17-17-14

It seems like because the `Clean0` task is located at page 2, so not taken into account. 
Screenshot from 2021-05-31 17-17-04


Anyway hope these feedback helps. Thanks a lot for the help.

ChawinTan:
Noted! Will take a look at the issues pointed out. For the point @tanyouliang95 pointed out, that might be due to pagination.

koonpeng:
> The State of the task does not get updated to "Cancelled". It remains "Queued"

It might be because of wrong sim time/real time settings either in rmf or in the api-server.

koonpeng:
After checking ros2 topic, it's probably because rmf never published the new state to `task_summaries`.

koonpeng:
> Noted! Will take a look at the issues pointed out. For the point @tanyouliang95 pointed out, that might be due to pagination.

With #348 merged, you might want to try using the new `/fleets/robots` endpoint to get a robot's tasks, or use the improved `/tasks` endpoints to filter the tasks by fleet, robot and task state. That would solve the problem of downloading a large list of task and doing client side filtering.

Yadunund (approved):
Thanks for all the changes! In my testing, I changed the api server to subscribe to `/filtered_task_summaries` and everything worked as expected.

@koonpeng feel free to merge if there are no other changes from your side.

PR #371, opened 2021-06-01
Fix/robot panel pagination

ChawinTan:
## What's new

- Refactor how tasks are fetched
- Solve pagination issue when tasks are fetched
- Remove the reversing of the task order as the priority filter was passed in descending order. Corrected that to ascending order.

## Self-checks


ChawinTan:
@Yadunund I just did a refactor of the robot panel from the original design now that I understand the moving parts much better. 
- Removed the setIntervals and shifted the handling of robots with tasks to the `Robot Panel` instead of `Robot Table`
- The tasks order should also be fixed. It's probably due to a git merge that is suppose to conflict, but for some reason, git merged the changes without giving me conflicts locally. 
- The other issue regarding the robot destination, location etc .... should also be fixed

Yadunund:
Just tested it but I'm still seeing a discrepancy in the order of the tasks. The order seemed to be correct in yesterday's last commit 🤔 
Here is a screenshot of the assignments 
image

This is what the assignments should be
image

Page 1 of Tasks
image

Page 2 of Tasks
image

Page 3 of Tasks
image

Page 4 of Tasks
image

PR #375, opened 2021-06-02
fix(dashboard): fix missing waypoints

### LINKED ISSUE #373: Only named waypoints from nav_graphs[0] visualized
### Required information:

- Operating system and version:
  - Ubuntu 20.04
- OpenRMF installation type:
  - Source 
- OpenRMF version or commit hash
  - `main`
- ROS distribution and version:
  - Foxy
- ROS installation type:
  - Binaries
- Package or library, if applicable:
  - `rmf-web`

### Description of the bug

For a given level, the dashboard only renders the named waypoints in the first nav_graph in `BuildingMap.levels.nav_graphs`.
I think the problem lies here https://github.com/open-rmf/rmf-web/blob/90a3930b1f8ec9dedb54885b37536483a04f57d0/packages/dashboard/src/components/schedule-visualizer/waypoints-overlay.tsx#L26

### Expected behavior

Named waypoints from all the `nav_graphs` should render on the canvas

### Actual behavior

Only waypoints from `nav_graphs[0]` gets rendered.

koonpeng:
## What's new

fixes #373
Waypoints now use the `PlacesContext` which should contain data from all nav graphs.

## Self-checks

### DIFF
--- a/packages/dashboard/package.json
+++ b/packages/dashboard/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "scripts": {
     "start": "concurrently npm:start:rmf-server npm:start:rmf npm:start:react",
+    "start:clinic": "RMF_DASHBOARD_DEMO_MAP=clinic.launch.xml npm start",
     "start:react": "react-scripts start",
     "start:rmf": "ts-node -P scripts/tsconfig.json -T scripts/start-rmf.ts",
     "start:rmf-server": "RMF_SERVER_USE_SIM_TIME=true npm --prefix ../api-server start",
--- a/packages/dashboard/rmf-launcher.ts
+++ b/packages/dashboard/rmf-launcher.ts
@@ -77,12 +77,13 @@ export class LocalLauncher {
     }
 
     const headless = !process.env.RMF_DASHBOARD_NO_HEADLESS;
-    const officeDemoArgs = ['launch', 'rmf_demos', 'office.launch.xml'];
+    const demoMap = process.env.RMF_DASHBOARD_DEMO_MAP || 'office.launch.xml';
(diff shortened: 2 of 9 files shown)


Yadunund:
Seems to work for single level maps. But things break down for multilevel ones. 
* Switching between levels shows the same collection of waypoints that seem to be a mix from the different levels.
* There could be waypoints with the same name but present in different nav graphs. In such cases it looks like only one of the waypoints is rendered. 

You can test on the `clinic` world
```ros2 launch rmf_demos clinic.launch.xml```

koonpeng:
> There could be waypoints with the same name but present in different nav graphs. In such cases it looks like only one of the waypoints is rendered.

How can waypoints be uniquely identified? If 2 graphs has the same vertex name, but different location and params, are they considered to be different waypoints? If so, then waypoint names are not unique, neither is their location nor params, when presenting such group of waypoints in forms like the create task form, there will be multiple options with the same name, how can we differentiate between them?

Yadunund:
> How can waypoints be uniquely identified?

Each fleet has a `rmf_traffic::agv::Graph` that is built from an element in `levels.nav_graphs`. In `rmf_traffic::agv::Graph`, waypoints are uniquely identified by their index. Two waypoints in the same Graph cannot have the same name.


> If 2 graphs has the same vertex name, but different location and params, are they considered to be different waypoints?

Yes

> If so, then waypoint names are not unique, neither is their location nor params, when presenting such group of waypoints in forms like the create task form, there will be multiple options with the same name, how can we differentiate between them?

This is a valid concern. I think it should be the responsibility of the system integrator to set up the waypoints and navigations graphs such that there is no ambiguity

Yadunund (approved):
Works with multiple levels now 👍🏼

PR #378, opened 2021-06-02
more rows and sort cancelled task

ChawinTan:

## What's new

- increase task rows from 10 to 20
- shift cancelled task to the back

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/tasks/task-page.tsx
+++ b/packages/dashboard/src/components/tasks/task-page.tsx
@@ -41,7 +41,7 @@ export function TaskPage() {
         undefined,
         limit,
         offset,
-        '-priority,-start_time',
+        'state,-priority,-start_time',
       );
       const taskProgresses: TaskProgress[] = resp.data.items;
       const task_summaries = taskProgresses.map((t) => t.task_summary);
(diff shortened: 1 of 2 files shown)


tanyouliang95:
Hi @ChawinTan Thanks for the fast fix. Should we merge this into `main` instead since this will also be common non-blocking feature? (if you think this is suitable)

Also other than shifting `Canceled`, can we also pushback `Failed` and `Completed` tasks?

koonpeng:
Note that a row count of 10 is chosen because one of the design goal of the wireframe is to minimize scrolling. I think a row count of 20 would almost always end up with a scroll bar.

The sorting is also part of the design, to sort by priority first then start time.

Let's consult with @vallq and @codebot to make sure that this is the direction we want to go.

ChawinTan:
My thoughts were that this feature is for debugging purposes for the demo only, so I think we should keep it to the throwaway/demo branch.

koonpeng:
> My thoughts were that this feature is for debugging purposes for the demo only, so I think we should keep it to the throwaway/demo branch.

Yeah, but I'm not sure about demoing changes that are in conflict with the design decisions we have. We would be demoing something that is not representation of the actual product we are "selling". Changes like these should go directly to the main branch, after we agree that these changes are acceptable to the design decisions we had.

tanyouliang95 (approved):
Yes understand. Since these design decisions are not finalized, we will only merge it into `throwaway/demo, and we will review this as time comes.

Anyway, tested it and it works as described. Thanks a lot!!

Yadunund:
I'm on the latest `throwaway/demo` but the tasks are not sorted 🤔 
image

ChawinTan:
My changes got removed probably during merging. Will make a pr now.

PR #383, opened 2021-06-02
Dashboard: Fix dashboard does not get icons from kubernetes cluster

matiasbavera:

1. run the `example-deployment/get-icon.sh` script to populate the  ws/rmf-web/packages/dashboard` with the icons.
2. build the dashboard's docker.

## Self-checks


cnboonhan:
thank you for your help :bow: 

i think there might be a few differences in how we are using rmf-web and the script,

could it be possible to use environment variable to specify paths for the `rmf-web` workspace? we put rmf-web in $HOME, not in `ws`.

In addition, we are using custom icons not available online, how do we modify the .resources file accordingly?

koonpeng:
I believe the example deployment git clones the rmf sources as part of the instructions, so you can point to script to the rmf directories via relative paths.

> i think there might be a few differences in how we are using rmf-web and the script,
> could it be possible to use environment variable to specify paths for the rmf-web workspace? we put rmf-web in $HOME, not in ws.
> In addition, we are using custom icons not available online, how do we modify the .resources file accordingly?

In that case, I don't think you should be using the example scripts, you can run the setup from the dashboard before building it as part of your deployment script.

matiasbavera:
> I believe the example deployment git clones the rmf sources as part of the instructions, so you can point to script to the rmf directories via relative paths.

You mean by using the js "ament"? 

> In that case, I don't think you should be using the example scripts, you can run the setup from the dashboard before building it as part of your deployment script.

Yeah, sounds good. So there is no need for the get-icons-locally script then. So the deploy script should show the resources prompt then? 

I'll create a new issue with this and let's merge just the icon fix then

koonpeng (approved):
If this is just the icon fix, I think it can go into the main branch.

PR #384, opened 2021-06-03
perf(dashboad): fix some performance regression in schedule visualizer

koonpeng:
## What's new

Many markers are being re-rendered needlessly, testing on the clinic world, these optimizations reduces the render times from ~10ms to ~2ms.

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/schedule-visualizer/robots-overlay.tsx
+++ b/packages/dashboard/src/components/schedule-visualizer/robots-overlay.tsx
@@ -1,7 +1,7 @@
-import * as RmfModels from 'rmf-models';
 import Debug from 'debug';
 import React, { useMemo } from 'react';
 import { RobotMarker as RobotMarker_, RobotMarkerProps } from 'react-components';
+import * as RmfModels from 'rmf-models';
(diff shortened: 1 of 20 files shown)


packages/dashboard/src/components/schedule-visualizer/robots-overlay.tsx
```diff
@@ -80,12 +75,17 @@ export const RobotsOverlay = (props: RobotsOverlayProps) => {
           robots.map((robot) => (
             <MarkerComponent
               key={robot.name}
-              robot={robot}
+              name={robot.name}
+              model={robot.model}
+              // Round to 2 decmial places (1cm) to prevent extra rendering when robot is idle
```
matiasbavera, line 80: nit: decimal*

matiasbavera (approved):
LGTM. I see it was the properties sent it as objects.

koonpeng:
> LGTM. I see it was the properties sent it as objects.

Yeah, and also some sneaky bits that uses fix length arrays like `[x, y]` to represent 2D points.

PR #389, opened 2021-06-04
fix(dashboard): only allow priority and number of loops to be >= 0

### LINKED ISSUE #386: Create Task: Priority and loop count inputs should always be positive integer
image

Currently the input allows negative integers for these fields

koonpeng:
## What's new

fixes #386 

## Self-checks


Yadunund (approved):
Thanks for the fix.

Issue #359, opened 2021-05-27
Storybook freezes when clicking Circular Progress Bar


vallq:
Storybook freezes when clicking on this component Circular Progress Bar
Related (but not entirely helpful): https://github.com/storybookjs/storybook/issues/1252
This only happened after moving the component into the robot folder and updating the imports.

File in question: `react-components/stories/robots/circular-progress-bar.stories.tsx`

Behaviour:
https://user-images.githubusercontent.com/28668944/119798466-be1d3f80-bf0d-11eb-999b-98f99b938bec.mp4

Does not impact the dashboard, only the react-component package

koonpeng:
fixed by #391

PR #391, opened 2021-06-04
Fix/storybook errors

ChawinTan:
## What's new

related to: #359 

- fix storybook freezing when clicking circular progress bar due to using the wrong import value `CircularProgressBar` instead of `CircularProgressBar_`. Storybook component was also using the same value as the original import, hence the confusion probably causes it to freeze
- fix header-bar stories `no TabContext` errors as material-labs `TaskPanel` api requires it be wrapped around a `TabContext`. Refer to the [this](https://material-ui.com/components/tabs/) docs under `experimental api`
- fix system notifications date time conversion errors due to migration to `date-fns` module

## Self-checks


koonpeng (approved):
Do you mind explaining the cause of the bug and why this fixes it?

ChawinTan:
@koonpeng updated the comments

Issue #393, opened 2021-06-05
End Time is an integer


Yadunund:
`rmf-web` : `main` branch

The `End Time` for a task as populated in the Robots tab seems to be an integer. It is however correctly populated in the card on the right side panel.

![image](https://user-images.githubusercontent.com/13482049/120888316-40061a80-c62a-11eb-8d40-108b6f16ec4f.png)

ChawinTan:
I think this is to represent the duration needed for the tasks since the value is `end_time - start_time`. If that's the case then changing the title to something more appropriate would do. Maybe @vallq can clarify the original intention?

vallq:
hi sorry, I missed this. but the original design was for this to be an ETA (duration) instead of End Time. The integer right now is an estimate in seconds with the start and end time

koonpeng:
fixed by #394

PR #394, opened 2021-06-07
change end time to duration

ChawinTan:
(cherry picked from commit 2c0589d81be297c57ffeca0bde0e99c39f09d87b)

## What's new

Related to #393

Change the `End Time` field of the robot table to `Duration` to better reflect what is displayed on the column.

## Self-checks

### DIFF
--- a/packages/react-components/lib/robots/robot-table.tsx
+++ b/packages/react-components/lib/robots/robot-table.tsx
@@ -160,7 +160,7 @@ export function RobotTable({
               <TableCell>Robot Name</TableCell>
               <TableCell>Start Location</TableCell>
               <TableCell>Destination</TableCell>
-              <TableCell>End Time</TableCell>
+              <TableCell>Duration</TableCell>
               <TableCell>Battery</TableCell>
               <TableCell>State</TableCell>
             </TableRow>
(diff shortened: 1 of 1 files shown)


packages/react-components/lib/robots/robot-table.tsx
```diff
@@ -160,7 +160,7 @@ export function RobotTable({
               <TableCell>Robot Name</TableCell>
               <TableCell>Start Location</TableCell>
               <TableCell>Destination</TableCell>
-              <TableCell>End Time</TableCell>
+              <TableCell>Duration</TableCell>
```
Yadunund, line 163: Is the duration in seconds? If so we should make the units clear. Maybe `Duration (s)`? Or we could leave the column header as suggested and append the `s` or `seconds` after the integer data.

Yadunund (changes requested):

packages/react-components/lib/robots/robot-table.tsx
ChawinTan: Added an `s` after the data to be consistent with the existing format of other columns

koonpeng (approved):

Yadunund:
After second thought `Duration` may be ambiguous as well. People might think it is the duration for which the robot has been active or the remaining operational duration before the battery dies. In the current implementation, the `Duration` is the estimated duration of the active task only. Perhaps a header that makes this apparent is better.

In a separate discussion, we can explore the wireframe again to find more useful data to populate this table. Maybe the task id of the active task could be an addition/replacement.

Yadunund (approved):

PR #396, opened 2021-06-08
feat(dashboard): auto refresh tasks in task page

koonpeng:
## What's new

This adds a button on the task page to toggle auto refresh. "Auto refresh" may different meaning to different people, this PR implements auto refresh like this:

* Currently shown tasks on the task page will be listened for new task updates.
* Auto refresh never changes the ordering or the list of tasks currently shown, only currently shown tasks will be updated.
* Clicking the refresh button will fetch a new set of tasks from the server using the current filter/page settings, this may change the task list and/or the ordering of the tasks.

This means that when a new task is created, it won't appear on the task page without a manual refresh, even if auto refresh is turned on. However, when a task state or status changes, e.g. from `ACTIVE` to `COMPLETED` it will be automatically refreshed if auto refresh is turned on and the task is currently shown on the active page.

Other things that this PR adds:

* Wait for the task summary to be written to database before returning from submit task endpoint. This ensures that subsequent calls to `/tasks` will contain the new task.
* Add a small margin after each task phases row to make each task more easily distinguished.
* Add unsubscribe feature in the server.
* Add task summary endpoints in socket.io.

## Self-checks

### DIFF
--- a/packages/api-client/lib/index.ts
+++ b/packages/api-client/lib/index.ts
@@ -14,7 +14,9 @@ import {
 } from 'rmf-models';
 import { io, Socket } from 'socket.io-client';
 
-export type Listener<T = unknown> = (resp: T) => void;
+// https://stackoverflow.com/questions/52667959/what-is-the-purpose-of-bivariancehack-in-typescript-types
+export type Listener<T = unknown> = { bivarianceHack(resp: T): void }['bivarianceHack'];
+export type Subscription = Listener;
 
 export class SioClient {
   public sio: Socket;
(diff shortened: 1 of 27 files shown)


Yadunund (approved):
Thanks for this feature, KP. It works great.
I notice that it takes 3-5 seconds for the Create Task window to close after i click Submit. Previously, it was instantaneous. Is this due to the wait that you introduced?

koonpeng:
> I notice that it takes 3-5 seconds for the Create Task window to close after i click Submit. Previously, it was instantaneous. Is this due to the wait that you introduced?

Yeah, it waits until the task is committed into the database before returning.

Issue #395, opened 2021-06-08
Only view tasks related to the user


chartrmf:
### Description

When a user accesses her RMF Web dashboard with her credentials, the user should only be looking at 'Tasks' she has created and has been given access to via her role and department.

_E.g. a staff from the hospital's A&E department should only look at tasks she created, whilst her manager should be looking at tasks her department has created._

This feature would allow users using RMF Web in a larger setting to focus only on tasks that matter to them.

### Implementation considerations

Each user could be tagged with a department and a role, and this would then be tagged to the task she creates.

PR #397, opened 2021-06-09
feat: authorization system

koonpeng:
# Proposal

## OpenID Connect

rmf-server does not manage user identities and access levels by itself, it uses an OpenID Connect compatibily authentication server. Therefore, in order to management an user's access level and roles, you should change it from the authentication server, there are no endpoints for account management in rmf-server for now. If using keycloak as the authentication server, see https://www.keycloak.org/docs/latest/server_admin/index.html#roles for detailed documentation.

### Format

OpenID Connect does not define the format of the access token, the canonical way for a resource server to validate an access token is to use the token introspection endpoint of the authentication server. In order to not have to connect to the authentication server for every request, rmf-server assumes the convention that the access token is a JWT that can be verified independently. Most modern authentication server and services like keycloak and auth0 follows this convention.

### Claims

OpenID Connect does not specify a conventional way to obtain an user's access levels or roles. It does however allows arbitrary claims to be defined for an user, to that end, rmf-server makes use of the `resource_access` claim defined by default in keycloak to obtain an user's roles. The claim should be in the form

```json
"resource_access": {
  "${cliend_id}": {
    "roles": [
      ...
    ]
  }
}
```

where `client_id` refers to the client id created in keycloak. This claim should be attached to the access token. If using other authentication servers, you need to config it such that it returns the same claim.

## Roles

While authentication and managing user access is done in the authentication server, authorization is done in-app by rmf-server. With the correct access token, rmf-server can securely find out about an user's roles, below are the list of builtin roles that rmf-server understands:

| Role | Description |
| - | - |
| _rmf_task_submit | Submit task |
| _rmf_task_cancel | Cancel an existing task |
| _rmf_task_admin | Able to manage tasks created by anyone |
| _rmf_superadmin | Superset of all roles |

On top of the builtin roles, users can be assigned to any roles that begins with `rmf_`, rmf-server will attach these roles to any resources they create, any users with any of the roles will have read access to these resources. Each resource also has an owner, the owner will always have full control over the resource.

For example, given an user alice, which has the roles `["rmf_kitchen", "_rmf_task_submit"]` and bob which has the roles `["rmf_kitchen"]`. Alice will be able to submit a new task, since alice is the owner of the task, she will be able to cancel it even though she doesn't have the `_rmf_task_cancel` role. Bob on the other hand will not be able to cancel the task unless he has the `_rmf_task_cancel` role. Bob also cannot submit new tasks because he does not have the `_rmf_task_submit` role, however, he will be able to see the tasks created by alice because he has the same `rmf_kitchen` role as alice.

Assume a third user, charlie who has the roles `["_rmf_task_cancel", "rmf_bedroom"]`, he will not be able to see or cancel the task created by alice because he doesn't have the `rmf_kitchen` role. However, if alice also has the `rmf_bedroom` role, then charlie will be able to see and cancel the task. Similarly, if charlie has the roles `["_rmf_task_admin"]`, he will be able to see the task created by alice, but will not be able to cancel it unless he also has `_rmf_task_cancel`.

## Self-checks


koonpeng:
> I have some trivial bike-shed suggestions about the naming conventions. Using a leading underscore _ to indicate the built-in topics seems like it will be challenging for people to remember. Do we need to have any convention at all? In other words, do we need rmf_ and _rmf_ prefixes on the RMF roles, or can we just use strings without prefixes? I assume that Keycloak would throw an error if we had a pre-defined role called task_submit and a super-admin tried to create a new role called task_submit ?

One reason I had for this is to avoid accidental role assignment from other keycloak clients. The other reason for an `_` prefix is to differentiate builtin roles and user defined roles. This proposal uses them slightly differently, builtin roles are used to check if an user can perform certain actions, while the user defined roles are used to check if a user can read certain resources. In the code, I actually separated them to roles and groups by using the `_`, but keycloak doesn't have a `groups` claim by default. Should we make the separation at keycloak, so that it returns an access token with `roles` and `groups` as opposed to differentiating them by `_` in the app?

> If I'm understanding the current list of pre-defined roles correctly, it seems there is an "implicit" ability to view robots and tasks that is always provided ? Which is fine, I'm just trying to understand.

This only covers tasks for now because that is what the feature request is about, I haven't thought about the set of roles for other resources. The visibility of tasks is determined by an user's user defined role (or groups), an user can only see tasks which are in the same group. Since rmf internal tasks and tasks created via "debug" interfaces like `ros2 topic pub` or rviz, does not have a owner/creator, so only superadmin and task admins can see them.

> I think that I didn't fully understand the user-defined roles and how they would partition the users into who can view what task. Are the user-defined roles relating to any spatial areas, or are they instead just grouping users together, so they could be called group_A or group_B or whatever?

I think that would be up to the admin on how they want to define roles. In the chart's feature request for example, the admin could define an `A&E` (maybe not the `&` character) group and assign users that should be avoid to view `A&E` tasks into that group. But one limitation is that rmf does not have a notion of task authorization group, so only task submitted through the ui will have groups attached to them, by default, the groups attached to a task is the same set of group the submitter belongs to, there isn't any functionality to group tasks accordingly to its contents.

matiasbavera:
> ## Roles
> While authentication and managing user access is done in the authentication server, authorization is done in-app by rmf-server. With the correct access token, rmf-server can securely find out about an user's roles, below are the list of builtin roles that rmf-server understands:
> 
> Role	Description
> _rmf_task_submit	Submit task
> _rmf_task_cancel	Cancel an existing task
> _rmf_task_admin	Able to manage tasks created by anyone
> _rmf_superadmin	Superset of all roles
> On top of the builtin roles, users can be assigned to any roles that begins with `rmf_`, rmf-server will attach these roles to any resources they create, any users with any of the roles will have read access to these resources. Each resource also has an owner, the owner will always have full control over the resource.
> 
> For example, given an user alice, which has the roles `["rmf_kitchen", "_rmf_task_submit"]` and bob which has the roles `["rmf_kitchen"]`. Alice will be able to submit a new task, since alice is the owner of the task, she will be able to cancel it even though she doesn't have the `_rmf_task_cancel` role. Bob on the other hand will not be able to cancel the task unless he has the `_rmf_task_cancel` role. Bob also cannot submit new tasks because he does not have the `_rmf_task_submit` role, however, he will be able to see the tasks created by alice because he has the same `rmf_kitchen` role as alice.
> 
> Assume a third user, charlie who has the roles `["_rmf_task_cancel", "rmf_bedroom"]`, he will not be able to see or cancel the task created by alice because he doesn't have the `rmf_kitchen` role. However, if alice also has the `rmf_bedroom` role, then charlie will be able to see and cancel the task. Similarly, if charlie has the roles `["_rmf_task_admin"]`, he will be able to see the task created by alice, but will not be able to cancel it unless he also has `_rmf_task_cancel`.

So from my understanding, and please correct me if I'm wrong we are using _Role-based access control (RBAC)_ to attack the task authorization problem. Those tasks that are used as an example look like action granular to me (submit, cancel, etc). If we want to have authorization per action maybe we want to have permissions per actions + roles instead of just roles (the good thing about just having roles is that it makes things easier to handle the authorization, the downside is that we lose flexibility or granularity per actions).  A related question to this, are we going to use permissions/roles to verify what to render on the UI side of things?

koonpeng:
@matiasbavera thanks, I think that is a good suggestion, it will allow us to support a large number of permissions without bloating the access token.

Do you think it is enough to have a fixed set of roles, or should we allow an admin to freely define new roles? I don't think there is a concept of permissions in keycloak so we would have to make an ui in the dashboard, would it be confusing for the user if they need to create roles in the dashboard but then assign those roles to users in the auth provider? Should we go a step further and manage user roles assignment from within dashboard? That would greatly increase coupling between the api server and keycloak though, making it a lot harder to use a different auth provider.

koonpeng:
> and the "log" panel should not show up for users who are not authorized to view logs.

There isn't a log panel, the log is in a separate webapp. So the reporting app should only allow users with the correct roles.

matiasbavera:
> @matiasbavera thanks, I think that is a good suggestion, it will allow us to support a large number of permissions without bloating the access token.
> 
> Do you think it is enough to have a fixed set of roles, or should we allow an admin to freely define new roles? I don't think there is a concept of permissions in keycloak so we would have to make an ui in the dashboard, would it be confusing for the user if they need to create roles in the dashboard but then assign those roles to users in the auth provider? Should we go a step further and manage user roles assignment from within dashboard? That would greatly increase coupling between the api server and keycloak though, making it a lot harder to use a different auth provider.

I got this from the new Keycloack documentation; 
https://www.keycloak.org/docs/latest/authorization_services/#_permission_overview
https://www.keycloak.org/docs/latest/authorization_services/#permission

Mmm.. I get what you are saying. When we have a better understanding of how the permissions work on Keycloak (how to assign to a role, if they are unique, how the API works, etc), I think the end-user will be able to create new roles directly on Keycloak and assign those permissions on Keycloak, because if we handle everything with permissions, we will not care about the role anymore (role will be just a group of permissions). Yeah, if we want the end-user to define a new role.

If there is actually no way to use those permissions in Keycloak. I agree with you in your comment on creating a UI to create roles and assign permissions and increase the coupling with auth provider. Some philosophical questions/comments regarding that: 
* Isn't that the long-term goal? In other words, end-users do not have to enter the Keycloak console and they can manage everything from the dashboard (?)
* If we/someone else changes the auth provider that mentioned UI would be a facade above the implementation. So if we change the auth provider it would be transparent for the end-user, but definitively not for us, would have to change the structure and the endpoints underneath.

matiasbavera:
I think this PR is a pre-requisite for https://github.com/open-rmf/rmf-web/issues/395

koonpeng:
> I got this from the new Keycloack documentation;
https://www.keycloak.org/docs/latest/authorization_services/#_permission_overview
https://www.keycloak.org/docs/latest/authorization_services/#permission

I wasn't able to get this working for public clients, the option to enable it only appears for private clients.

> Isn't that the long-term goal? In other words, end-users do not have to enter the Keycloak console and they can manage everything from the dashboard (?)

I thought part of the goal is to support multiple auth providers, should we still pursue that flexibility, or just have a hard dependency on keycloak?

koonpeng:
Keycloak itself is quite flexible, it can act as a middleman to various other providers so even if we put a hard dependency on keycloak it is still possible to be plugged in to many enterprise systems. The features available may be limited however, I don't know if it is possible for keycloak to convert and forward admin functions like creating users/groups, assigning roles etc to the backend provider. We may end up with a hybrid approach whereby keycloak only sync to the backend provider for certain functions it supports, and maintains its own database for others.

In some cases, we might also need to write keycloak plugins to support backend providers not supported by default. I don't know how hard it is to write those plugins, but at the very least keycloak is a java app so we would need to learn the java language and toolchains.

If we don't put a dependency on keycloak, then we need to write an adapter for each provider. It's probably not too hard to do it but I think the main challenge comes from maintaining those adapters, we need to continuously test them, and with different versions and/or configs of those providers to ensure they continue to work.

> to make an "ultra-minimalist auth" system for CI and local-development purposes that is as simple as possible to set up, ideally with a single shell command, etc.

The current system is to skip authentication for these situations, but extending it to user management will be quite alot more work. We need to add operations to create, read, update and delete users, groups, roles, permissions etc. Theres probably at least 16+ rest endpoints we would have to implement, that also goes for the adapters, it will be at least 16 functions that it needs to implement as well. Most of those functions will probably be quite simple, so I think the main challenge is testing.

koonpeng:
superseded by #442

PR #400, opened 2021-06-14
Reporting server/task

vallq:
## What's new
Addition of the task summary model to the reporting server and UI.

## Self-checks

## Discussion
Having some problems with testing on minikube:
The UI seems to still call the old endpoint (/report/task_state) instead of the new endpoint (/report/task_summary) even though it has been updated on the deployment workspace (ws/rmf), (I rebuilt the images for reporting and reporting server and redeployed them too)
task summary report
wrong endpoint


matiasbavera:
> The UI seems to still call the old endpoint (/report/task_state) instead of the new endpoint (/report/task_summary) even though it has been updated on the deployment workspace (ws/rmf), (I rebuilt the images for reporting and reporting server and redeployed them too)

Hi @vallq, actually the script is building the image from `/ws/rmf-web/` so need to go to `ws/rmf-web/` and set your branch there. To do that you may want to add a remote (`git remote add rmf  https://github.com/open-rmf/rmf-web.git`) and then move to your remote branch `git checkout reporting-server/task -b reporting-server/task` now you can rebuild your Docker image. 

```bash
docker build -t rmf-web/reporting -f docker/reporting.dockerfile ws/rmf-web
```

Once you have the new image you need to save it ( or publish) on the Minukube's docker repository `docker save rmf-web/reporting | bash -c 'eval $(.bin/minikube docker-env) && docker load`.

After that, you need to delete the current deploy:  

```bash
.bin/minikube kubectl -- delete -f k8s/reporting.yaml
```

And deploy it again.

```bash
.bin/minikube kubectl -- apply -f k8s/reporting.yaml
```

And voilà =) 

HTH,

packages/reporting-server/models/task_summary.py
```diff
@@ -0,0 +1,128 @@
...
+    task_type: TaskType = fields.JSONField()
+    station: Station = fields.JSONField()
+    loop: Loop = fields.JSONField()
+    delivery: Delivery = fields.JSONField()
+    clean: Clean = fields.JSONField()
+
+
+class TaskProfile:
+    task_id = fields.CharField(max_length=50)
+    submission_time: Time = fields.JSONField()
```
matiasbavera, line 105: Are we using all of these models? Are we going to have a foreign key to these models?

packages/reporting-server/parsers/task_summary_parser.py
```diff
@@ -0,0 +1,24 @@
...
+            "fleet_name": state_json["fleet_name"],
+            "task_id": state_json["task_id"],
+            "task_profile": state_json["task_profile"],
+            "state": TaskSummary.service.get_task_state_name(state_json["state"]),
+            "status": state_json["status"],
+            "submission_time": state_json["submission_time"],
+            "start_time": state_json["start_time"],
+            "end_time": state_json["end_time"],
+            "robot_name": state_json["robot_name"],
+        }
```
matiasbavera, line 22: Just to check, are these field always going to have values? or should add nullable fields to the model? or should we add default values?

packages/reporting-server/parsers/test_task_summary_parser.py
```diff
@@ -0,0 +1,44 @@
+import unittest
+
+from models.task_summary import TaskSummary
+
+from .task_summary_parser import task_summary_parser
+
+# {'log': 'INFO:app.BookKeeper.task_summary:{"fleet_name": "tinyRobot", "task_id": "Loop0", "task_profile": {"task_id": "Loop0", "submission_time": {"sec": 131, "nanosec": 553000000}, "description": {"start_time": {"sec": 1623383402, "nanosec": 0}, "priority": {"value": 0}, "task_type": {"type": 1}, "station": {"task_id": "", "robot_type": "", "place_name": ""}, "loop": {"task_id": "", "robot_type": "", "num_loops": 1, "start_name": "supplies", "finish_name": "coe"}, "delivery": {"task_id": "", "items": [], "pickup_place_name": "", "pickup_dispenser": "", "pickup_behavior": {"name": "", "parameters": []}, "dropoff_place_name": "", "dropoff_ingestor": "", "dropoff_behavior": {"name": "", "parameters": []}}, "clean": {"start_waypoint": ""}}}, "state": 0, "status": "", "submission_time": {"sec": 0, "nanosec": 0}, "start_time": {"sec": 1623383362, "nanosec": 348338289}, "end_time": {"sec": 1623383449, "nanosec": 79154833}, "robot_name": "tinyRobot2"}\n', 'stream': 'stdout'}
```
matiasbavera, line 7: You can remove this example from here and you can use it on the API test instead.

packages/reporting-server/rest_server/default_config.py
```diff
@@ -3,7 +3,9 @@
     "host": "127.0.0.1",  # ip or hostname to bind the socket to
     "port": 8002,
     "port_fluentd": 8003,
-    "db_url": "sqlite://:memory:",
+    # "db_url": "sqlite://:memory:",
+    # Adding this for development
+    "db_url": "postgres://postgres:postgres@localhost:5432/reporting",
```
matiasbavera, line 8: Remember revert this once it's done =)

packages/reporting-server/rest_server/repositories/log_creation_handler.py
```diff
@@ -1,12 +1,27 @@
...
 from models.raw_log import RawLog
 from parsers.auth_event_parser import auth_event_parser
 from parsers.log_type_parser import get_log_type
 
 from .parser_dispacher import log_model_dispacher
 
-
 # Function that receives all the logs and store them on the database
+
+logger = logging.getLogger("rest_app")
```
matiasbavera, line 13: maybe `rest_app` -> `rest_app:log_creation_handler`

packages/reporting-server/rest_server/repositories/test_parser_dispatcher.py
```diff
@@ -68,6 +71,12 @@ async def test_fleet_state_created(self):
         self.assertEqual(instance[1].fleet_name, "tinyRobot")
         self.assertEqual(instance[1].robot_name, "tinyRobot2")
 
+    async def test_task_summary_created(self):
+        data = 'task_summary:{"fleet_name": "tinyRobot", "task_id": "Loop0", "task_profile": {"task_id": "Loop0", "submission_time": {"sec": 131, "nanosec": 553000000}, "description": {"start_time": {"sec": 1623383402, "nanosec": 0}, "priority": {"value": 0}, "task_type": {"type": 1}, "station": {"task_id": "", "robot_type": "", "place_name": ""}, "loop": {"task_id": "", "robot_type": "", "num_loops": 1, "start_name": "supplies", "finish_name": "coe"}, "delivery": {"task_id": "", "items": [], "pickup_place_name": "", "pickup_dispenser": "", "pickup_behavior": {"name": "", "parameters": []}, "dropoff_place_name": "", "dropoff_ingestor": "", "dropoff_behavior": {"name": "", "parameters": []}}, "clean": {"start_waypoint": ""}}}, "state": 0, "status": "", "submission_time": {"sec": 0, "nanosec": 0}, "start_time": {"sec": 1623383362, "nanosec": 348338289}, "end_time": {"sec": 1623383449, "nanosec": 79154833}, "robot_name": "tinyRobot2"}\n'
```
matiasbavera, line 75: Maybe for a follow-up issue, we can add this information in a "mock" file.

packages/reporting/src/components/reporter-side-bar-structure.tsx
```diff
@@ -58,6 +58,11 @@ export const buildReportMenuStructure = (
       title: 'Lifts',
       onClick: () => setCurrentReport(Reports.showLiftStateReport),
     },
+    {
+      icon: <ArrowDropUpIcon />,
```
matiasbavera, line 62: Can you find an icon that represents the tasks better?

packages/react-components/lib/reports/task-summary/task-summary-report-table.tsx
```diff
@@ -0,0 +1,112 @@
+import React from 'react';
```
matiasbavera, line 1: Remember to add tests for this

packages/react-components/lib/reports/task-summary/task-summary-report.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
```
matiasbavera, line 1: Remember to add tests for this

matiasbavera (changes requested):
Hi Val, left a couple of comments

vallq:
> Hi @vallq, actually the script is building the image from /ws/rmf-web/ so need to go to ws/rmf-web/ and set your branch there. To do that you may want to add a remote (git remote add rmf https://github.com/open-rmf/rmf-web.git) and then move to your remote branch git checkout reporting-server/task -b reporting-server/task now you can rebuild your Docker image

Yea I only cloned my branch for this image building, for `ws/rmf-web`! but I'll try again
maybe I missed out deleting the deployment before redeploying hmm, will look into it, thanks!

packages/reporting-server/parsers/task_summary_parser.py
vallq: Status is an empty field ("") sometimes, if I am not wrong. Robot name as well, when the task has not yet been assigned to the robot. I think we could include an empty "" field?

packages/reporting-server/parsers/test_task_summary_parser.py
vallq: Okay, got it

packages/reporting/src/components/reporter-side-bar-structure.tsx
vallq: Ok!

packages/reporting-server/models/task_summary.py
vallq: I included these models because if we want to break down the main Task Summary model into the "smallest" components, then those would be necessary (Description, Clean etc), right? Otherwise the 'task_profile' field would be a json with a lot of information

vallq:
To Do for Val
- [ ] add tests for task summary components in react-components

vallq:
task summary report

Hi @codebot @koonpeng What information would be expected of the report for Tasks? Right now we are able to retrieve the task information as logged by Fluentd. It's stored as it is in the reporting server db. But there are fields that can be removed (especially in Task Description as can be seen in the picture). Should all the info be kept in the original structure in the db and the required information be rendered on the frontend?

Also to clarify, there are 2 `submission_time` fields, does 1 refer to the time captured upon user submission (ie at t = 0) and another refers to the system / ros time?

vallq:
Gotcha, so as an example:
for a Loop task, in Task Description we would render: Number of Loops, Start Location, End Location
Delivery: Items, Pick Up Place, Drop Off Place (I'm not sure if the other fields are necessary for the frontend reporting for now). Unless Pick Up Dispenser and Drop Off Ingestor provide more clarity on the task compared to the location values

vallq:
task summary report v2
Different task description rendering based on task type

packages/reporting-server/rest_server/repositories/test_parser_dispatcher.py
matiasbavera: This is used in `test_task_summary_parser` as well. Can you please link the issue here?

packages/reporting-server/parsers/test_task_summary_parser.py
```diff
@@ -0,0 +1,42 @@
...
+from models.task_summary import TaskSummary
+
+from .task_summary_parser import task_summary_parser
+
+
+class TestCaseTaskSummary(unittest.IsolatedAsyncioTestCase):
+    def setUp(self):
+        self.data = 'task_summary:{"fleet_name": "tinyRobot", "task_id": "Loop0", "task_profile": {"task_id": "Loop0", "submission_time": {"sec": 131, "nanosec": 553000000}, "description": {"start_time": {"sec": 1623383402, "nanosec": 0}, "priority": {"value": 0}, "task_type": {"type": 1}, "station": {"task_id": "", "robot_type": "", "place_name": ""}, "loop": {"task_id": "", "robot_type": "", "num_loops": 1, "start_name": "supplies", "finish_name": "coe"}, "delivery": {"task_id": "", "items": [], "pickup_place_name": "", "pickup_dispenser": "", "pickup_behavior": {"name": "", "parameters": []}, "dropoff_place_name": "", "dropoff_ingestor": "", "dropoff_behavior": {"name": "", "parameters": []}}, "clean": {"start_waypoint": ""}}}, "state": 0, "status": "", "submission_time": {"sec": 0, "nanosec": 0}, "start_time": {"sec": 1623383362, "nanosec": 348338289}, "end_time": {"sec": 1623383449, "nanosec": 79154833}, "robot_name": "tinyRobot2"}\n'
+
+    async def test_parse_and_get_values(self):
```
matiasbavera, line 12: Related to my comment about the `null=True` can you check if those empty values are being saved as None? (just in the properties, the JSON can stay as it is)

matiasbavera (changes requested):
Looking good! Thank you Val. I left some comments.

matiasbavera (approved):
Thanks for this LGTM

PR #404, opened 2021-06-22
reporting-server:Created script and cronjob to delete logs from database > 7 days

### LINKED ISSUE #399: reporting-server: Create a cronjob that runs a query to delete data > 7 days
## Feature request

Remove data >7 days from the reporting-database

### Description

One of the specifications was to remove data from the reporting database if they're older than 7 days. One way of doing that is using a Kubernetes cronjob that calls a query to clean the database.
  
https://kubernetes.io/docs/tasks/job/automated-tasks-with-cron-jobs/

matiasbavera:

fixes #399 

## What's new
Created script to delete logs from database > 7 days. This script is built into the application.
To run it you need to activate the virtual environment and run `reporting_server_clean_logs`.

## Self-checks

### DIFF
--- a/example-deployment/README.md
+++ b/example-deployment/README.md
@@ -352,6 +352,8 @@ deploy it
 
 ## CronJobs
 
+Cronjobs are jobs that run periodically on a given schedule. You can configure the schedule following this [cron schedule syntax](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#cron-schedule-syntax)
+
 deploy cronjobs
 
 ```bash
--- a/example-deployment/README.md
+++ b/example-deployment/README.md
@@ -350,6 +350,14 @@ deploy it
 .bin/minikube kubectl -- apply -f k8s/reporting.yaml
 ```
 
+## CronJobs
+
+deploy cronjobs
+
+```bash
+.bin/minikube kubectl -- apply -f k8s/cronjobs.yaml
+```
+
 ## Test the deployment
 
 If not done so already, launch the office demo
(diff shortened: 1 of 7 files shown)


packages/reporting-server/setup.py
```diff
@@ -31,6 +31,7 @@
     entry_points={
         "console_scripts": [
             "reporting_server=rest_server.__main__:main",
+            "clean_logs=rest_server.clean_logs:main",
```
koonpeng, line 34: Can we name this to be more specific? `clean_logs` doesn't really imply that it is an auxiliary tool for the reporting server. Maybe add a prefix to the name, have a combined script that expose this function and other future functions as sub commands, or integrate this into the main script with a sub command.

example-deployment/README.md
```diff
@@ -350,6 +350,14 @@ deploy it
 .bin/minikube kubectl -- apply -f k8s/reporting.yaml
 ```
 
+## CronJobs
+
+deploy cronjobs
```
koonpeng, line 355: Should we have an example on how to configure the job schedule, maybe a link to relevant docs in kubernetes and/or cron format.

matiasbavera: Yeah! that's a good idea! Thanks!

packages/reporting-server/setup.py
matiasbavera: I added this `reporting_server_clean_logs` for now. And for the future, if I understand this right, do you want me to create something like: `reporting_server_script --clean_logs`,  `reporting_server_script --migration`, etc?

koonpeng: I think it's fine as long as one can easily identify that the script is meant for reporting server. A name like `clean_logs` can be quite vague as to what logs it's cleaning. So either a prefix on all of the reporting server scripts, or a meta-script that combines the functionalities of all the other scripts.

koonpeng (approved):

PR #406, opened 2021-06-24
reporting-server: refactor reporting-server with mocks

matiasbavera:

## What's new
The data being received from fluentd was being replicated in several places. What was done in this PR is to place all the mocks in one place, so if the structure changes we should change it in one place and if the tests fail it will be easier to detect.

## Self-checks


vallq (approved):
Thanks for this, Matias! Tested and works great; the files are cleaner with this refactor. I'll go ahead and merge this into main.

PR #407, opened 2021-06-29
Refactor/port some changes from throwaway/demo

ChawinTan:
## What's new
- cherry picked some stylistic changes from `throwaway/demo` branch
- added state sorting to fetching tasks
- added auto refresh to the robot panel page

Related to : https://github.com/orgs/open-rmf/projects/3#card-63802503

## Self-checks

## Discussion

After considering the options, I implemented the auto refresh using a `setInterval` hook to periodically fetch the robots. I feel like this would eliminate the need to do all sorts of assignment and other operations (tracking, synchronization) on the ui side and potentially a lot of complicated edge cases on the backend side should we make a new `rmfEvent`. Is it fine to use setInterval or would it introduce technical debt that I have not considered?


packages/dashboard/src/components/robots/robot-page.tsx
```diff
@@ -14,13 +14,37 @@ const useStyles = makeStyles((theme) => ({
...
+    };
+    if (autoRefresh) {
+      intervalId = setInterval(ticker, delay);
+      return () => clearInterval(intervalId);
+    } else {
+      return;
+    }
+  }, [delay, autoRefresh]);
+}
+
```
koonpeng, line 38: Is this functionally same as `setInterval` inside a `React.useEffect`?

---

`savedCallback` is not in the dependencies of the effect so you would get bugs if you change the callback. The use of `savedCallback` is also questionable, you could `useCallback` instead.

packages/dashboard/src/components/robots/robot-page.tsx
```diff
@@ -38,14 +62,14 @@ export function RobotPage() {
     return resp.data.items;
   }, [fleetsApi, page]);
 
-  React.useEffect(() => {
-    fetchVerboseRobots();
-  }, [fetchVerboseRobots]);
+  useAutoRefresh(fetchVerboseRobots, 2000, autoRefresh);
```
koonpeng, line 65: This fetches a new set of data, overriding any of the current data. So this will cause rows to jump around, items in a page to change etc. I'm not sure if this is desirable for users, and also this is different from how it works in the tasks page, we should keep the behaviour of "auto refresh" consistent.

packages/dashboard/src/components/tasks/task-page.tsx
```diff
@@ -45,7 +45,7 @@ export function TaskPage() {
         undefined,
         10,
         page * 10,
-        '-priority,-start_time',
+        'state,-priority,-start_time',
```
koonpeng, line 48: Has this been discussed with the design team?

packages/react-components/lib/tasks/task-table.tsx
```diff
@@ -19,7 +27,11 @@ const useStyles = makeStyles((theme) => ({
     },
   },
   phasesCell: {
-    padding: `0 ${theme.spacing(1)}px`,
+    padding: `0 ${theme.spacing(1)}px ${theme.spacing(1)}px ${theme.spacing(1)}px`,
```
koonpeng, line 30: material already has a separator between rows, can we instead add a margin below the phases row so that it is more prominent?

```tsx
  phasesRow: {
    marginBottom: theme.spacing(1),
  },
```

```tsx
        <TableCell className={classes.phasesCell} colSpan={5}>
          <TaskPhases className={classes.phasesRow} taskSummary={task}></TaskPhases>
        </TableCell>
```

koonpeng (changes requested):

packages/react-components/lib/tasks/task-table.tsx
ChawinTan: I think this change is already in place. I will restore it back to the original version since it should look better once the task phases are out.

ChawinTan:
On hold for now, since there are many changes to throwaway/demo that may not be applicable here and it seems to be deviating quite far

Issue #411, opened 2021-06-30
dashboard: Robot trajectories continue to be polled even when map is unmounted


koonpeng:
## Bug report

To reproduce go to the building map page, then change to the tasks page and wait for the token to expire.

Because updating robot trajectories includes refreshing an expired token, we can see that the update is still going on by inspecting the network activity.

Stack trace of the refresh call

![image](https://user-images.githubusercontent.com/45189696/123941285-987fcc00-d9cc-11eb-9422-b94f726190c5.png)

This is the part of the code that triggers the refresh.

```tsx
  React.useEffect(() => {
    let interval: number;
    (async () => {
      async function updateTrajectory() {
        if (!curMapFloorLayer || !trajManager) {
          return;
        }

        const resp = await trajManager.latestTrajectory({
          request: 'trajectory',
          param: {
            map_name: curMapFloorLayer.level.name,
            duration: trajLookahead,
            trim: true,
          },
          token: authenticator.token,
        });

        debug('set trajectories');
        if (showTrajectories === undefined || showTrajectories) {
          setTrajectories((prev) => ({
            ...prev,
            [curMapFloorLayer.level.name]: resp,
          }));
        } else {
          setTrajectories({});
        }
      }

      await updateTrajectory();
      interval = window.setInterval(updateTrajectory, trajAnimDuration);
    })();
    return () => clearInterval(interval);
  }, [trajManager, curMapFloorLayer, trajAnimDuration, showTrajectories, authenticator.token]);
```

Even though the timer should be cleared when the map is unmounted, we can see that it is not the case from the bug. My hypothesis is that in some cases, the cleanup function is called before the async function gets scheduled. In that case, the cleanup function would clear a non existent timer, and the rogue timer will continue running indefinitely.

### Required information:

- Operating system and version:
  - 
- OpenRMF installation type:
  - 
- OpenRMF version or commit hash
  - 
- ROS distribution and version:
  - 
- ROS installation type:
  - 
- Package or library, if applicable:
  -

PR #414, opened 2021-07-01
fix(dashboard): Robot trajectories continue to be polled even when map is unmounted #411

koonpeng:
Signed-off-by: Teo Koon Peng <koonpeng@openrobotics.org>

## Self-checks


matiasbavera (approved):
Code LGTM

PR #419, opened 2021-07-05
task recurrent scheduler component

matiasbavera:

## What's new
This package is responsible for generating pre-scheduled tasks. This module exposes REST endpoints for creating and managing pre-scheduled tasks. The idea is that the API-server (or rmf-server) serves as a bridge between this module and the UI. (WIP) This module should also be connected to the RMF task manager to send the task at a specific time.

Among the supported schedule are:

* At a specific time
* At a specific time on a daily schedule
* At a specific time on a weekly schedule
* At a specific time on a monthly schedule
* At a specific time on a weekday monthly schedule

It supports:
Create task rules.
Allow enable / disable specific tasks
Delete rules and specific tasks
Add additional Information to a rule task.

In the following picture we can see how this package will interact with the whole system
scheduler

task-scheduler

## Tasks to complete the feature
- [ ] Connect dashboard -> api-server -> task-scheduler. Now we are by-passing the `api-server`. dashboard->task-scheduler.
- [ ] Conect the task-scheduler package to the task dispatcher node . (Replicate the `api-server `conection)
- [ ] Add tests to the UI components
- [ ] Fix the scheduled task and rule tables pagination
- [ ]  Once we connect to the api-server, we need to generate the types of the `scheduled-task` and 
` task-rule` on OpenAPI and update the api-client with the new definitions. Remove the task-scheduler manager from the dashboard one the definitions are on Open-api

## Self-checks

- [ ] I tested the behavior of the components that interact with the backend, with an e2e test.


koonpeng:
Is this only meant for the frontend? Do you have a proposal for how the whole system will interact?

matiasbavera:
> Is this only meant for the frontend? Do you have a proposal for how the whole system will interact?

Cool, we can discuss this here, so my idea is to have a job scheduler in the backend, we could use this library https://docs.python.org/3/library/sched.html .

**jobs**: is a job to be performed by the scheduler.
**task**: is an rmf task.

We could create a DB model to save the recurrent tasks. When we send a new recurrent task definition from the dashboard to the `api-server` we should: 
1. Create the recurrent task entry on the `recurrent_task` table and 
2. Trigger or add a job to the scheduler with the task to be called in the future (with a limit to be defined), these jobs on the scheduler (once recurrent tasks are applied) should have some kind of relationship with the parent task on the `recurrent_task` table. Is important to have the previous relation, because one of the requirements is to delete the recurrent task (and from the job scheduler), and we can do that by having an identifier.

What do you think about it?

koonpeng:
Some points which I think the proposal should answer.

* How should the ui display recurring tasks, should it integration them in the tasks page, or have another page for recurring tasks?
* rmf already has a limited task scheduling functions, how much would the design make use of it and how much would it expand on the existing scheduling system?
  * If scheduling is done by api server, when should it schedule submit tasks to rmf? e.g. x mins before the start time etc.
* How should the scheduler behave when
  * It misses the time for a scheduled tasks
  * An existing instance of the task is already running when the recurring task is scheduled to run again
  * It misses the task completed event that should trigger the start of another recurring task
* The possible downsides for the proposal
  * For example, if we have a design that puts the scheduling on the api server side, the api server becomes a "critical" component of rmf as without it, scheduled tasks will not run, delivery deadlines will be missed, KPI will be degraded etc.
    * In some cases, customers have strict policies on what language is allowed in "critical" components. We have to re-evaluate where the api server stands on the hierarchy and in the worst case, we need to rewrite it completely because python may not be allowed in a "critical" component.

I think the current proposal is to do task scheduling completely on api server? If so then the biggest concern is what happens if the server is down, if it changes the "important-ness" of the api server.

Yadunund:
Just adding my 5cents here. I agree that the `api-server` should not be responsible for scheduling tasks. As @codebot  pointed out, we should develop a new library/package within RMF for this purpose ("Task Scheduler" in the diagram below). This package could maintain its own database of tasks (similar to the Schedule Database in `rmf_traffic`). Users can read/write/erase recurring tasks to this database over ROS 2 msgs. Then this `Task Scheduler` can periodically submit tasks to the `Dispatcher` which will assign the task to the most capable robot. The interval at which the Task Scheduler submits these tasks can be set through a parameter as this could be implementation dependent. No changes to the way Ad-hoc tasks will be submitted.

image

Yadunund:
Or we could also beef up the `Dispatcher Node` by developing the capabilities of the proposed `Task Scheduler` into it.

matiasbavera:
Thank you all for your comments. Being that the case I totally agreed with you that the `api-server` should not handle the task schedule. So if I understand correctly it should be something like this

scheduler

The task scheduler should be a "microservice" with its own database to store the tasks. And it should send the tasks to the task dispatcher node. I have a couple of questions:
* Should the `task scheduler` receive requests from somewhere other than the api-server? (E.g should you be listening to a topic?)
* As the `task scheduler` is a` microservice` and has its own database, by design the `api-server` should not access the` task scheduler` database directly, it should be through an endpoint. In the case of wanting to see all the recurring tasks in an interface (to modify or delete them) we should go the following way:
`UI -> api-server -> task-scheduler`. The `api-server` will be our middleman in every transaction between the UI and Task scheduler. Do we want that, or we want that the UI can connect directly with the task scheduler and be something like this?


schedulermicro

koonpeng:
I prefer diagram 1, authentication and authorization is a big problem and I think diagram 1 makes it easier to tackle. In order to return only the tasks an user can see, the scheduler needs the user's permission.

For diagram 1, the api server can pass the authorization along with the query.

For diagram 2, the scheduler needs to connect to the api server to query the user's permissions, but then that connection itself needs to be authenticated as well, so api server needs to implement alternative authentication for bot users. In diagram 2, the scheduler also needs to implement authentication etc, it also has problems in exposing an unified api to the client, we need to put it behind a reverse proxy, while not a big problem in deployment, it makes development harder as we need that reverse proxy even for local access.

However, in diagram 1, even though the scheduler is behind the api server, I think it should still develop its endpoints with first class support for the frontend in mind. Meaning it should support all the queries needed by the frontend, and if there is a new query not currently supported, it should be added directly in the scheduler instead of having the api server pre/post process the data. Instead of an ros endpoint, it may even expose rest endpoints, because rclpy is known to have performance problems, having the `api server <-> task scheduler` link be in http may greatly increase throughput.

vallq:
On hold until [Task Scheduler Node](https://github.com/orgs/open-rmf/projects/3#card-66980807) is addressed

aaronchongth:
Some feedback from show-and-tell,
- [ ] The task scheduling tab to look similar to google calendar, where times can be clicked to add a rule, and to be set as recurring or not

koonpeng:
stale and out of date

PR #425, opened 2021-07-08
Fix/reduce size of logs v1

matiasbavera:
## What's new
Hi @cnboonhan this should reduce quite a bit the size of our database, we are now adding less information to the table, so therefore the size of the table should decrease.

Since we don't have Aerich on the `main branch` yet, the changes may not be reflected on the database, if that the case and if it's possible, you may want to nuke your database and get new logs.

As a follow-up PR and working on having more FK to reduce the number of bytes used to store information.

## Self-checks


vallq (approved):
Thanks for removing the payload property and updating the tests, @matiasbavera!
@cnboonhan if it's proportional (not sure how it works) then I guess the logs should be reduced by half or so since the payload is equivalent to an entire log message

PR #428, opened 2021-07-09
Fix/Merge main into throwaway branch, along with simple panel changes

ChawinTan:

## What's new

- Merged the changes from main to `throwaway/demo`
- Changed start location to `NA` for cleaning tasks
- Rounded off battery percent to whole number precision
- Hardcoded `docking` mode to show `Cleaning`

## Self-checks


ChawinTan:
> Do we need to update the auto-generated message as well ? Otherwise it could be confusing to have `RobotMode.MODE_DOCKING` be printed out as "Cleaning" to someone stumbling on this later (?)

@codebot Thanks for your suggestions! Do u mean the reporting server when you say auto-generated messages? I think it could be helpful for debugging.

ChawinTan:
I see. This is mostly for the mile stone 3 demo and this branch would be merged into throwaway/demo branch. I personally don't think there is a need for this level of detail for a demo branch!

Yadunund:
> I'd suggest we create a new constant with a different integer in the upstream message file called MODE_CLEANING or something like that, and then whatever fleet adapter is being used, sends MODE_CLEANING when it wants the UI to print "Cleaning"

@codebot Thanks for the suggestion. I think moving forward that is exactly what we should do. Right now, the cleaning process is triggered through a docking request. This is not ideal and we plan on introducing a cleaning specific phase in the fleet adapter which can then set the RobotMode to MODE_CLEANING. With that, we should be able to differentiate between docking or cleaning.

Yadunund (approved):
Thanks for the changes @ChawinTan !
Tiny nitpick would be to center the CLEANING text 
image

Yadunund (changes requested):
Just noticed that the changes here undo the change introduced here https://github.com/open-rmf/rmf-web/commit/a70ff615d10ffbeaaed67f198b7e37ce1a64c12d

Let's aim to retain the filtered dropdown list for tasks.

Yadunund:
Seems like the sorting https://github.com/open-rmf/rmf-web/commit/61702b4d5b9ffb1c2d7ec1664704ce964406fe83 was also undone

ChawinTan:
> Seems like the sorting [61702b4](https://github.com/open-rmf/rmf-web/commit/61702b4d5b9ffb1c2d7ec1664704ce964406fe83) was also undone,
> 
> When merging `main` in, can you check that you resolved the conflicts by retaining the custom changes in this branch?

I think this is already handled by including the `state` parameter in the api `order_by` field.

```
      const resp = await tasksApi.getTasksTasksGet(
        undefined,
        undefined,
        undefined,
        undefined,
        undefined,
        undefined,
        undefined,
        undefined,
        undefined,
        10,
        page * 10,
        'state,-priority,-start_time',
      );
```

ChawinTan:
Text is already centered align. Increase the button size instead to accommodate the longer text.

increase_button_size


Added back the filtered drop down list. I noticed that there are repeated way points for `zone_x_start` and `zone_x` and removed them. Tested it locally and there doesn't seem to be any issues.

Yadunund:
I'm running into this error when i run `npm start`

```
[start:rmf-server] AttributeError: 'RmfRepository' object has no attribute 'save_task_summary'
[start:rmf-server] 
[start:rmf-server] ERROR:    Application startup failed. Exiting.
```
Possibly related to https://github.com/open-rmf/rmf-web/issues/433 ?

ChawinTan:
> I'm running into this error when i run `npm start`
> 
> ```
> [start:rmf-server] AttributeError: 'RmfRepository' object has no attribute 'save_task_summary'
> [start:rmf-server] 
> [start:rmf-server] ERROR:    Application startup failed. Exiting.
> ```
> 
> Possibly related to #433 ?

Possibly, but I can't seem to reproduce this error on my side

Yadunund (approved):
I did a clean build (deleted `node_modules/`) and the error seems to have disappeared 🤔 

The changes here are good 👍🏼  I guess we will handle the sorting in a separate PR?

ChawinTan:
Yup, the pr is coming up!

PR #430, opened 2021-07-09
Fix/reduce size of logs

matiasbavera:
## What's new

* Refactor the project to be able to use FK (related fetchings).
* Created new tables to reference data.
* Updated reports endpoints
* Update UI reports
* Updated the reporting server and UI tests.
* Refactored the log creation function.

## Self-checks


vallq (approved):
Tested the changes, thanks @matiasbavera they look good!

matiasbavera:
@cnboonhan  I was able to reproduce the error, I think now it's fixed, can you test again whenever you can?

cnboonhan:
hello, sorry it took so long to review this, I tripped a few times trying to get it deployed. So far I think its pretty aewsome! Only two comments, not directly related to reducing log size:
1. The time range to query the reporting-server gets refreshed every time we change tabs on the left-hand side panel. It might be helpful to persist the time range.
2. Currently, the IP address that is being logged in the reporting-server is the ip address of the reverse proxy used in a kubernetes deployment. Not sure if there is a way to "fix" this, but currently the ip field is "non-informational" with this ( arguably common) deployment architecture.

PR #435, opened 2021-07-15
Reporting E2E

### LINKED ISSUE #319: Add e2e test for the reporting package
## Feature request
Add e2e tests for the reporting package. 

This e2e test should use the reporting-server package as well. This would need a script to populate the database in order to return results to the UI.

This should be done using the same libraries that the Dashboard/e2e/ package use.

vallq:
Continued https://github.com/open-rmf/rmf-web/pull/417
## What's new

This PR covers the end-to-end testing for the reporting server and the reporting UI packages.
Current endpoints tested:
- Dispensers
- Doors
- Ingestors
- Health
- Fleets
- Lifts
- Task Summary

fixes #319

## Self-checks

## Discussion

Current issue is that I'm not sure how to eliminate the reporting process
image
image
I've tried using `fuser` and `killall pgrep reporting` on the `onComplete` function in the `wdio.conf.js` but those cause the test to fail alongside the following error
```
[3] Cannot stat file /proc/88773/fd/0: Permission denied
[3] Cannot stat file /proc/88773/fd/1: Permission denied
[3] Cannot stat file /proc/88773/fd/2: Permission denied
```
have included the `killOthers: ['failure', 'success'],` in the concurrently config.


vallq:
Edit:
```
[3]   {
[3]     level: 'SEVERE',
[3]     message: 'http://localhost:8088/auth/realms/master/protocol/openid-connect/login-status-iframe.html/init?client_id=reporting&origin=http%3A%2F%2Flocalhost%3A5000 - Failed to load resource: the server responded with a status of 403 (Forbidden)',
[3]     source: 'network',
[3]     timestamp: 1626320807248
[3]   },
```
Received the above error log when running `npm run test`, seems like the authentication does not recognise the reporting UI as a client?

Steps to reproduce:
1. checkout branch e2e/reporting-2
2. lerna bootstrap --scope="reporting-e2e"
3. cd packages/reporting-e2e
4. set logLevel in wdio.conf.js to 'debug'
4. npm run test

packages/reporting-e2e/README.md
```diff
@@ -0,0 +1,77 @@
...
+NOTE: This section only pertains to running the e2e tests in github workflows.
+
+This is to document the flow and interaction of the e2e services in the github environment when:
+
+- starting up the services
+- running the tests
+
+## Starting up the services
+
+Below is a diagram representing the flow of commands when running `npm run test:e2e`
```
matiasbavera, line 69: Is just `test` right?

packages/reporting-e2e/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "test:dev": "E2E_REPORTING_URL=http://localhost:3000 RMF_LAUNCH_MODE=none wdio"
+  },
+  "devDependencies": {
+    "@types/mocha": "^8.0.3",
+    "@wdio/cli": "6.12.1",
+    "@wdio/local-runner": "6.12.1",
+    "@wdio/mocha-framework": "6.11.0",
+    "@wdio/spec-reporter": "6.11.0",
+    "@wdio/sync": "6.11.0",
+    "chromedriver": "*",
```
matiasbavera, line 21: I need it to do the following "chromedriver": "91.0.1", for this to work

matiasbavera:
Yeah, I got the same error as you. That's weird =/. Did you check how the e2e test on the dashboard managed to work with the keycloak config?

vallq:
@matiasbavera boonhan helped with debugging the issue, turns out the reporting client was not being persisted in keycloak.
I updated the `keycloak.mv.db` file in both reporting-e2e and dashboard-e2e.

as for the process not exiting properly, I added a few lines to the concurrently config (ref [here](https://dev.to/raspberrytyler/exiting-node-js-when-programmatically-using-concurrently-to-run-multiple-scripts-1o78)) and to the wdio.conf.js so that the process will exit upon the termination of the test run. not exactly sure what triggered the problem but I think it has something to do with the test-e2e.js process not exiting.

I've also added in the additional test file to separate the logs creation test. it has a `before(//populate once)` before running the test because it seems that sometimes the test is executed faster than the db being 'ready'?

let me know what you think!

packages/reporting-e2e/docker/docker-compose.yml
```diff
@@ -0,0 +1,10 @@
+version: '3.5'
+services:
+  keycloak:
+    build: keycloak
+    image: rmf-web-e2e/keycloak
+    ports:
+      - 8088:8080
```
matiasbavera, line 7: If we want to run locally the `dashboard-e2e` and the `reporting-e2e` at the same time, I think we can have problems because the keycloak has the same port.

.github/workflows/reporting-e2e.yml
```diff
@@ -0,0 +1,49 @@
+name: Reporting End-to-End
+on:
+  pull_request:
+  push:
+    branches:
+      - main
```
matiasbavera, line 6: Do we want this test to run every time or just when:
* reporting-server
* reporting
* react-component
* rmf-auth
* reporting-e2e 
packages are updated?

matiasbavera (commented):
Hi @vallq, great work. A couple of comments:
* Maybe I'm wrong but `packages/reporting-e2e/docs/resources/e2e-E2e-start-process.png` doesn't look to be updated.
* Left a comment about the keycloak port 
* Can you create a follow up issue saying that we need to create a common project (or new package) that contains the similar logic for dashboard-e2e and reporting-e2e?

packages/reporting-e2e/docker/docker-compose.yml
vallq: oh hmmm. this might be good to add into the new [issue](https://github.com/open-rmf/rmf-web/issues/439)
maybe we can discuss how to separate the common components / utils as well

.github/workflows/reporting-e2e.yml
vallq: i'm not sure what common convention is but it's alright to do it as you've suggested then I think that would be better!

packages/reporting-e2e/docker/docker-compose.yml
matiasbavera: Yeah, makes sense!

matiasbavera (approved):

PR #441, opened 2021-07-19
tooltips and disable buttons

ChawinTan:
(cherry picked from commit 8eeda1b1a4532e9f937e9584609c9d59ba4745ba)

## What's new

- Tooltip helper text under `priority` field in create task form
- Toolip with name and state over robots and doors icons 
- Lifts, Dispensers and Negotiations inactive in omnipanel main menu
- Fix some e2e test and unit test as a result of these changes

### DIFF
--- a/packages/react-components/lib/tasks/create-task.tsx
+++ b/packages/react-components/lib/tasks/create-task.tsx
@@ -552,6 +552,9 @@ export function CreateTaskForm({
                       id="priority"
                       label="Priority"
                       margin="normal"
+                      helperText={
+                        'Tasks that are not active with higher priority will be given precedence'
+                      }
                       value={task.priority || 0}
                       onChange={(_ev, val) => {
                         task.priority = val;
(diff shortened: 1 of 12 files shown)


packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -552,6 +552,9 @@ export function CreateTaskForm({
                       id="priority"
                       label="Priority"
                       margin="normal"
+                      helperText={
+                        'Tasks that are not active with higher priority will be given precedence'
```
Yadunund, line 556: Can we change the text to "A task with non-zero priority score will be given precedence"

Yadunund (changes requested):
Thanks for these changes. The tooltips are pretty useful and should be carried over to `main` at some point. Just a tiny nitpick on the priority text.

Yadunund (approved):

PR #444, opened 2021-07-22
Feature/omni panel zoom

### LINKED ISSUE #445: Dashboard crash on mouse scroll wheel.
## Status

Need reproducible test case.

## Bug report
The Dashboard crash on the mouse scroll wheel. Happened while we were trying to zoom into the map with the mouse scroll wheel.

### Description of the bug

From @youliangtan I get this error sometimes. (running on a larger world map)

### Screenshots

image

image
image

### Additional information

From @ChawinTan : There were some cases where mapRefs became undefined after a while.

ChawinTan:
fixes #445 

## What's new

Zooming in to an object when the accordian is expanded on the main menu. The magnitude of zoom is currently set at 5.

When the accordion is closed, the map zooms out back to its original center and zoom value, [-66.375, 154.5], 2 on my laptop.

### DIFF
--- a/packages/dashboard/.gitignore
+++ b/packages/dashboard/.gitignore
@@ -25,5 +25,8 @@ npm-debug.log*
 web_server.log
 yarn-debug.log*
 yarn-error.log*
+.rmf_schedule_node.yaml
+log.css
 
 /.rmf
+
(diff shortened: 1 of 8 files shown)


Yadunund (changes requested):
This feature works really well and again, we should get this into `main` at some point. Can I request that we don't zoom-out when the accordion is closed? So we just retain the same view.

Yadunund (approved):

PR #451, opened 2021-07-27
Fix/make icons larger

ChawinTan:
## What's new

Use trajectory footprint for robot icons instead of hardcoded values. For now, trajectory footprint is 0.7.


Yadunund (approved):
This is great. Can we also bring this into `main` ?

PR #457, opened 2021-08-02
Refactor/requested changes 1

ChawinTan:
## What's new

1) change robot trajectory to grey when robot is cleaning
2) Remove coordinates and yaw field in omnipanel
3) Increase width of Doors
4) Unchecked waypoints
5) Increase size of tooltip when hovering
6) Increase size of robot name of the icon and shifted them out to the top right corner

### DIFF
--- a/packages/react-components/lib/robots/robot-accordion.tsx
+++ b/packages/react-components/lib/robots/robot-accordion.tsx
@@ -28,13 +28,6 @@ const RobotInfo = (props: RobotInfoProps) => {
     { name: 'Name', value: robot.name },
     { name: 'Model', value: robot.model },
     { name: 'Fleet', value: fleetName },
-    {
-      name: 'Location',
-      value: `${robot.location.level_name} (${robot.location.x.toFixed(
-        3,
-      )}, ${robot.location.y.toFixed(3)})`,
-    },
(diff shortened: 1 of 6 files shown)


Yadunund (changes requested):
In the robot onmipanel, can we keep the Location row with the `map_name`. Just need to remove the (x,y) coordinates.

packages/react-components/lib/robots/robot-accordion.tsx
```diff
@@ -28,6 +28,12 @@ const RobotInfo = (props: RobotInfoProps) => {
     { name: 'Name', value: robot.name },
     { name: 'Model', value: robot.model },
     { name: 'Fleet', value: fleetName },
+    {
+      name: 'Location',
+      value: `${robot.location.level_name} (${robot.location.x.toFixed(
```
Yadunund: Umm seems like you added back the (x, y)?

Yadunund (changes requested):

packages/react-components/lib/robots/robot-accordion.tsx
ChawinTan: Oops, I thought that was on Yaw

Yadunund (approved):

PR #458, opened 2021-08-02
refactor(api-server): tests and authz

koonpeng:
## What's new

* mock out ros2 in api server tests, this improves testing performance and reliability
* fastio's watch function is now called at subscribe time instead of when an event is received. It also no longer automatically generates a GET route.
  * the old pattern has a major flaw with authorisation. In order for authorisation to be enforceable, the decorated function needs to be called at subscription time (to check for authorisation), the old pattern calls it when an event is received. A decorator can only decorate one function, so it is not possible for the old pattern to enforce authorisation.
  * because of this change, the underlying rxpy subscription is now lazy, i.e. it only subscribes when a client calls the endpoint. This also fixes a minor problem where the old pattern needs `RmfEvent` to be available before the app is started, as a result, it requires a "global" `RmfEvent` preventing it from being restart-able.
* `RmfRepository` now requires an user, it is now designed to be used as a fastapi dependency.
  * Previously, the usage of `RmfRepository` or the tortoise models directly is a mess. With this change, route handlers should never use the tortoise models directly, "system level" operations like `RmfBookKeeper` and `HealthWatchdog` should use the tortoise models directly when they don't need a user.
* use absolute imports when importing "up" and relative imports when importing "down"
  * PEP discourages the use of relative imports, but those guides are old and there doesn't seem to be a generally accepted guideline https://stackoverflow.com/questions/4209641/absolute-vs-explicit-relative-import-of-python-module
* change the `DELETE /admin/roles/{role}/permissions` route to `POST /admin/roles/{role}/permissions/remove`
  * while not disallowed by the http spec, the `DELETE` method is normally not expected to have a body, some clients does not support it for that reason

## Self-checks


matiasbavera (approved):
Thanks for this, code LGTM, I built it from scratch and it's deploying and working correctly.

PR #460, opened 2021-08-03
fix(example-deployment): fix reporting infinite redirect

koonpeng:
## What's new

The root cause is because `/etc/nginx/conf.d/default.conf` already contains a `location / {}` block and that conflicts with the declaration in `/etc/nginx/conf.d/reporting.conf`. The reason it works in dashboard is because `dashboard.conf` gets loaded before `default.conf` because or file system sorting ("da" comes before "de").

This also adds the admin user for use with the new authorization changes. It should be used in favor of "example" when performing some things like submitting tasks.

## Self-checks

### DIFF
--- a/example-deployment/docker/dashboard.dockerfile
+++ b/example-deployment/docker/dashboard.dockerfile
@@ -19,9 +19,17 @@ FROM nginx:stable
 COPY --from=0 /root/rmf-web/packages/dashboard/build /usr/share/nginx/html/dashboard
 SHELL ["bash", "-c"]
 RUN echo -e 'server {\n\
+  listen 80;\n\
+  server_name localhost;\n\
+\n\
   location / {\n\
     root /usr/share/nginx/html;\n\
     index index.html index.htm;\n\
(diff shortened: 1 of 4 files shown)


example-deployment/docker/dashboard.dockerfile
```diff
@@ -19,9 +19,17 @@ FROM nginx:stable
 COPY --from=0 /root/rmf-web/packages/dashboard/build /usr/share/nginx/html/dashboard
 SHELL ["bash", "-c"]
 RUN echo -e 'server {\n\
+  listen 80;\n\
+  server_name localhost;\n\
```
cnboonhan, line 23: should this be example.com, in the example-deployment context?

cnboonhan, line 23: ( i accidentally deleted the previous comment sorry ) should this be example.com, in the example-deployment context?

cnboonhan: cancel the above, i think its incorrect. much to learn!

Issue #455, opened 2021-07-29
Incorrect paths in icons setup script


cnboonhan:
When i use option 2 of the icon setup script, the resources seem to get copied to the wrong path one folder too deep:
```
web@rmf-web:~/rmf-web/packages/dashboard$ ls public/assets/icons/
web@rmf-web:~/rmf-web/packages/dashboard$ npm run setup
> rmf-dashboard@0.0.1 setup /home/web/rmf-web/packages/dashboard
> node ./scripts/setup/setup.js && node ./scripts/setup/get-icons.js
Hello! 
IMPORTANT  You are about to configure the source of the resources for the RMF Dashboard project 
You have two options:
→ 1. get the assets from a remote GH source.
→ 2. copy assets from a specific folder.
→ 3. get from ws assets from workspace folder (Do not forget to source rmf_demos project before using this option).
? What option are you going to choose? 2
? Set absolute PATH /home/web/rmf/rmf_dashboard_resources
File .resources.json was created successfully.
stdout: 
The icons have been successfully obtained. Check /home/web/rmf-web/packages/dashboard/public/assets/icons/
web@rmf-web:~/rmf-web/packages/dashboard$ ls public/assets/icons/
rmf_dashboard_resources
web@rmf-web:~/rmf-web/packages/dashboard$ ls public/assets/icons/rmf_dashboard_resources/
logos  main.json  robots
web@rmf-web:~/rmf-web/packages/dashboard$ 
```
The correct output of `ls public/assets/icons/` should be `logos  main.json  robots`, i believe.

PR #461, opened 2021-08-03
minimal package: single responsibility app 

### LINKED ISSUE #420: Minimal App package
### Description
We expect that there will be operators who are in charge of specific tasks within a facility and therefore would need a single static interface that allows them to perform the operations they need. These operators would not need to see the overview of the facility or anything more complex that does not relate to their immediate tasks.

### Additional information
Reference
single task ui

vallq:
## What's new
Closes #420 

The minimal app package is a sibling package to the dashboard and acts as an interface for single responsibility functions for operators in a facility.

~~Current Issue in Deployment:~~
~~Facing a CORS error that I've not been able to resolve.~~
~~1. Create the client / scope info for minimal app in keycloak~~
~~2. `Rmf-server-config` config["aud"]: ["dashboard", "minimal"] OR just default "dashboard" value~~
~~3. Run the deployment at example.com~~
~~4. `/reporting` and `/dashboard` are fine but `/minimal` faces a CORS error~~

## Self-checks

## Discussion
Hi @matiasbavera @codebot @koonpeng, could you give your feedback and comments regarding the following:
Ideally, the interface that is launched will correspond to the user's assigned role. Is this possible / feasible with the new auth system?
~~I'm thinking of addressing comments for the app directly first and then having another PR to address deployment issues, so I would eventually remove the changes made in the example-deployment folder.~~


example-deployment/docker/minimal.dockerfile
```diff
@@ -0,0 +1,33 @@
...
+COPY . /root/rmf-web
+SHELL ["bash", "-c"]
+
+RUN npm config set unsafe-perm && cd /root/rmf-web && \
+  CI=1 npm install -g lerna@4 && lerna bootstrap --scope=minimal
+
+RUN cd /root/rmf-web/packages/minimal && \
+  PUBLIC_URL='/minimal' \
+  REACT_APP_REPORTING_SERVER='https://example.com/logserver/api/v1' \
+  REACT_APP_AUTH_PROVIDER='keycloak' \
```
cnboonhan, line 18: i suspect you might need to add a few environment variables here, based on the contents of [this file](packages/minimal/.env):

```
  SKIP_PREFLIGHT_CHECK=true \
  REACT_APP_TRAJECTORY_SERVER=ws://example.com:8006 \
  REACT_APP_RMF_SERVER=https://example.com/rmf/api/v1 \
  REACT_APP_USER=admin \
```

example-deployment/docker/minimal.dockerfile
```diff
@@ -0,0 +1,33 @@
...
+  REACT_APP_KEYCLOAK_CONFIG='{ "realm": "rmf-web", "clientId": "minimal", "url": "https://example.com/auth" }' \
+  npm run build
+
+###
+
+FROM nginx:stable
+COPY --from=0 /root/rmf-web/packages/minimal/build /usr/share/nginx/html/minimal
```
cnboonhan, line 25: will probably need to patch this https://github.com/open-rmf/rmf-web/pull/460 in, so that you do not encounter infinite redirects.

example-deployment/docker/minimal.dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+COPY . /root/rmf-web
+SHELL ["bash", "-c"]
+
+RUN npm config set unsafe-perm && cd /root/rmf-web && \
+  CI=1 npm install -g lerna@4 && lerna bootstrap --scope=minimal
+
+RUN cd /root/rmf-web/packages/minimal && \
+  PUBLIC_URL='/minimal' \
+  REACT_APP_REPORTING_SERVER='https://example.com/logserver/api/v1' \
```
cnboonhan, line 17: is there a reason for REPORTING_SERVER to be in this package?

example-deployment/k8s/minimal.yaml
```diff
@@ -0,0 +1,58 @@
...
+  labels:
+    app: minimal
+spec:
+  type: LoadBalancer
+  selector:
+    app: minimal
+  ports:
+    - protocol: TCP
+      port: 80
+      targetPort: 80
```
cnboonhan, line 14: i think this is patched out in https://github.com/open-rmf/rmf-web/pull/460

packages/minimal/scripts/setup/get-icons.js
```diff
@@ -0,0 +1,191 @@
+'use strict';
+const { exec, execSync } = require('child_process');
```
cnboonhan, line 2: i think this script needs to be fixed: https://github.com/open-rmf/rmf-web/issues/455

example-deployment/keycloak-tools/bootstrap-keycloak.js
```diff
@@ -152,7 +167,9 @@ async function createAudienceClientScope(headers, name, description, audience) {
       'reporting',
     );
 
-    // get existing clients (dashboard and reporting)
+    await createAudienceClientScope(authHeaders(token), 'minimal', 'minimal scope', 'minimal');
```
cnboonhan, line 170: ocd: could we format this the same as in previous paragraphs :laughing:  

also worth double checking this part works, i think i tried this script once and the audience scope didn't propagate, but I might have been careless, and looking at the script I dont see anything wrong so its probably me.

cnboonhan (changes requested):
looks great!! but needs some patching i think. 

Also, still stuck with the following error but in my local testing most of the CORS errors should be resolved with the commented changes. I'm not sure what `user` is about.
image

example-deployment/docker/minimal.dockerfile
vallq: addressed!

packages/minimal/src/managers/resource-manager-dispensers.ts
```diff
@@ -0,0 +1,63 @@
+import { RESOURCE_PREFIX } from './resource-manager';
```
koonpeng, line 1: Duplicated code, at first glance, this probably better belongs in the new util/tools package, but the problem is that this is runtime code that depends on build time configurations, I expect the util/tools package to contain mostly build time tools, so including this in it would force dependants to add a build tools package as a runtime dependency.

So maybe we should move this to `react-components` instead, this does create a hidden dependency on the tools package for `react-components`, but maybe thats not a problem.

Another option (which on 2nd thought, probably doesn't work) is to simplify the resource loading, icons could be loaded via a standard path so we don't need extra logic of parsing a descriptor files. But there are also things like dispenser locations, which should have never been part of the configuration in the first place, but alas rmf lacks the api for us to obtain those info so we have to embed it. We can move these info into the api server, but that doesn't really solve the problem, now the api server needs to embed these info instead.

packages/minimal/src/managers/rmf-health-state-manager.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export interface HealthStatus {
+  door: ItemSummary<SpoiltDoor>;
+  lift: ItemSummary<SpoiltLift>;
+  dispenser: ItemSummary<SpoiltDispenser>;
+  robot: RobotSummary;
+}
+
+// TODO - fill up health state manager logic once backend is up
+export default class RmfHealthStateManager {
```
koonpeng, line 17: Duplicated code, I don't think this is used, but if it needs to be, it should probably be moved to `react-components`.

packages/minimal/src/managers/trajectory-socket-manager.ts
```diff
@@ -0,0 +1,51 @@
+import EventEmitter from 'eventemitter3';
```
koonpeng, line 1: Duplicated code which should be moved to `api-client`.

packages/minimal/src/util/calculation-helpers.ts
```diff
@@ -0,0 +1,15 @@
+export function radiansToDegrees(radians: number) {
```
koonpeng, line 1: Duplicated code which should be moved into `react-components`.

packages/minimal/src/util/cublic-spline.ts
```diff
@@ -0,0 +1,164 @@
+export interface Pose2D {
```
koonpeng, line 1: Duplicated code which should be moved into `react-components`.

packages/minimal/src/util/react-helpers.ts
```diff
@@ -0,0 +1,60 @@
+/**
```
koonpeng, line 1: Duplicated code which should be moved into `react-components`.

packages/rmf-auth/lib/authenticator.d.ts
```diff
@@ -0,0 +1,31 @@
+import EventEmitter from 'eventemitter3';
```
koonpeng, line 1: This and all the other declaration files shouldn't be committed.

packages/rmf-auth/lib/authenticator.js
```diff
@@ -0,0 +1 @@
+export {};
```
koonpeng, line 1: This and all the other js files shouldn't be committed.

koonpeng (changes requested):

koonpeng:
There are lots of duplicated code, we should probably make a large scale refactor to solve them before merging in this.

packages/api-server/api_server/models/fleets.py
```diff
@@ -1,8 +1,7 @@
 from typing import List
 
-from pydantic import BaseModel
-
 from api_server.models.tasks import Task
+from pydantic import BaseModel
```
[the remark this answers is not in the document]
vallq: this is the .isort.cfg
```
[settings]
profile=black
# project=api_server
```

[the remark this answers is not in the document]
koonpeng: Doesn't seem like it is changed, but the import sorting is wrong. Maybe you have another version of isort?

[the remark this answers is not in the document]
vallq: this is what I get when I run isort in the terminal: VERSION 5.9.3
is this what it should be?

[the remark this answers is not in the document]
koonpeng: thats the same version I have, maybe you have some config which overrides the one in the project? Or maybe your ide is configured to not use the project config?

vallq:
- Moved setup and get icons script and rmf-launcher into an rmf-tools package
- I initially removed robot and dispenser resource managers but with the new changes I've kept them there; trying to figure a way to move the manager / resources component into rmf-tools (any advice here?)
- Removed extra code (Tabs, Settings panel, Rmf health status etc)
- Not sure why but pulling in new changes from main did not update the import line order in affected files for `api-server` hence the linting errors, but those have been addressed

Other things to be done:
- [ ] Deployment instructions
- [ ] Readme
- [ ] Dynamic updating of document.title
- [ ] Moving resource manager into rmf-tools (make the different resources optional based on what the app requires?) 
Is it possible to select the resources needed based on an interface's requirements? (ie the minimal app does not require robot icons, so those should not be in the src/assets/resources folder)
- [ ] DataConfig / Structure of the app
> Is this (and other configs in the directory) meant to be configurable according to the deployment? If so, we should simplify these configs to a declarative format (json or "json-like js") and have it be generated/copied in a setup phase (similar to how the configs with resources works).
It still needs work but what I envision is that authentication / retrieving an env variable would specify the user's role, which in turn would render the right 'options' or selectable choices in the task submission fields on the app, and even the type of tasks that can be sent. Is this realistic/feasible?

aaronchongth:
Per VC, this draft PR will be closed as the main branch's development has diverged

Issue #422, opened 2021-07-06
Make phase info more subtle in the Task Panel


Yadunund:
![image](https://user-images.githubusercontent.com/13482049/124565878-12a0cc80-de75-11eb-86cc-548992bbfb3d.png)

This ticket is more about providing user feedback on the visuals in the Task Panel. Currently the phase information for a task seems to dominate the task itself. When looking at the Task Panel for the first time, users are naturally drawn to the phase arrows and it is not obvious that the phase arrows are describing the status of the task in the row above.

Some helpful tweaks could be to make the font size of the row above the phase info to be larger and bolder in comparison? Perhaps making the arrows narrower and more blank padding on either side so that it looks like its "contained" within the description above?

An alternative visualization of the phases could be in "timeline form"?

![image](https://user-images.githubusercontent.com/13482049/124572317-24856e00-de7b-11eb-9c23-991aff3ca49b.png)

Looking forward to discuss more 👍🏼 

Edit: Using this document for "wireframes" 
https://docs.google.com/presentation/d/1RtM44yDQmMK42UbT7T4KZrKIp74v5zExnSS9HnRzYNU/edit#slide=id.gb80df710a2_0_74

Yadunund:
Adding more ideas:
Retain the horizontal chevron for phases 

![image](https://user-images.githubusercontent.com/13482049/124577832-39b0cb80-de80-11eb-940a-3d80e6f22c25.png)


Phase info is "collapsed" by default. Clicking on a row, makes it visible
![image](https://user-images.githubusercontent.com/13482049/124577959-56e59a00-de80-11eb-808f-99239e8ebb17.png)

codebot:
I'm not a UI expert, but I think having it collapsible is probably good for small screens (as well as being able to see more tasks on large screens) Or I guess we could be super advanced and have it "responsive" based on screen size, whether it defaults to collapsed or expanded :sparkles:

Yadunund:
@codebot thanks for the feedback! 🙇🏼‍♂️

vallq:
https://user-images.githubusercontent.com/28668944/124895220-f042b680-e00e-11eb-96a9-08ac80397392.mp4
turns out Material UI has a collapsible table

Yadunund:
@vallq that looks good

koonpeng:
I think we need at least 3 design, for mobile, tablet and desktop respectively.

[...]

PR #464, opened 2021-08-06
ui/phase info fix

vallq:
## What's new
continuing from #422 

Option 1:
https://user-images.githubusercontent.com/28668944/128492109-d735c671-caac-41db-a99a-706a4a005414.mp4
I tried my best with this but the expansion of each entry will always open the side panel

Option 2 (as Yadu mentioned about having the phase info in the side panel):
image
timeline

## Self-checks

## Discussion


vallq:
Hi @koonpeng and @Yadunund ,
I added 2 panels to the task page in the dashboard to show the different options we have, and I will proceed to make amendments in the modified files once we've decided on which to go with / other options. Let me know what you both think! If need be, we could poll on the sg channel as well.

koonpeng:
I haven't tested it yet, judging from the video, I can see the problems with option 1.

* When you hover on top of the row, it is highlighted but the cursor doesn't change to a pointer, the highlight suggests it is clickable but the pointer does suggests otherwise.
  * If you make the pointer change, then it is extra confusing as there is nothing that indicates clicking on the down arrow has a different behavior.
* Clicking on the task only shows it on the side panel, with no indication that the down arrow is also clickable to show the phases.
  * This is in contrast to a typical accordion where the whole header is clickable.

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -0,0 +1,122 @@
...
+            <TimelineOppositeContent>
+              <Typography variant="body2" color="textSecondary">
+                {idx == 0 && rosTimeToJs(taskSummary.start_time).toLocaleTimeString()}
+                {idx != 0 &&
+                  idx == timelineInfo.length - 1 &&
+                  rosTimeToJs(taskSummary.end_time).toLocaleTimeString()}
+              </Typography>
+            </TimelineOppositeContent>
+            <TimelineSeparator>
+              <TimelineDot className={timelineDotProps[idx].className} />
```
koonpeng, line 106: since the object is called `timelineDotProps`, I would expect it to be passed directly to `TimelineDot`.

```suggestion
              <TimelineDot {...timelineDotProps[idx]} />
```

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -0,0 +1,122 @@
...
+              {idx < timelineInfo.length - 1 && <TimelineConnector />}
+              {idx == timelineInfo.length - 1 && (
+                <TimelineConnector style={{ background: 'transparent' }} />
+              )}
```
koonpeng, line 110: Why not just skip rendering the last connector?

packages/react-components/lib/tasks/task-timeline.spec.tsx
```diff
@@ -0,0 +1,34 @@
...
+import React from 'react';
+import * as RmfModels from 'rmf-models';
+import { TaskTimeline } from './task-timeline';
+import { makeTask } from './test-data.spec';
+
+describe('Task Timeline', () => {
+  it('shows all task phases', () => {
+    const task = makeTask('task_0', 3, 3);
+    const root = render(<TaskTimeline taskSummary={task} />);
+    expect(root.getByText('Phase 1 test phase 1')).toBeTruthy();
```
koonpeng, line 11: The test is "shows all task phases", but it only checks the first task phase.

packages/react-components/lib/tasks/task-timeline.spec.tsx
```diff
@@ -0,0 +1,34 @@
+import { render } from '@testing-library/react';
+import React from 'react';
+import * as RmfModels from 'rmf-models';
+import { TaskTimeline } from './task-timeline';
+import { makeTask } from './test-data.spec';
+
+describe('Task Timeline', () => {
```
koonpeng, line 7: Should add a test that it shows the time as well.

packages/dashboard/src/components/tasks/task-page.tsx
```diff
@@ -92,23 +92,45 @@ export function TaskPage() {
...
-      cleaningZones={placeNames}
-      loopWaypoints={placeNames}
-      deliveryWaypoints={placeNames}
-      submitTasks={submitTasks}
-      cancelTask={cancelTask}
-      onRefresh={handleRefresh}
-      onAutoRefresh={autoRefreshDispatcher.setEnabled}
-    />
+    <div>
+      <TaskPanel
```
koonpeng, line 96: I assume the 2 task panels is just for demo?

vallq: yes this is just for demo purposes, will remove when we've made a decision on which alternative to use

packages/react-components/lib/tasks/task-table.tsx
```diff
@@ -15,26 +20,34 @@ const useStyles = makeStyles((theme) => ({
     background: theme.palette.action.hover,
   },
   infoRow: {
+    boxShadow: `${theme.shadows[1]}`,
     '& > *': {
       borderBottom: 'unset',
     },
```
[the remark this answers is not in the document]
vallq: oh this was something mentioned awhile back when we were on the throwaway/demo branch that the default line was rather faint hence I thought to implement a shadow instead. I could remove this first and add it back if needed

packages/react-components/lib/tasks/task-table.tsx
```diff
@@ -15,26 +20,34 @@ const useStyles = makeStyles((theme) => ({
...
+    boxShadow: `${theme.shadows[1]}`,
+    '&:last-child': {
+      paddingRight: `${theme.spacing(1)}px`,
+    },
   },
   phasesRow: {
     marginBottom: theme.spacing(1),
+    marginTop: theme.spacing(1),
   },
 }));
```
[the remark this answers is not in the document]
vallq: yea! but I was thinking we could have a separate story / component using the phases and keep it as an option within react-components

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -0,0 +1,122 @@
+import React from 'react';
```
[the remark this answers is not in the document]
vallq: will look into this!

packages/react-components/lib/tasks/task-timeline.tsx
vallq: I think when I was implementing this, I had a situation where removing the last connector will shift the last dot and info component towards the left if the connector is not rendered, but that seems to not happen now, so I will remove this additional connector. might have missed it previously

packages/react-components/lib/tasks/task-table.tsx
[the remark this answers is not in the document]
koonpeng: The phases is in its own component, I don't think it is worth it to create different variants of the table, not unless we actually have a use case for different variants, better to keep things simple to avoid bloating it.

packages/react-components/lib/tasks/task-timeline.tsx
koonpeng: :+1: feel free to add this in if that problem comes back, maybe with a comment explaining why. Just want to make sure this is intended and not an oversight.

vallq:
1. removed unused code
2. fixed the info div to have a fixed width and max height, and to be scrollable when the content gets long
3. shifted the timeline to be left-aligned
4. fixed failing CI test by checking for all possible time values with regex

image

vallq:
@Yadunund how does this look? I think a highlight of the cell would be more consistent because I think that's what Chawin has implemented for the robots page as well, in his theming PR
task-page-updates

Yadunund:
@vallq thanks for iterating! Yes, let's have a vote in `sg-web`

PR #467, opened 2021-08-11
Feature/building-tab-components

ChawinTan:
## What's new

Remove the omnipanel and replace doors, lifts and dispensers with a fixed side panel. The side panel can toggle between cards/cells and a table format.

Cell Format:
building_with_panel

Table Format:
building_panel_table

Other changes:
- move unused components into a `_trash` folder 
- unit tests
- story for the components
- changes to e2e tests

## Self-checks


packages/react-components/lib/doors/utils.ts
```diff
@@ -0,0 +1,75 @@
...
+  AntiClockwise = -1,
+}
+
+export enum DoorMode {
+  Open = RmfModels.DoorMode.MODE_OPEN,
+  Closed = RmfModels.DoorMode.MODE_CLOSED,
+  Moving = RmfModels.DoorMode.MODE_MOVING,
+}
+
+export interface DetailedDoor extends RmfModels.Door {
```
koonpeng, line 23: I would prefer to avoid making these extension to the models. It creates confusion where some places uses  `DetailedDoor` and others using the base `Door`. Sometimes its also confusing how can I create a `DetailedDoor`.

Some workaround:
* Instead of inheritance, use composition, and name it something like `DoorData`, with type `{ door: Door, level: string }`. At least to me, it's more obvious that you can create the object by mapping the building map level with the doors. Inheritance can mean that `Door` needs to be modified or processed in some way while composition makes it more obvious that you can use the `Door` from building map "as is".
* For cases where it only needs one door, just have 2 props, the `Door` and the level.
* For cases where you need an array, one solution is to use a tuple `[level: string, door: Door]`, but it isn't as readable as using composition (actually this is also a composition in a way).

packages/react-components/lib/lifts/lift-panel.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+      />,
+    );
+  });
+
+  it('should open up form dialog when Request Form button is clicked', () => {
+    userEvent.click(root.getAllByRole('button', { name: /Request Form/i })[0]);
+    expect(root.getByPlaceholderText('Pick a Destination')).toBeTruthy();
+  });
+
+  it('should call show layout when view mode button is clicked', () => {
```
koonpeng, line 35: Same comment as dispenser.

packages/react-components/lib/lifts/lift-request-form.tsx
```diff
@@ -91,70 +110,77 @@ export const LiftRequestForm = (props: LiftRequestFormProps): JSX.Element => {
...
-          options={['', ...lift.levels]}
-          renderInput={(params) => (
-            <TextField
-              {...params}
-              label="Pick a Destination"
-              placeholder="Pick a Destination"
-              variant="outlined"
-              error={!!destinationError}
-              helperText={destinationError}
+    <Dialog open={showFormDialog} onClose={() => onClose()} fullWidth={true} maxWidth={'md'}>
```
koonpeng, line 113: Might want to use the `ConfirmationDialog` (see the authz components in dashboard for an example) to keep a consistent ui/ux.

packages/react-components/lib/lifts/lift-table.spec.tsx
```diff
@@ -0,0 +1,11 @@
+import { render } from '@testing-library/react';
+import React from 'react';
+
+import { LiftTable } from './lift-table';
+import { testLifts, testLiftStates } from './test-utils.spec';
+
+describe('lift table', () => {
+  it('should render properly', () => {
```
koonpeng, line 8: Should at least test the relevant info can be seen (lift names, level etc).

packages/react-components/lib/lifts/lift-table.tsx
```diff
@@ -0,0 +1,83 @@
+import React from 'react';
```
koonpeng, line 1: It seems that this view is missing the functionality to open the lift request dialog.

packages/react-components/lib/lifts/lift-request-form.tsx
```diff
@@ -1,12 +1,19 @@
-import { makeStyles } from '@material-ui/core';
+import { makeStyles, Dialog, IconButton, DialogContent } from '@material-ui/core';
```
koonpeng, line 1: Since this is now a dialog, might want to rename it as such.

koonpeng (changes requested):

koonpeng:
I feel that the panels are missing title/headers.

Do you think it would look better if we use `Card` with `variant="outlined"` for the panel container? So each panel are clearly separated.

ChawinTan:
- Change dispensers to workcells
- Use a card for all the panels
- Kept the workcell panel as one panel, but within, it displays ingestors and dispensers subpanels.

koonpeng (commented):
image

I think suffixes like "... Panel", "... Table" only makes sense for developers and is not user friendly.

Some of the text, e.g. when a door is closed is very strong and hard to read.

Needs some additional styling to differentiate the name of the object from its data (bigger text etc).

---

I think I will put this on hold until we can confirm if we are going for the omnipanel or multi panel design.

packages/react-components/lib/doors/door-panel.spec.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { DoorPanel } from './door-panel';
+import { doorStates, makeDetailedDoors } from './test-utils.spec';
+
+function renderDoorPanel() {
+  const mockControlClickSubmit = jasmine.createSpy();
+  return render(
+    <DoorPanel
+      doors={makeDetailedDoors()}
+      doorStates={doorStates}
+      onDoorControlClick={mockControlClickSubmit}
```
koonpeng, line 14: The spy is not accessible from the tests so theres probably no need to have them here. If onDoorControlClick is a required props, should probably make it optional.

packages/react-components/lib/lifts/lift-panel.spec.tsx
```diff
@@ -0,0 +1,39 @@
+import { render } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
+
+import { LiftPanel } from './lift-panel';
+import { testLifts, testLiftStates } from './test-utils.spec';
+
+function renderLiftPanel() {
+  const mockRequestSubmit = jasmine.createSpy();
```
koonpeng, line 9: Can be removed as its not accessible to tests

PR #469, opened 2021-08-17
feat: trajectory time control

### LINKED ISSUE #432: Allow users to specify the duration and trim parameters for trajectories
Currently the `trajLookahead` and `trim` [parameters](https://github.com/open-rmf/rmf-web/blob/5ac7415728639a9b56f71e66cd4f3aa462e29b65/packages/dashboard/src/components/schedule-visualizer/index.tsx#L199-L200) are hardcoded to be 1min and true resp. For large maps where robots have travel times of greater than 1min, the trajectories are partially rendered due to these parameters. It would be great if users can set these parameters from the dashboard along with the animation style.

image

koonpeng:
## What's new

https://user-images.githubusercontent.com/45189696/129659966-17c3d8a4-bc37-4779-8eb7-f3b23bce12e2.mp4

Fixes #432

* remove trajectory animation setting as there isn't much point in being able to change it, allowing different animations also brings about unnecessary complexity in rendering different trajectory length, state etc.
* remove settings drawer as there is no settings currently.
* remove tour as it depends on settings drawer and is no longer up to date, the "dashboard" it works on is also about to be removed according to the design plan.
* added trajectory time control widget in the schedule visualizer.

It is not possible to make the `trim` parameter optional because it would result in uncontrollable trajectory lengths, causing either animations to skip or animation speed to be wrong.

### Known issue

When dragging the time control slider, the leaflet maps moves as well. This is caused by a conflict on how react handle events. leaflet provides a helper function to stop events from propagating from the controls to the map, but react <= 16 installs a global event handler on the document root, if we stopped the propagation on the controls, it would not reach the document root, hence react won't see those events.

## Self-checks


Yadunund:
Thanks for adding this feature 🙇🏼‍♂️ 

During my testing while adjusting the duration slider back and forth, I encountered this error
image

Secondly, I notice that while dragging the slider, the background map tends to pan as well. Its a minor inconvenience and could be addressed in the future.

Yadunund:
I think the issue occurs when the slider is moved all the way to the left. It takes on `NaN`

image

koonpeng:
@Yadunund Should be fixed now.

Yadunund (approved):
Works great 👍🏼

Issue #214, opened 2021-01-25
Update material ui from v4 to v5


ChawinTan:
Update Material-ui from v4 to v5 once it is officially released

PR #474, opened 2021-08-20
Update/material v4 to v5

ChawinTan:
## What's new

Upgraded material v4 to v5. This fixes some issues regarding the date pickers.

Major changes include:
- shift away from `makeStyles` to emotion.
- remove dependency on `material-table` in the report components and replace it with material `datagrid` since it is not compatible with v5 and development on it seems to be slow.

Useful docs:
https://next.material-ui.com/guides/migration-v4/

Related to #214 

## Self-checks

## Discussion

There are many changes to material ui and it will affect existing prs. Should we wait till the more important feature prs are merged first before merging this?


packages/react-components/lib/reports/log-management/custom-lookup-filter.tsx
```diff
@@ -22,72 +20,33 @@ const MenuProps = {
...
 
 export const CustomLookupFilter = (props: CustomLookupFilterProps): React.ReactElement => {
-  const { tableId, filterValue, filterOnItemSelect, onFilterChanged, lookup } = props;
-  const [selectedFilter, setSelectedFilter] = React.useState(filterValue || []);
-  React.useEffect(() => {
-    setSelectedFilter(filterValue || []);
-  }, [filterValue]);
-
+  const { lookup, selectedFilter, setSelectedFilter } = props;
+  const [filterTracker, setFilterTracker] = React.useState<string[]>(selectedFilter);
```
koonpeng, line 31: This state seems useless and introduces potential bugs. If `selectedFilter` changes, `filterTracker` will not be updated.

packages/react-components/lib/reports/log-management/custom-lookup-filter.tsx
```diff
@@ -22,72 +20,33 @@ const MenuProps = {
...
+    <FormControl style={{ width: '100%', padding: '1rem' }}>
+      <InputLabel htmlFor={'select-multiple-checkbox'}>Level Filter</InputLabel>
       <Select
         multiple
-        value={selectedFilter}
+        value={filterTracker}
         onClose={() => {
-          if (filterOnItemSelect !== true) onFilterChanged(tableId.toString(), selectedFilter);
+          setSelectedFilter(filterTracker);
+          setFilterTracker(filterTracker);
```
koonpeng, line 40: `setSelectedFilter` does not actually do anything because `filterTracker` is not updated when `selectedFilter` changes. There will be bugs when the parent component process the new set of filters before updating them, this component will not respond to the processed filters.

packages/react-components/lib/reports/log-management/custom-lookup-filter.tsx
```diff
@@ -22,72 +20,33 @@ const MenuProps = {
...
         }}
         onChange={(event) => {
-          setSelectedFilter(event.target.value);
-          if (filterOnItemSelect === true) onFilterChanged(tableId.toString(), event.target.value);
+          if (typeof event.target.value === 'string') {
+            setFilterTracker([...filterTracker, event.target.value]);
+          } else {
+            setFilterTracker(event.target.value);
+          }
```
koonpeng, line 47: These changes are not reported to the parent component, you need to call `setSelectedFilter`, but as mentioned above, `setSelectedFilter` is useless and buggy.

All these can be fixed by simply removing `filterTracker` state, and just use `selectedFilter` and a `onFilterChange` callback. The parent should be responsible for updating the filters.

packages/react-components/lib/reports/log-management/log-table.tsx
```diff
@@ -19,24 +19,31 @@ export type LogRowsType = {
 
 export interface LogTableProps {
   rows: LogRowsType | [];
-  tableSize?: string; // units vh or rem
+  tableSize?: number; // units vh or rem
```
koonpeng, line 22: This doesn't seem to be used?

packages/react-components/lib/reports/log-management/search-log-form.spec.tsx
```diff
@@ -17,9 +17,8 @@ describe('Search log form tests', () => {
     const root = render(<SearchLogForm logLabelValues={logLabel}></SearchLogForm>);
     expect(root.getByText('ALL')).toBeTruthy();
     const currentDate = format(new Date(), 'MM/dd/yyyy HH:mm');
-    const fromLogDate = root.container.querySelector('#fromLogDate-datetime-local');
-    expect(fromLogDate?.getAttribute('value')).toBe(currentDate);
-    const toLogDate = root.container.querySelector('#toLogDate-datetime-local');
-    expect(toLogDate?.getAttribute('value')).toBe(currentDate);
+    const getInputs = root.container.querySelectorAll('.MuiOutlinedInput-input.MuiInputBase-input');
+    expect(getInputs[2].getAttribute('value')).toBe(currentDate);
+    expect(getInputs[3].getAttribute('value')).toBe(currentDate);
```
koonpeng, line 22: Should use a better selector, the input fields should have a label which you can use to select it.

packages/react-components/lib/reports/log-management/search-log-form.tsx
```diff
@@ -27,39 +51,31 @@ export const SearchLogForm = (props: SearchLogFormProps): React.ReactElement =>
...
-      setLogLevel(event.target.value as LogLevel);
-    },
-    [],
-  );
+  const handleLogLevelChange = React.useCallback((event: SelectChangeEvent<React.ReactText>) => {
+    setLogLevel(event.target.value as LogLevel);
+  }, []);
 
-  const handleFromLogDateChange = React.useCallback((date: MaterialUiPickersDate) => {
+  const handleFromLogDateChange = React.useCallback((date: any) => {
```
koonpeng, line 69: Don't use `any`, you can use `Required<DateAndTimePickers>['onChange']` to grab the callback type, or use `Date` since we already know `fromLogDate` is a `Date`.

packages/react-components/lib/reports/utils.ts
```diff
@@ -1,8 +1,6 @@
-import { MaterialUiPickersDate } from '@material-ui/pickers/typings/date';
-
 export interface ReportConfigProps {
-  fromLogDate?: MaterialUiPickersDate;
-  toLogDate?: MaterialUiPickersDate;
-  onSelectFromDate?: (date: MaterialUiPickersDate) => void;
-  onSelectToDate?: (date: MaterialUiPickersDate) => void;
+  fromLogDate?: Date;
+  toLogDate?: Date;
+  onSelectFromDate?: (date: any) => void;
```
koonpeng, line 4: Don't use `any`

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -9,51 +9,56 @@ import {
...
 import { rosTimeToJs } from '../utils';
 
-const getPhaseColors = (theme: Theme) => ({
-  pending: theme.palette.info.light,
-  completed: theme.palette.success.light,
-  failed: theme.palette.error.light,
-});
+interface TimeLinePropsWithRef extends TimelineProps {
+  ref?: React.RefObject<HTMLUListElement>;
+}
```
koonpeng, line 20: This type definition is potentially unsafe, we do not know if the ref is actually a `HTMLUListElement`, there are several mui components which you can customize to change the html tag, so there might not even be a fixed type.

Would it be possible to use `React.PropsWithRef`?

koonpeng (changes requested):
According to the mui docs, the preferred why for styling is to use the `sx` props for one-time styles and styled components for reusable components.

The approach used here may be slightly faster but I'm not sure if it is enough to justify deviating from the recommendations.

@vallq @codebot What are your thoughts on this?

packages/react-components/.storybook/main.js
```diff
@@ -4,8 +4,18 @@ module.exports = {
...
+  webpackFinal: async (config) => {
+    return {
+      ...config,
+      resolve: {
+        ...config.resolve,
+        extensions: ['.ts', '.tsx', '.mjs', '.js', '.jsx', '.cjs'],
+        alias: {
+          ...config.resolve.alias,
+          '@emotion/core': require.resolve('@emotion/react'),
+          'emotion-theming': require.resolve('@emotion/react'),
```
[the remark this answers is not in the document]
ChawinTan: https://github.com/mui-org/material-ui/issues/24282#issuecomment-796755133
https://mui.com/guides/migration-v4/#storybook-emotion-with-v5

We need to let storybook use the latest version of emotion so that the theme gets picked up correctly.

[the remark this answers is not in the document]
koonpeng: Can you add a FIXME explaining this?

packages/react-components/lib/map/label-marker.spec.tsx
```diff
@@ -6,7 +6,7 @@ describe('NameLabel', () => {
   it('smoke test', () => {
     const root = render(
       <svg>
-        <NameLabel sourceX={0} sourceY={0} sourceRadius={0} text="test" />,
+        <NameLabel sourceX={100} sourceY={100} sourceRadius={0} text="test" />,
```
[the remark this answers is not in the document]
ChawinTan: There was some problem with using 0 initially. Revereted.

packages/dashboard/src/components/admin/manage-roles-dialog.tsx
```diff
@@ -12,25 +12,32 @@ import {
...
-const useStyles = makeStyles((theme) => ({
-  action: {
+const prefix = 'manage-roles-dialog';
+const classes = {
+  action: `${prefix}-action`,
+  list: `${prefix}-list`,
+  dialogContent: `${prefix}-content`,
+  dialogButton: `${prefix}-button`,
+};
+const ManageRolesDialogRoot = styled((props: CardProps) => <Card {...props} />)(({ theme }) => ({
```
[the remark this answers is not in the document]
ChawinTan: Should I also change all those whose base components are HTML tags to `StyledXXX`?

packages/react-components/.storybook/preview.tsx
```diff
@@ -1,47 +1,46 @@
...
   const theme = getTheme(context.globals.theme);
-  useStyles();
   return (
-    <ThemeProvider theme={theme}>
-      <StorybookBodyStyles />
-      <Story {...context} />
-    </ThemeProvider>
+    <StyledEngineProvider injectFirst>
+      <EmotionThemeProvider theme={theme}>
+        <ThemeProvider theme={theme}>
```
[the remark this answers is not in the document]
ChawinTan: Hmmm not sure how `EmotionThemeProvider` is related to the maps?

packages/react-components/lib/commands/form-styles.ts
```diff
@@ -1,24 +1,31 @@
-import { makeStyles } from '@material-ui/core';
+import { styled } from '@mui/material';
 
-export const useFormStyles = makeStyles((theme) => ({
-  form: {
+export const commandFormsClasses = {
```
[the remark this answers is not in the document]
ChawinTan: We probably still need a `command-form` in front to ensure specificity. If there is `full-width` in a different component, the 2 classes would conflict.

packages/react-components/lib/header-bar.stories.tsx
```diff
@@ -14,6 +14,16 @@ export default {
...
+const StyledTab = styled((props: TabProps) => <Tab {...props} />)(({ theme }) => ({
+  color: 'rgba(255, 255, 255, 0.7)',
+  '&.Mui-selected': {
+    color: theme.palette.text.primary,
+  },
+  '&.Mui-focusVisible': {
+    backgroundColor: 'rgba(100, 95, 228, 0.32)',
+  },
+}));
+
```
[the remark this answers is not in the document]
ChawinTan: This is a copy of the `StyledTab` in `dashboard/appbar`. Without it, the Tab component would look weird. I think I will create a separate component for the StyledTab and remove the existing ant-tab since it is not in use.

packages/react-components/lib/reports/log-management/custom-lookup-filter.tsx
ChawinTan: I think I remember why I did that. The `select` dropdown would close every time a new filter is selected which is bad for UX. Adding an extra `filterTracker` in onChange first before triggering `onFilterChange` in onClose prevents unnecessary rerenders but I guess that was not a good way to solve that issue. :sweat_smile:

packages/react-components/lib/tasks/task-timeline.tsx
ChawinTan: It seems like the `TimeLine` RefObject type is fixed as `HTMLUListElement`. Can't use React.PropsWithRef

packages/dashboard/src/components/admin/manage-roles-dialog.tsx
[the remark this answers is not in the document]
koonpeng: Yeah, at least thats the pattern I see from the examples.

packages/react-components/lib/map/map.tsx
```diff
@@ -10,8 +10,11 @@ import { LabelsPortalContext } from './labels-overlay';
 import { SVGOverlay } from './svg-overlay';
 import { viewBoxFromLeafletBounds } from './utils';
 
-const useStyles = makeStyles(() => ({
-  map: {
+const classes = {
+  map: 'map-root',
+};
+const MapRoot = styled((props: LMapProps_) => <LMap_ {...props} />)(() => ({
+  [`&.${classes.map}`]: {
```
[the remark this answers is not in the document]
ChawinTan: `@mui/systems` seems promising at first but is used to build a design system from scratch and it seems to break when I tried to use it on just one component.

https://mui.com/guides/classname-generator/
This is the closest to making new classname, but it seems limited to modifying and renaming existing classnames

PR #476, opened 2021-08-24
Fix/robot pagination

ChawinTan:
This bug is caused by a robot not queried in the list of robots as it is out of scope of the pagination, but exists in the task that it is assigned to, causing the api server to keep throwing an error as the task is being assigned to a `noneType` robot.

Fixed this on the front end and backend:

- Increase the pagination size on the robot panel
- Restrict the logging only to a situation when robot is None and task is unassigned (task.fleet_name is empty)
- Restrict the robot assignment to a situation when only Robot exist and task is already assigned to a robot

Hopefully, a pagination size of 20 is enough to accommodate all the robots for demoing and even if it exceeds, the api server would not throw an error anymore but only log warnings.


Yadunund (approved):
Thanks for the fix, works great now.

PR #477, opened 2021-08-24
fix - error when robot number exceeds pagination

ChawinTan:
Refer to https://github.com/open-rmf/rmf-web/pull/476 for description.

### DIFF
--- a/packages/api-server/api_server/routes/fleets.py
+++ b/packages/api-server/api_server/routes/fleets.py
@@ -68,21 +68,23 @@ class FleetsRouter(FastIORouter):
                 # multiple fleets with the same robot name and there are active tasks
                 # assigned to those robots and the robot states are not synced to the
                 # tasks summaries.
-                if r is None:
+                if r is None and t.fleet_name == "":
                     logger.warn(
                         f'task "{t.id_}" is assigned to an unknown fleet/robot ({t.fleet_name}/{t.robot_name}'
                     )
-                r.tasks.append(
-                    Task(
-                        task_id=t.task_id,
-                        authz_grp=t.authz_grp,
-                        summary=t,
-                        progress=get_task_progress(
-                            t,
-                            app.rmf_gateway().now(),
-                        ),
+
+                elif r and t.fleet_name:
+                    r.tasks.append(
+                        Task(
+                            task_id=t.task_id,
+                            authz_grp=t.authz_grp,
+                            summary=t,
+                            progress=get_task_progress(
+                                t,
+                                app.rmf_gateway().now(),
+                            ),
+                        )
                     )
-                )
 
             return list(robots.values())
 


packages/api-server/api_server/routes/fleets.py
```diff
@@ -68,21 +68,23 @@ async def get_robots(
                 # multiple fleets with the same robot name and there are active tasks
                 # assigned to those robots and the robot states are not synced to the
                 # tasks summaries.
-                if r is None:
+                if r is None and t.fleet_name == "":
```
koonpeng, line 71: Can you explain how this fixes the bug? As far as I see, if `fleet_name` is "", `r` should always be `None` as a robot must always belong to a fleet.

koonpeng:
> This bug is caused by a robot not queried in the list of robots as it is out of scope of the pagination, but exists in the task that it is assigned to, causing the api server to keep throwing an error as the task is being assigned to a `noneType` robot.

I don't quite understand the bug, the tasks query should only return tasks assigned to a robot that is part of the query, so it shouldn't return a `None`robot unless the scenario mentioned in the comments happens, or if there is a bug in the task query.

ChawinTan:
There is nothing wrong with the system actually. It only happens when there are more robots than what is shown in the pagination limit.

Say there are 11 robots but the pagination limit is at 10 robots per page. A task (lets call it `task_11`) gets assigned to robot 11 which is not fetched in this block of code due to the pagination limit:

```
            robots = {
                f"{r.fleet}/{r.name}": r
                for r in await rmf_repo.query_robots(
                    pagination,
                    fleet_name=fleet_name,
                    robot_name=robot_name,
                )
            }
```

Later, while looping the tasks, `task_11` tries to access robot 11 but it is not in the robots array, resulting in a `noneType` error thrown by python and preventing the robot page ui to be loaded even though it is already successfully assigned. 
```
            for t in tasks:
                r = robots.get(f"{t.fleet_name}/{t.robot_name}", None)
                # This should only happen under very rare scenarios, when there are
                # multiple fleets with the same robot name and there are active tasks
                # assigned to those robots and the robot states are not synced to the
                # tasks summaries.
                if r is None:
                    logger.warn(
                        f'task "{t.id_}" is assigned to an unknown fleet/robot ({t.fleet_name}/{t.robot_name}'
                    )
                r.tasks.append(
                    Task(
                        task_id=t.task_id,
                        authz_grp=t.authz_grp,
                        summary=t,
                        progress=get_task_progress(
                            t,
                            app.rmf_gateway().now(),
                        ),
                    )
                )
```
If the number of robots is under the pagination limit, robot page will be loaded again once all the tasks has been assigned. This scenario did not happen back probably because there was no batch upload. When tasks are uploaded by batch, it seems like the task assignment takes quite a while to complete, leading to the robot page crashing when you switch over to robot page when there are tasks still in the pending state.

This is just a quick fix and probably not ideal, but I can't think of another way atm. If you think this pr is not necessary, then I can close it.

koonpeng:
if the task query is working correctly, it should not return tasks assigned to robots not in the query.

It looks like the actual bug is due to a missing `else`, so the `r.tasks.append` statement ends up running even if `r` is `None`. So adding as `else` should be enough to fix it and you don't need to check `t.fleet_name == ""`.

ChawinTan:
Just tested it and the recent changes seem to have fix the issue on. This pr is probably not needed and I'm going to close it. :sweat_smile:  I probably should merge in the changes to throwaway/demo as the the performance is so much better since the changes.

Issue #484, opened 2021-08-30
Batch tasks submitted to dispatcher in reverse order


Yadunund:
It seems that when a batch task file containing tasks [A, B, C] is uploaded, the dashboard submits the tasks to the dispatcher in the reverse sequence, ie, C->B->A. This causes minor inconvenience with scripting scenarios for demos.

koonpeng:
```tsx
  const submitTasks = React.useCallback<Required<TaskPanelProps>['submitTasks']>(
    async (tasks) => {
      if (!tasksApi) {
        throw new Error('tasks api not available');
      }
      await Promise.all(tasks.map((t) => tasksApi.submitTaskTasksSubmitTaskPost(t)));
      handleRefresh();
    },
    [tasksApi, handleRefresh],
  );
```

Since rmf does not support batch tasks, the client is submitting the requests in parallel, the order is probably a byproduct of how the browser handles scheduling of promises. If we want to change this behavior, we will have to submit tasks in sequence which will slow down the process.

Yadunund:
Ah I see. Thanks for explaining why this happens. Is it safe to say that the behaviour is also browser specific, ie, the same file uploaded via dashboards running on Firefox and Chrome may have different outcomes? If so we will make sure to use the same browser to upload th scenario for the demo.

koonpeng:
Currently batch upload is unordered, so we can assume the order may be different everytime. Because we are uploading tasks in parallel, external factors like network latency may change the order as well.

koonpeng:
I believe this is only fixed on the throwaway branch, it remains a question if we should port it to the main branch, use the current parallel impl or wait for open-rmf/rmf_ros2#115 to be implemented first and use actual batch task submissions rather than hacking around it on the client side.

aaronchongth:
This can be achieved using the task scheduler now

PR #489, opened 2021-09-06
submit task sequentially

ChawinTan:

## Self-checks


koonpeng:
This is probably related to #484. Submitting task is designed to be parallel to improve performance. We can implement sequential submission of tasks for now but ideally I think rmf should support batch task submissions. Without batch task support from rmf, we are unable to provide proper error/success response and rollbacks in the case where some of the tasks have invalid inputs or failed for any reason.

Opened an issue in rmf_task open-rmf/rmf_task#40

PR #494, opened 2021-09-08
Feature/fire alarm

ChawinTan:
## What's new

Add fire alarm trigger feature. 

When user toggles the switch button on the map, it sends a boolean value to the api-server backend which subsequently publishes the value to rmf.

When the alarm is triggered, the header bar would turn red, with a title that displays `Emergency Alarm Activated`.


youliangtan:
It works perfectly on my end. :tada:  
One nit-pick comment, is it better if we can label the fire alarm toggle and position it to a better placement ? (now it is floating on the canvas, maybe place it at the top/side bar?)

ChawinTan:
> One nit-pick comment, is it better if we can label the fire alarm toggle and position it to a better placement ? (now it is floating on the canvas, maybe place it at the top/side bar?)

@Yadunund what do you think of you liang's suggestion?

Yadunund:
I agree that we definitely need a label. Let's not worry too much about the placement in this PR. We should discuss that with the figma mockup in mind. But great job implementing this!

ChawinTan:
Added a label to the switch

Yadunund:
The disjointed label and switch looks a bit odd. Can we bundle them into a white box?

image

ChawinTan:
white_background

Added a paper to the background of the Formgroup

Yadunund (approved):
Looks awesome. Thanks for iterating.

Issue #496, opened 2021-09-10
Icons not being loaded


cnboonhan:
`128b23be8240b94c121985b3ca9960d26ee9c341` is the last working commit for icons.

From discussions, its probably due to https://github.com/open-rmf/rmf-web/commit/32638d89fdd39c769af07fb0e2e1b88fc61c1fce#diff-7bcb25a51193443ea006d7e8cc24b8bd504e5d0ce5ab3d15f54e40180b24bcaeR194-R196 not triggering a render on the map.

PR #507, opened 2021-09-13
clone objects to trigger rendering

cnboonhan:
Fixes https://github.com/open-rmf/rmf-web/issues/496 by doing a shallow copy that forces a render.

Naively I am understanding there are some drawbacks to using [shallow copies](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign), but i'm not sure if they are relevant in our case. The alternative deep copy seems to be `JSON.parse(JSON.stringify())`.

It is taking about 1 seconds to load icons now switching tabs away / back to the buildings tab, when before it took minimal time, i'm wondering if that is something that can be fixed.


vallq:
Think we can merge this in for now so that at least we have some visual icons first and address the time to load subsequently :)

koonpeng:
The delays might have something to do with `setInterval` only running on the first timeout.

Issue #495, opened 2021-09-09
TypeError when parsing task batch file


Yadunund:
![image](https://user-images.githubusercontent.com/13482049/132612313-cf725a67-bf12-4d51-b5f9-0c40903a8793.png)

As seen in the screenshot above, when an incorrectly populated batch file is uploaded, the dashboard throws a TypeError. I think we should handle this exception differently. Maybe notify the user that there is a problem with their file as opposed to throwing an error.

PR #509, opened 2021-09-14
Fix - display error message instead of throwing an error when uploading batch task

ChawinTan:

## What's new

Related to: https://github.com/open-rmf/rmf-web/issues/495

Instead of throwing an error within the `checkField` function, display the error in the area where the tasks are displayed whenever there is more than one error in the submitted batch task.

Error_create_task


koonpeng (changes requested):
This doesn't seem very user friendly, the task lists suddenly changed to an error list.

Some suggestion:
Change text color of problem tasks to error color. Focusing on those tasks shows the error fields like in single task mode.

ChawinTan:
Change the color of the task title to red if it has error. Also included a tooltip to show what the errors are.

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -347,6 +351,23 @@ function defaultTask(): SubmitTask {
...
+      {errMsgs.map((msg) => {
+        return <div>{msg}</div>;
+      })}
+    </div>
+  );
+}
+
+export interface SubmitTaskDetails {
+  submitTask: SubmitTask[];
+  errors: { [index: string]: string[] };
```
koonpeng, line 368: type is declared as `string` but it's being accessed with `number` in several places.

koonpeng (changes requested):
It appears that there is no helper text to describe errors currently. So I would recommend to add those, this would improve the ux for both the single task and batch task scenarios.

The catch is that the create task form component would have to perform validation, which is almost overextending it's role as a presentation component. The solution to that would be a accept a validator prop, but it will be hard to describe which error goes into which field. I don't mind adding validation logic into the component, so it depends if you want implement the validator prop and keep the component purely presentational.

koonpeng:
Another suggestion is that since the ui doesn't allow an invalid value to be inputted. We can just stop the loading of batch tasks of the json contains invalid values. i.e. It is the job of the parser to display an error and create task form don't have to do anything.

ChawinTan:
Went with the second option.

- Removed the changes in create tasks form
- show a snackbar error alert when there is an error
- only return the SubmitTask promise array when it is error free

packages/dashboard/src/components/tasks/utils.ts
```diff
@@ -2,46 +2,66 @@ import { SubmitTask } from 'api-client';
 import * as RmfModels from 'rmf-models';
 
 /* istanbul ignore next */
-function checkField(obj: Record<string, unknown>, field: string, type: string): void {
-  if (!Object.prototype.hasOwnProperty.call(obj, field) || typeof obj[field] !== type) {
-    throw new TypeError(`expected [${field}] to be [${type}]`);
+function checkField(
```
koonpeng, line 5: Suggest to remove `errMsgs` param and wrap this in a `checkFile` function. The function will keep its own list of error messages and call `checkFields` with try/catch.

On another node, the try/catch here is useless because you are throwing and immediately catching any exceptions, you can just call `errMsgs.push` without any try catch.

packages/dashboard/src/components/tasks/utils.ts
```diff
@@ -2,46 +2,66 @@ import { SubmitTask } from 'api-client';
...
+  try {
+    if (!Object.prototype.hasOwnProperty.call(obj, field) || typeof obj[field] !== type) {
+      throw new TypeError(`expected [${field}] to be [${type}]`);
+    }
+  } catch (err) {
+    errMsgs.push(err.message);
   }
 }
 
+export interface SubmitTaskDetails {
```
koonpeng, line 20: This shouldn't be needed as `parseTasksFile` should just throw an error on failure (which should be documented). The caller should call this in a try/catch and show an error snackbar if it fails, the error message is available from the thrown error.

koonpeng (changes requested):

ChawinTan:
- remove `errMsgs` array and `SubmitTaskDetails` type
- increase ErrorSnackbar duration to 5s as 2s feels too quick
- Also defined some custom errors to be used instead

packages/dashboard/src/components/tasks/utils.ts
```diff
@@ -1,67 +1,87 @@
 import { SubmitTask } from 'api-client';
 import * as RmfModels from 'rmf-models';
 
+const genericErrors = {
+  checkArray: 'Expected an array of tasks',
+  checkTaskObject: 'Expected task to be an object',
+  defaultError: 'Unknown task type',
```
koonpeng, line 7: Unknown task type is an explicit error and should not be used as a catch-all.

packages/dashboard/src/components/tasks/utils.ts
```diff
@@ -1,67 +1,87 @@
...
+export function parseTasksFile(contents: string): SubmitTask[] {
   const tasks = JSON.parse(contents);
-  let errMsgs: string[] = [];
-  let res: SubmitTaskDetails = { submitTask: [], errors: {} };
   if (!Array.isArray(tasks)) {
-    throw new TypeError('expected an array');
+    throw new TypeError(genericErrors.checkArray);
   }
 
   tasks.forEach((t, i) => {
```
koonpeng, line 58: The way this works doesn't produce helpful error messages, it stops on the first error, and the error message does not easily identify which task contains the error.

I suggest something like this.
```ts
let errMsg = '';
tasks.forEach((task, idx) => {
  try {
    ...checkField();
  } catch (e) {
    errMsg += `Task ${idx + 1}: ${e.message}`;
  }
  try {
    ...checkField();
  } catch (e) {
    errMsg += `Task ${idx + 1}: ${e.message}`;
  }
  ...repeat
});
if (errMsg.length > 0) {
  throw new Error(errMsg);
}
```

To reduce the number of try/catch, `checkField` functions can be refactored to not throw and instead return a non-empty string if there is an error.

`customTaskFileErrors` is also not needed as the `checkField` function already contains all the information needed to return the error message. If the goal is to standardize error messages, you should define a custom error class instead.

```ts
class MyError extends Error {
  constructor(public fieldName: string, public expectedType: string) {
    super();
    this.message = 'my error message';
  }
}
```

ChawinTan:
Apart from the above, also make some changes to the display in `Alert` when there are more then one message

koonpeng (approved):

PR #512, opened 2021-09-20
dashboard: improved markers

koonpeng:
## What's new

https://user-images.githubusercontent.com/45189696/134300890-a9e90541-43c5-42c6-a54e-d3f6f4aac240.mp4

1. Markers automatically change to a fixed size wrt the screen when map is zoomed out.
2. Each marker has a label pointing to it.
    1. Compared to the previous design where the name is part of the marker, this allows better readability, especially for smaller markers like waypoints and doors.
3. The labels automatically avoid collision with each other so that they are still readable even when multiple entities close to each other. This is more apparent when the map is zoomed out, and the new fixed sized markers take up a larger portion of the screen.

## Self-checks


vallq:
I think the labels look good and it's overall a helpful feature! thanks kp
there's only a minor bug that I'm seeing on the dashboard, and it's that the robot marker does not rotate accordingly to the direction it is moving towards. am thinking this can be rectified here or a new ticket can be created to address this

Screenshot 2021-09-30 at 15-30-44 RMF Dashboard

koonpeng:
Should we hide some of the labels by default to remove clutters?

Issue #511, opened 2021-09-15
[ Discussion ] Purpose of Cancel and Ok button in Lift Form?


cnboonhan:
I'm wondering if anyone has any thoughts on the UI for the Lift Form? Apologies if there is a design document I grossly skipped over.
![liftform](https://user-images.githubusercontent.com/12628669/133435170-0cc44c71-9423-417f-8ffb-eb9a09aca21a.png)

* Firstly I noticed the difference between a Task Request  ( on the task panel )and a Lift Request, is that the Lift Request form does not automatically close when submitted. This may make the user confused and click to submit multiple requests.
* The Level field seems to switch after each request to the current level of the Lift. As a result of the combination of this point and point 1, the user can easily accidentally send the lift to their target level, and back to the current level.
* Specifically, what is the difference between the Request and Ok button? they seem to do the same thing. As a result, the user may accidentally send "contradictory" lift requests by trying to use the ok button to close the form.
* What is the use of the Cancel button? Would the expected behavior to send the lift back to the previous level it was on? Or is it equivalent to clicking the cross on the top right of the form?

Example of user accidentally messing up their submission:

https://user-images.githubusercontent.com/12628669/133436020-18273f25-b118-48b0-a601-1d04ab1e5bb8.mp4

ChawinTan:
I think this was my oversight when I shift the lift request form to the `confirmation-dialog` component which is currently the default that we use for anything requires a dialog box. 

Will work on it once I finish whatever I have on hand.

PR #516, opened 2021-09-23
fix - hide dialog actions in lift request form

ChawinTan:

## What's new

Related to #511

Added a `showDialogActions` optional props to show/hide the action buttons. When the props is set to `false` the buttons would be hidden. If it is not specified by default or set to true, then the action buttons would be displayed.

## Self-checks

### DIFF
--- a/packages/react-components/lib/confirmation-dialog.spec.tsx
+++ b/packages/react-components/lib/confirmation-dialog.spec.tsx
@@ -17,22 +17,4 @@ describe('ConfirmDialogActions', () => {
     userEvent.click(root.getByText('OK'));
     expect(onSubmit).toHaveBeenCalled();
   });
-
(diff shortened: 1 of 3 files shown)


koonpeng (changes requested):
A confirmation dialog by design always has an ok/cancel button, or at least some dialog actions. The best solution to make it customisable is to get a dialog actions element as an prop as pass in an empty div, but that still wouldn't fix the problem as it would still draw the dividers and paddings etc.

This seems like a problem with the lift request dialog instead, it shouldn't be rendering an "ok" button if it is already using a confirmation dialog.

ChawinTan:
Removed the Request button in lift request form and modified some texts using confirmation dialog props

koonpeng (approved):

PR #523, opened 2021-10-04
fix: various fixes and performance improvements to markers

koonpeng:
## What's new

* More conditions to exit labels colliding searches.
* Fix markers from other levels being shown (regression).
* Fix type errors in stories.
* Labels are now always on top of all other marker types.

## Self-checks


cnboonhan:
https://user-images.githubusercontent.com/12628669/135840929-1acd9921-3625-44b7-8a3b-c884f20f167d.mp4


Here are the results as I've tried out.
* When waypoints are active, it seems pretty laggy still
* When i deactivate waypoints (40 second mark) performance is very much improved. Interestingly, the text bubbles are no longer hidden. The yellow dots for waypoints disappear, but the text bubbles remain and resize correctly. 

It seems a bit odd to see the text markers even when the waypoints checkbox is removed. However, if this performance can be replicated instead to when the waypoints are active, I think that would be perfect :+1:

koonpeng:
> Interestingly, the text bubbles are no longer hidden. The yellow dots for waypoints disappear, but the text bubbles remain and resize correctly.

Thats not intended, but at least it shows that the problem actually does not lies in the text bubbles. It's probably just due to the standard rendering time and can be fixed with memos.

@cnboonhan Can you test the latest changes again?

cnboonhan:
https://user-images.githubusercontent.com/12628669/136027355-8e153761-803f-4af1-82ef-37d0cdef750b.mp4

There still seems to be a significant difference in "FPS" with or without waypoints, but now the bubbles are hidden correctly :+1: 

Its not functionally a problem for airport_terminal demo but i do wonder how performance will scale

koonpeng:
Can you run the browser and react devtools profiler and share the results? From my runs, it seems like waypoints are no longer the cause of the problem, the most time spent is in the doors and robot state updates, in both the map and the panels.

You can also try commenting out the doors and lift panels and try running with just the map (and vice versa) to see if it affects the performance.

cnboonhan:
Here are profiles of firefox and chrome

[profiling.zip](https://github.com/open-rmf/rmf-web/files/7290320/profiling.zip

lag_firefox
)
lag_chrome

koonpeng:
@cnboonhan please test the latest changes, the main slowdown now should be from robot markers, but I dont see any way to improve the performance without escaping reacting or moving to a canvas-based map.

cnboonhan:
just tested, I think the results are similar as before. I guess we should stick with chrome for larger maps for now. thank you for looking into it!

koonpeng:
just to confirm, performance still tanks only when waypoint layers is enabled? If so then it is most certainly leaflet/firefox/system configuration problem, the react profiler does not show any significant work done on the waypoints overlays anymore.

Issue #504, opened 2021-09-13
dashboard: Interactive map in Robot tab


aaronchongth:
### Description

The same interactive map seen in the Building/Overview(https://github.com/open-rmf/rmf-web/issues/502) tab, should also be applied to the Robot tab.

### Implementation considerations

This will allow users to zoom in, see trajectories of any specific robot as well as center the map onto the robot. (https://github.com/open-rmf/rmf-web/issues/503)

koonpeng:
We should start with a figma prototype first.

aaronchongth:
Prototype on Figma for reference,
![image](https://user-images.githubusercontent.com/5383623/134844202-7f986841-2269-4bb5-bcb3-836a4398b265.png)

aaronchongth:
Some feedback from show-and-tell,
* There wasn't any input as to what should go into the empty panel below the interactive map, consider making the map fill up the entire column.

aaronchongth:
Prototype of Figma is looks great! This task is ready for assignment.

Issue #503, opened 2021-09-13
dashboard: Clicking a component in panels moves interactive map to center on its icon


aaronchongth:
### Status

Per VC, this is blocked by the implementation of different Tabs, 
- [x] https://github.com/open-rmf/rmf-web/issues/504
- [ ] https://github.com/open-rmf/rmf-web/issues/505
- [ ] Implement settings tab (TBC)

but can be started on independently when each tab is completed.

### Description

It is sometimes difficult for users to find the object of interest (robot, door, lift, etc) on the interactive map if the scale of the map is large. Consider centering the interactive map onto the object's icon when the object is clicked on the other panels, in the tabs that the interactive map exists.

### Implementation considerations

* There might be users who do not want this feature, as they have zoomed out to make the whole map visible, this might make them need to zoom out every time they click on an object to check its state. Consider adding a toggle in the settings if this feature should be enabled.

koonpeng:
We should start with a figma prototype first.

vallq:
I think @ChawinTan has built this centering feature in one of the branches previously? Correct me if I am wrong!

But also there would be the design issue of one-click performing 2 different interactions, similar to what we discussed previously for the task panel right, @koonpeng? if clicking on a component shows information on a side panel and performs the centering on the map.

![image](https://user-images.githubusercontent.com/28668944/133952023-5ea75583-3fb8-4c66-b206-15189a1cf885.png)
I foresee it would look something like this? the space below the map is empty for now, perhaps the toggle mentioned could be here, or it can be in the settings drawer

aaronchongth:
> We should start with a figma prototype first.

Agreed! @koonpeng, I just added this line in the self-check, let me know what you think, https://github.com/open-rmf/rmf-web/pull/513

> I think @ChawinTan has built this centering feature in one of the branches previously? Correct me if I am wrong!

Yup! It is in one of the branches and it was previously blocked by the removal of the omni-panel. I think the individual tabs should be a priority, before the feature for centering of icons on itneractive maps can be added in.

aaronchongth:
Per VC, this is blocked by the implementation of different Tabs, 
- [ ] https://github.com/open-rmf/rmf-web/issues/504
- [ ] https://github.com/open-rmf/rmf-web/issues/505

but can be started on independently when each tab is completed.

vallq:
[...]

PR #530, opened 2021-10-21
Feature/add map to robot tab

ChawinTan:
## What's new

Related to #503 and #504  

- Added schedule visualizer into the robot page tab
- Swap the positions of the battery and task progress
- Add zooming and centering when a robot is clicked

~~There is a minor inconsistency currently in the zooming feature. When a robot is assigned a task and on the move~~, ~~you would need to click on the `refresh` button on the panel to get the latest position of the robot before clicking on~~ ~~the robot row, otherwise the map would center to its current position when the robot has already move away.~~

updated-screenshot-robot-tab

## Self-checks


packages/dashboard/src/components/robots/robot-page.tsx
```diff
@@ -1,31 +1,81 @@
...
+  const [fleets, setFleets] = React.useState<Fleet[]>([]);
+  GetFleets(rmfIngress, setFleets);
+  const fleetStatesRef = React.useRef<Record<string, RmfModels.FleetState>>({});
+  SubscribeFleet(sioClient, fleets, fleetStatesRef);
 
+  // robot panel stuff
   const [hasMore, setHasMore] = React.useState(true);
   const [page, setPage] = React.useState(0);
   const [verboseRobots, setVerboseRobots] = React.useState<VerboseRobot[]>([]);
   const fetchVerboseRobots = React.useCallback(async () => {
```
vallq, line 73: would it help if the fetching of the robots states can be toggled with `autorefresh` like the tasks tabs to fix the problem of the position og the centering feature?

vallq (commented):
https://user-images.githubusercontent.com/28668944/139637092-bce52919-62a4-4354-802f-edc93bcc7769.mov

I'm not sure if it's just on my end but there seem to be 2 different panning motions (1 immediate, 1 delayed) when there are no active tasks. if robot1 is currently centered, clicking on robot2 moves the position immediately (no panning motion) but if there are no robots in focus/center, clicking on either entry will show the desired panning motion
is this something that can be kept the same?

also could the mouse icon change to a suitable icon that indicates the entries are clickable?

ChawinTan:
> I'm not sure if it's just on my end but there seem to be 2 different panning motions (1 immediate, 1 delayed) when there are no active tasks. if robot1 is currently centered, clicking on robot2 moves the position immediately (no panning motion) but if there are no robots in focus/center, clicking on either entry will show the desired panning motion
> is this something that can be kept the same?

This will require some digging into the behavior of how `setView` works.

> also could the mouse icon change to a suitable icon that indicates the entries are clickable?

Thanks for pointing that out!

packages/dashboard/src/components/robots/robot-page.tsx
ChawinTan: Sure! Maybe I could create a separate task for this.

ChawinTan:
Turns out the panning motion can be fixed by simply setting the animation option in `setView` to true.

packages/dashboard/src/components/robots/robot-page.tsx
```diff
@@ -37,24 +87,44 @@ export function RobotPage() {
...
-        count: hasMore ? -1 : page * 10 + verboseRobots.length,
-        rowsPerPage: 10,
-        rowsPerPageOptions: [10],
-        page,
-        onChangePage: (_ev, newPage) => setPage(newPage),
-      }}
-      verboseRobots={verboseRobots}
-    />
+    <Grid container className={classes.container}>
+      <Grid item xs={3}>
```
vallq, line 98: What are your thoughts on making the space for the map bigger? 4 or 5, cos I'm thinking if the map is bigger like the clinic or hotel world, a larger space for the map would provide better visualization

image
this is with `xs={5}`

vallq (approved):

packages/dashboard/src/components/robots/robot-page.tsx
ChawinTan: Fixed the positioning issue by calling `fetchVerboseRobots` when clicking the robot.

packages/dashboard/src/components/robots/robot-page.tsx
ChawinTan: Changed the xs value to 4 as the table at 5 looks too small on my laptop. @koonpeng  should we remove the table `min-width` value in favor of a bigger map?

koonpeng: The screenshoot above looks weird, I feel there is no panels that clearly shows the center of attention.

We can reduce the min width, but the table should be the center of attention, so it shouldn't be too small that the map takes away the attention.

Does the head post contains the latest changes? If not, can you update them regularly?

packages/react-components/lib/robots/robot-panel.tsx
```diff
@@ -37,6 +39,7 @@ export interface RobotPanelProps extends React.HTMLProps<HTMLDivElement> {
 export function RobotPanel({
   paginationOptions,
   verboseRobots,
+  leafletMap,
```
koonpeng, line 42: You should not couple `RobotPanel` with a map, the panel should just be the panel itself and not be aware of the state of another map component. Such interactions should be handled by a common parent, i.e. `RobotPage`.

koonpeng (changes requested):

packages/dashboard/src/components/robots/robot-page.tsx
ChawinTan: I think the screen shot above is just to show the bigger map size and not the zooming in. Updated the head post with the latest screenshot

ChawinTan:
- Removed all the `setLeafletMap` and associated props and exported the map from schedule visualizer instead
- other changes in unit test and changing function to hooks

packages/react-components/lib/robots/robot-panel.tsx
```diff
@@ -32,12 +32,14 @@ export interface RobotPanelProps extends React.HTMLProps<HTMLDivElement> {
...
+  onRobotZoom?: (robot: VerboseRobot) => void;
 }
 
 export function RobotPanel({
   paginationOptions,
   verboseRobots,
   fetchVerboseRobots,
+  onRobotZoom,
   ...divProps
 }: RobotPanelProps): JSX.Element {
```
koonpeng, line 44: Just some legacy problems, in keeping the spirit of re-usable components, I think `fetchVerboseRobots` should be replaced with `onRefresh`, and leave it to the parent component to handle the refresh logic.

For now, you can just put a FIXME to refactor it in the future.

packages/dashboard/src/components/schedule-visualizer/index.tsx
```diff
@@ -48,6 +49,7 @@ const TrajectoryUpdateInterval = 2000;
 // of the whole app when it changes.
 const SettingsKey = 'scheduleVisualizerSettings';
 const colorManager = new ColorManager();
+export let mapHandler: React.MutableRefObject<Map<MapProps, L.Map> | null | undefined>;
```
koonpeng, line 52: Avoid global singletons.

packages/dashboard/src/components/schedule-visualizer/index.tsx
```diff
@@ -61,6 +63,7 @@ export interface ScheduleVisualizerProps extends React.PropsWithChildren<{}> {
    * default: 'normal'
    */
   mode?: 'normal' | 'negotiation';
+  zoom?: number | undefined;
```
koonpeng, line 66: Instead of adding this prop, you may consider forwarding the ref of `LMap`. This component already has a ref callback, so you need to manually set the ref

```tsx
export const ScheduleVisualizer = React.forwardRef((props, ref) => {
  return (
    <LMap
      ref={(map) => {
        ...
        ref.current = map;
     }
     </LMap>
    )
});
```

---

Actually, I realized this props is not used?

koonpeng (changes requested):

packages/dashboard/src/components/schedule-visualizer/index.tsx
```diff
@@ -322,7 +309,7 @@ export default function ScheduleVisualizer({
           setLayersUnChecked((prev) => ({ ...prev, [ev.name]: true })),
         );
         registeredLayersHandlers.current = true;
-        mapHandler.current = cur;
+        if (typeof ref === 'function') ref(cur);
```
koonpeng, line 312: You should handle non functional ref as well.

koonpeng (changes requested):

koonpeng (approved):

PR #531, opened 2021-10-21
ui/theme colors

### LINKED ISSUE #520: dashboard: theme colors
**Discussion**

From previous slack discussion and integrating the theme colors with the figma design, we'd have to look into a more manageable (read: eye-friendly 👀) palette of colors for `rmf-light` and `rmf-dark` themes (open-sourced or coming up with one ourselves)

Current palettes used:
`material-ui`'s default theme (note: this is v4's palette, v5 has slight changes to secondary and info colors)
palette-mui
`rmf-dark`
palette-dark
`rmf-light`
palette-light

**Suggestions and Readings**
https://coolors.co/282960-fafafa-ffffff-212121-7aaa74-ec605b-f0bd6a-adbdff-e0e0e0
 https://blog.prototypr.io/how-to-design-a-dark-theme-for-your-android-app-3daeb264637

vallq:
## What's new
fixes #520 

This PR serves to update the dashboard theme colors

`rmf-light`
Screenshot 2021-10-20 at 15-53-41 RMF Dashboard

`rmf-dark`
Screenshot 2021-10-20 at 22-45-07 RMF Dashboard

## Self-checks


koonpeng (changes requested):
The filters on the map is based on the old rmf-dark theme, can you update that as well?

On a related note, we can consider switching to a canvas based map like in traffic editor. That would require us to rewrite many components but it allows us to render a map from a scene graph so that we can control the colors used in the map.

@codebot Is it possible to export some kind of scene graph from the traffic editor, and make the webgl rendering code into a portable library so we can use it on the dashboard?

packages/react-components/lib/confirmation-dialog.tsx
```diff
@@ -77,7 +76,7 @@ export function ConfirmationDialog({
           <Button
             variant="contained"
             type="submit"
-            color="primary"
+            color="secondary"
```
koonpeng, line 79: Any reason why this was changed to secondary?

packages/react-components/lib/doors/door-panel.tsx
```diff
@@ -30,39 +30,50 @@ export interface DoorInfoProps {
...
     color: theme.palette.getContrastText(theme.palette.primary.main),
   },
   buttonGroup: {
+    marginTop: theme.spacing(1),
     display: 'flex',
     justifyContent: 'center',
   },
+  buttons: {
+    backgroundColor: theme.palette.primary.main,
+    color: theme.palette.getContrastText(theme.palette.primary.main),
```
koonpeng, line 71:
```suggestion
    color: theme.palette.primary.main.contrastText,
```

I think this way is slight more performant.

Same goes for all the other places uses `getContrastText`.

packages/react-components/lib/map/label-marker.tsx
```diff
@@ -177,7 +177,7 @@ export function LabelContainer(props: LabelContainerProps): JSX.Element | null {
           height={labelLocation.borderBBox.maxY - labelLocation.borderBBox.minY}
           rx={contentBorderRadius}
           ry={contentBorderRadius}
-          fill={theme.palette.background.paper}
+          fill={theme.palette.info.dark}
```
koonpeng, line 180: Can you explain the rational for changing the color? I feel the info color makes it pop out too much and doesn't look good.

packages/react-components/lib/doors/door-panel.tsx
vallq: sorry I must have missed this out

packages/react-components/lib/confirmation-dialog.tsx
vallq: with the primary color it does not look like a clear CTA button when there are at least 3 buttons so I thought for the submit button to take on the secondary color so that it stands out
image

and also initially the cancel button was outlined with the secondary color, which does not show up clearly for dark mode so I removed the secondary color class
image

packages/react-components/lib/map/label-marker.tsx
vallq: image
oh because in light mode there was a conflict and I thought it was too subtle in dark mode

but ok I can update the text color instead, so it will look like this
image

image

is this better?

packages/react-components/lib/doors/door-panel.tsx
vallq: oh i've been confusing the infrastructure tab changes with these because they do overlap, but yep will update the getContrastTexts!

packages/react-components/lib/map/label-marker.tsx
koonpeng: Seems that the label text color should be fixed to use the theme colors, the light theme should have black text and the dark theme should have white text.

packages/react-components/lib/confirmation-dialog.tsx
koonpeng: > with the primary color it does not look like a clear CTA button when there are at least 3 buttons so I thought for the submit button to take on the secondary color so that it stands out

That looks like a problem with the theme, if we look at the material default theme, the primary color is blue, with a white background, there is no problem of it blending into the background. The way the dark theme colors are chosen makes it have less contrast, if that is a problem then we should change the theme colors, and not the component design.

> and also initially the cancel button was outlined with the secondary color, which does not show up clearly for dark mode so I removed the secondary color class

That looks like a theme problem as well, again comparing it to the default theme, there is no readability problem. I like removing the secondary color though, it makes it blend into the background more, enforcing that the OK button should be the one the user is most interested in clicking. Also the secondary color is not meant to be a "negative" color, so using it as such is probably wrong. Still, the theme's secondary color should be fixed as it should not result in readability problems.

vallq: ok i will revise the theme colors and update this!

packages/react-components/lib/doors/door-panel.tsx
```diff
@@ -68,11 +68,11 @@ const useStyles = makeStyles((theme) => ({
   },
   buttons: {
     backgroundColor: theme.palette.primary.main,
-    color: theme.palette.getContrastText(theme.palette.primary.main),
+    color: theme.palette.primary.contrastText,
```
koonpeng, line 71: These are not properly applied to the button texts, the class is applied to `ButtonGroup` which does not seem to propgate to the buttons.

koonpeng (changes requested):
The panels have broken themes, the background should not be the primary color, it looks ok on the rmf themes because their primary colors are close to the background colors, but we can't make that assumption. If the theme does not have good colors, the components shouldn't be changed to fix the theme.

image

packages/react-components/lib/doors/door-panel.tsx
vallq: Hm okay will look at this and above comment on themes

packages/react-components/lib/map/label-marker.tsx
koonpeng: @vallq I notice something weird, the label border and text should be using the primary color, but in the buggy screenshot above, the label text seems to be using the background color. Maybe something messed up with the styles.

On another note, since the label text is using primary color, it might blend in too much with the background for the rmf theme, if so then there is a problem with the theme and we should look at a different primary color.

koonpeng (approved):

PR #533, opened 2021-10-25
Add option to use postgres

cnboonhan:
This is a small PR to deliver instructions and option to use PostgreSQL on a local installation for rmf-web.

There is a one-time installation and setup as described in the main README.md. Then, to launch a postgres instance, I was thinking the user should manually run
```
npm run start:rmf
npm run start:rmf-server:psql
num run start:react
``` 
The alternative is to make an additional `npm run start:psql` or something, but i feel having to enumerate all the different combinations of possible launch patterns eventually (psql, mysql, auth or without auth) would be too much hassle.

This PR can hopefully help document a basic configuration required for using psql. It is also necessary for implementing aerich database migration, as the version of sqlite that bundles with Ubuntu 20.04 does not implement all the necessary SQL calls for aerich to work properly. [ref](https://stackoverflow.com/questions/6268064/alter-table-drop-column-syntax-error-using-sqlite)

In addition, i tidied up the dependencies as it seemed `numpy` was missing.


cnboonhan:
pushing things along, it seems that when i update Pipfile.lock the version of pylint updates to 2.11.1. This gives an error when running api-server tests. https://github.com/open-rmf/rmf-web/runs/3996938480?check_suite_focus=true#step:6:18

I fixed the test to use the "old" version as specified in pipfile.lock and i think the tests should work now.

koonpeng:
> pushing things along, it seems that when i update Pipfile.lock the version of pylint updates to 2.11.1. This gives an error when running api-server tests. https://github.com/open-rmf/rmf-web/runs/3996938480?check_suite_focus=true#step:6:18

Looks like pylint is not pinned, I don't know what is pipenv's behavior when this happens, for npm, if it should prefer the locked version even if there is a newer version. Maybe you can read the pipenv's docs to find out its behavior, and pin pylint if its required?

koonpeng:
Don't know how it works, but you may try looking at using `pipenv install` (or one of the commands) instead of modifying the pipfile manually.

cnboonhan:
> > pushing things along, it seems that when i update Pipfile.lock the version of pylint updates to 2.11.1. This gives an error when running api-server tests. https://github.com/open-rmf/rmf-web/runs/3996938480?check_suite_focus=true#step:6:18
> 
> Looks like pylint is not pinned, I don't know what is pipenv's behavior when this happens, for npm, if it should prefer the locked version even if there is a newer version. Maybe you can read the pipenv's docs to find out its behavior, and pin pylint if its required?

Yes pinning the version of pylint helps fix this. I'm however stuck with another error regarding the api-server unittests
```
======================================================================
FAIL: test_success_with_valid_token (api_server.test_sio_auth.TestSioAuth)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/ubuntu/rmf-web/packages/api-server/api_server/test_sio_auth.py", line 58, in test_success_with_valid_token
    self.assertTrue(self.try_connect(token))
AssertionError: False is not true
```

This does not happen when running the unittest in main. However, once i add a single asynpg==2.4.0 dev dependency to the Pipfile, some other dependencies seem to shuffle which breaks this test. I have not yet been able to isolate what is the root cause.

koonpeng:
Could be some transitive dependencies causing the problem. If so, that that would mean it can fail in production as well so its good that we discovered it now.

just a minor point, asyncpg should not be pinned because it is a runtime dependency, there could be important security fixes which we will miss out on. Assuming that it uses semver, we should track the major version.

cnboonhan:
it seems like the issue was that the dependency update changed the error string here: github.com/open-rmf/rmf-web/blob/main/packages/api-server/api_server/test_sio_auth.py#L42

from 
```
Connection refused by the server
```

to 

```
HTTPConnectionPool(host='127.0.0.1', port=8000): Max retries exceeded with url: /socket.io/?transport=polling&EIO=4&t=1636083489.7947085 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f985c218610>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

Should we just remove the string compare and default to trying 10 times? Alternatively we can update the try/except to catch this error type.

koonpeng:
Stopping early is preferred so that tests are faster, what do you think is the best solution that supports both versions?

cnboonhan:
i suppose we could check for the substring "ErrNo 111" which should hopefully always be indicative of this error type. for other error types, the tests should fail immediately. 

i suppose it would have been better if the errors were enumerated, but I didnt manage to find any way to do this.

packages/api-server/README.md
```diff
@@ -29,6 +29,10 @@ e.g.
 ```bash
 RMF_API_SERVER_CONFIG='my_config.py' rmf_api_server
 ```
+To run the api-server with PostgreSQL, assuming it has been set up to listen on 127.0.0.1:5432 with user `postgres` and password `postgres`:
+```
+RMF_API_SERVER_CONFIG=api_server/psql_config.py rmf_api_server
```
koonpeng, line 34: I think this needs to be updated with the new path.

packages/api-server/api_server/test_sio_auth.py
```diff
@@ -39,7 +39,9 @@ def try_connect(self, token: Optional[str] = None) -> bool:
                     client.connect(self.base_url, auth=auth)
                     return True
                 except socketio.exceptions.ConnectionError as e:
-                    if str(e) == "Connection refused by the server" and tries < 10:
+                    # Check specifically for connection refused "Errno 111"
+                    # Likely a race condition, wait for server to start
+                    if "Errno 111" in str(e) and tries < 10:
```
koonpeng, line 44: What if we reverse this check? If we know the error caused by bad authentication, we can check for that instead, and retry for all other errors.

koonpeng (changes requested):

packages/api-server/api_server/test_sio_auth.py
cnboonhan: the error message returned when authentication fails is of type ConnectionError with string "One or more namespaces failed to connect" which is not very descriptive. The source is [here](https://github.com/miguelgrinberg/python-socketio/blob/main/src/socketio/client.py#L347-L348), its not specific to authentication errors. Its not obvious to me how to explicitly target authentication errors.

separately, i'm thinking we can replace the "tries" variable with https://github.com/miguelgrinberg/python-socketio/blob/main/src/socketio/client.py#L44-L46, Let me know what you think?

koonpeng: That at least tells us if it is because the socket fails to connect or the socketio protocol fails to connect. We want to retry only when the socket fails, so we should not be using `reconnection_attempt`.

To be fair, the current code doesn't detect authentication error as well, retries when there is a socket error, and just return false when it is any other error (which may not be authentication error, or even socketio errors). If it is easier to detect any socketio error, we can instead return false when it happens and retry for all other errors.

packages/api-server/README.md
cnboonhan: 1ea7f6e7264238858eb53280a3cb46e0bae64c92 thanks for catching this!

packages/api-server/api_server/test_sio_auth.py
cnboonhan: ef1fcfc535ef39bcee3ad712721085b0464ce09a should (hackily, unfortunately) test for this using a string comparison.

koonpeng (approved):

PR #534, opened 2021-10-25
build: switch to openapi generator

koonpeng:
## What's new

I discovered there is a community version of the swagger code generator called "openapi generator", it seems to be more updated, and more importantly, it does not have the typing bugs in the typescript-axios generator. Since there have been no movement on the PRs to fix typings in the swagger generator, this PR switches to the community generator.

This also attempts to use types from `api-client`, which contains the types from the `api-server` (instead of from ROS) whenever possible. There are still many places which uses `rmf-models` because of the lack of enum support (related https://github.com/open-rmf/rmf/issues/108).

Overall, this should improve the type checkings as there will be no need to perform type conversions at many components. In the future, when RMF supports a web-based public api, we should also remove all the legacy generated ros typings, so this serves as a stepping stone for the future work.

## Self-checks


vallq (commented):
I think this is good to go but I've also requested for another set of eyes from @cnboonhan cos I'm not super familiar with the python side of things as well

cnboonhan:
this looks incredibly useful! I realize most files were autogenerated. 

I noticed a `numpy` dependency was missing after doing a full lerna bootstrap and trying to run `generate-openapi.sh`. 

I was also wondering if its worth adding a note to state that stubs for python and many other languages can be generated too, by changing the `-g` flag?

other than that i think this is pretty much perfect to merge 👍

cnboonhan (approved):
this looks incredibly useful! I realize most files were autogenerated.

I noticed a numpy dependency was missing after doing a full lerna bootstrap and trying to run generate-openapi.sh.

I was also wondering if its worth adding a note to state that stubs for python and many other languages can be generated too, by changing the -g flag?

other than that i think this is pretty much perfect to merge 👍

koonpeng:
> I noticed a numpy dependency was missing after doing a full lerna bootstrap and trying to run generate-openapi.sh.

I think that problem happened before, it's due to some ros/rmf packages not installing the deps correctly, we are not using numpy.

> I was also wondering if its worth adding a note to state that stubs for python and many other languages can be generated too, by changing the -g flag?

I think the user can go to the generator's docs if they are interested in generating their own client.

PR #538, opened 2021-11-01
Feature/use aerich

cnboonhan:
## What's new
This PR aims to add database migration tooling using Aerich. This will be helpful in production settings with a persistent databases in the scenario of database schema changes.  It tries to improve on https://github.com/open-rmf/rmf-web/pull/351, this should not require docker and be `pip3 install`-able. Basic instructions are written [here](https://github.com/open-rmf/rmf-web/blob/b54c022e5d8a3b7b0c3f4268e6a2b7b58a6d89ae/packages/api-server/README.md#database-migration).

## Self-checks

## Discussion
Currently, aerich dependency is installed manually. We could specify it as a dependency in the Pipfile. In my mind, in a deployment setting, it seems aerich would be used outside of whatever container / machine is running the api-server / database, so it might still be necessary to do a manual `pip3` installation.

Will explore how to better automate the upgrade / downgrade process, any tips would be much appreciated. One idea seemed to be to get this working with `git revert`.


aaronchongth:
Closing this via https://github.com/open-rmf/rmf-web/pull/654

PR #549, opened 2021-11-22
(feat) - minimal app

ChawinTan:

## What's new

Related to:  #420

updated_minimal_app

I did a very simple version of what is discussed in #420. The destination <b>(Place B)</b> will always be fixed and is disabled in the group of radio buttons (or we can remove it totally?). The loading bay <b>(Place A)</b> can be selected by the operator. There is a submit button for the operator to submit the task.

@vallq and @codebot  would love to hear your inputs.


vallq:
Just took a look at this, and yea it's mostly similar to what I had tried to do previously! I think the destination option can be removed from the radio buttons, and I like the Destination <option> that you added. 

One of the things mentioned previously was to have a consumable config file that dictates what the minimal app would feature (whether it's a form + task progress or a single form alone). Another point of discussion was for the minimal app to also share the same authentication system as the dashboard.

Seems to me that this could have more discussion within the web team too, to see what's possible and should be worked on moving forward.

koonpeng:
stale and out of date

PR #525, opened 2021-10-12
[react-components]: updating the infrastructure components

vallq:
fixes #505 #491

## What's new
- Updating the name of the tab
- Components colors are updated to match the latest figma prototype
- Each cell is sized to the smallest occupiable space given its contents
- Scrolling within each panel to accommodate a large number of cells

Screenshot 2021-10-12 at 11-54-20 RMF Dashboard
Screenshot 2021-10-12 at 11-52-19 RMF Dashboard

## Self-checks

- [x] I have prototyped this new feature (if necessary) on Figma 
- [x] I'm familiar with and follow this [Typescript guideline](https://basarat.gitbook.io/typescript/styleguide)
- [ ] I added unit-tests for new components
- [ ] I tried testing edge cases
- [ ] I tested the behavior of the components that interact with the backend, with an e2e test

## Discussion
final_61650a0c850fb200b1dddbae_832634
I'm slightly bothered by the different sizes of the cells in the clinic example and that is due to the different lengths of the door names, but truncating the words does not seem like a good idea? allowing the words to wrap to the next line will cause more deviation in the alignment of the cells. would appreciate any suggestions or comments on this!

PR #553, opened 2021-11-24
[react-components/dashboard] upgrade react-leaflet and fix map overlays

### LINKED ISSUE #532: Issue on dashboard map rendering when switching floors
## Bug report

This issue occurs when running a multilevel rmf world (e.g. https://github.com/open-rmf/rmf_demos/pull/98). The first level of the `hotel_world`, level `L1` looks fine.

Screenshot from 2021-10-21 15-26-42

However, when switching to L2 or L3, we can see the misalignment of the map as shown below. Even on `L2`, the map still shows `building_maps_msgs` elements (doors, waypoints, robots...) of `L1`.

Screenshot from 2021-10-21 15-27-18

On a side note, there's also some scaling issue with the lift/robots/doors when zooming in and out the canvas.

## Issues
To recap, the issues are listed here:
- [ ] Incorrect transformation of the floorplan png. Further offset/scaling can be applied to the floorplan png via the information provided [in AffineImage](https://github.com/open-rmf/rmf-web/blob/80555f5a2750df4d017625321303b134b004e14e/packages/rmf-models/lib/ros/rmf_building_map_msgs/msg/AffineImage.ts#L7-L10)
- [ ] Scaling of the assets (lifts, doors, robots) on the map for different zoom levels. Assets should depict their actual scale. (if im not wrong on this)
- [ ] Assets rendering when switching floor levels. Assets (also trajectories) from different floors should be filtered out from the map in accordance with the current floor level. This filtering can be done either on the client/server-side. If on the server-side, we might need to add a param `level_name` to the rest API. Similarly for socketio.

vallq:
## What's new
fixes #532 

- The initial version of react-leaflet used was no longer supported and the API broke, hence upgrading was necessary.
- Upgraded to use react-leaflet `v3.1.0` 
The reason for not using the latest `v3.2.2` is because there is an existing incompatibility of `v3.2.0` and above with Webpack 4 which create-react-app uses as outlined [here](https://github.com/PaulLeCam/react-leaflet/issues/883). I wasn't able to resolve the issues with the temporary solutions provided hence next best move is to use `v3.1.0`

- Clicking on the different levels rendered will now show the corresponding icons
map-fix-2021-11-24_17 18 51

## Self-checks

## Others
- Temporal removal of Trajectory Time Control; still trying to figure out how to create a custom control with the new react-leaflet bindings
- Fix ui for panels


aaronchongth:
Per VC, this should be targetting after https://github.com/open-rmf/rmf-web/pull/530 and https://github.com/open-rmf/rmf-web/pull/525 are merged, in order to use the new interactions and API.

aaronchongth:
Per VC, there were breaking changes in the newer release of `react-leaflet`, which will break the click-centering behavior, https://github.com/open-rmf/rmf-web/issues/503. Ideally we would like to have preserve behaviors with each new incoming feature, hence we might want to update the click-centering feature to use the newer API.

https://github.com/open-rmf/rmf-web/issues/558

vallq:
**Update**
Current Progress:
- [x] Updated the map to show the icons on respective levels
- [x] Added the state references for the doors and lifts on the robot page, otherwise, they would not have been receiving any updates

Fixes needed: 
- [ ] Custom control for Trajectory Time Control
- [ ] Visibility of label markers on the map

Trajectory Time Control
- react-leaflet v3 uses custom hooks and factory apis to create custom controls as functional components. the logic of having `createElement` and `updateElement` follows.
- Resources: [custom control layer](https://codesandbox.io/s/goofy-water-13hld?file=/src/controlledLayer.js), [extending tile layer](https://stackoverflow.com/questions/65663826/how-to-extend-tilelayer-component-in-react-leaflet-v3/65713838#65713838), [react-leaftlet docs](https://react-leaflet.js.org/docs/core-api/#high-level-component-factories)

Label Markers
- I'm not sure where the problem lies with this, maybe broken referencing..?

aaronchongth:
closing with https://github.com/open-rmf/rmf-web/pull/644

PR #565, opened 2022-01-03
refactor: backport tasksv2 architecture changes

koonpeng:
## What's new

The problem is that the whole fastapi stack relies on singletons, the api-server tries to avoid them but the new tasksv2 system requires us to run multiple servers under the same process making this avoidance of singletons much harder. This PR changes the api-server to embrace the singleton patterns of fastapi.

Some tests like tasks routes and permissions have to be removed because it uses mocks which are hard to set up in the new architecture, the current tasks system will be replaced anyway so there is not much point in making the mocks work. The permissions tests are technically not needed if permissions are enforced throughout the app (the normal get/post tests should cover authorisations as well), it is not done so now because it is lacking rmf support.

## Self-checks


youliangtan:
Hi @koonpeng , i noticed you are using `datamodel-code-generator` to create the `rmf_api_msgs` class model from the defined schemas .  I like the way how the py model generation is implemented here.

FYI, I also tried to implement it by just exporting the json schemas string merely as python functions. https://github.com/open-rmf/rmf_api_msgs/pull/9. 
Would like to hear your thoughts on this. Instead of generating the pydantic models here in `rmf-web`, is it better if we can run `datamodel-code-generator` in the `rmf_api_msgs` pkg during colcon compilation? So these models will be available in `rmf_api_msgs`.

koonpeng:
I do like upstream provided data models to ease downstream integrations, but I think there are some changes needed in order to make that possible.

* You should export and install the pydantic data models as a python package so that downstream packages can find them.
* pydantic is not the only data model in python, and python may not be the only language used by downstream, so if you want to look into future implementations, there should be a way to export and (optionally) install multiple packages, downstream packages should only need to install only the data models they are using.
* If `rmf_api_msgs` simply generate the models, we do not know if they actually work, so there should be a way to test the models (there have been cases where the generated models are invalid).

youliangtan:
Thanks for the feedback. I have updated the pending `rmf_api_msgs` PR to generate the pydantic data models in the upstream `rmf_api_msgs`. Similar to here, it will run the `datamodel-codegen` script, and compile it as a "python&cpp hybrid package". Thus, downstream packages can just use the modules directly. [example in readme](https://github.com/open-rmf/rmf_api_msgs/blob/feature/schemas-as-py-module/README.md).

To make sure that the generated data models work as intended, we can have some testing scripts (along with CI) in the `rmf_api_msgs` repo. 

Meanwhile, I think the https://github.com/open-rmf/rmf_api_msgs/pull/9 pr should not block the work in this pr. So feel free to treat this as just a discussion.

PR #568, opened 2022-01-14
build: tasksv2/better dev defaults

koonpeng:
## What's new

* Adds `ros_arg` option to config.
  * `RMF_SERVER_USE_SIM_TIME` takes precedence over the config file.
* `npm start` uses disk-backed sqlite3 (app default is still in memory sqlite3)
* `npm start` uses sim time by default (app default is still no sim time)

These defaults, along with the recent change of `clock.py` to use ros time should allow relative `between` query to work as expected.

## Self-checks

### DIFF
--- a/packages/api-server/.gitignore
+++ b/packages/api-server/.gitignore
@@ -1,8 +1,8 @@
 /api_server.egg-info
-/static
 /test_artifacts/
 /.coverage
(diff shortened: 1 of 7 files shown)


mayman99 (approved):
I just tested it and it works as expected.

PR #573, opened 2022-01-27
WIP taskv2 - dashboard

mayman99:
## Self-checks

- [x] Task progress paper looks funny.
- [x] coloring of task logs containers (don't know according to what yet)
- [x] scroll bar in logs and progress
- [x] color task log boxes (yellow log: warnings, delays, etc
red log: errors, manual interventions required, failures, etc)
- [x] Using event name and phase category from the `taskState` in the `taskLog` panel (this info is not part of the log, but the state)
- [x] test the tree with two level dependency events
- [x] Move Timeline to react components  
- [ ] Override TimeLineDot default colors using by the theme colors (Don't know how to do this yet if someone has an idea, it seems a very minor thing at the moment so I'm not going to invest more time into it)
- [ ] remove duplicates events in the tree (should appear once in case of dependency) 
- [ ] scale down events in the tree as it goes down

## Discussion
To test:
`ros2 launch rmf_demos_ign office.launch.xml`
for api server
`npm run start`
for dashboard
`npm run start:react`
for task
`ros2 launch rmf_demos office_conflict.launch.xml`

Then you should see something like this

Screenshot from 2022-01-28 00-09-20


mayman99:
Hey @koonpeng could you give me your opinion about if I should move this code to `react-components` or leave it in the dashborad?

mayman99:
After going through react-components package, I believe this should be refactored and moved there, I'll start working on this now.

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -48,47 +47,99 @@ const StyledTimeLine = styled((props: TimeLinePropsWithRef) => <Timeline {...pro
...
+    const deps = eventsStates[child] ? eventsStates[child].deps : [];
+    return (
+      <TreeItem
+        key={`event-${child}`}
+        nodeId={`parent-${eventsStates[child].name}`}
+        label={eventsStates[child].name}
+      >
+        {deps?.map((id) => {
+          if (eventsStates[id].deps?.length) {
+            nestedEvents(eventsStates, id);
```
koonpeng, line 61: missing `return`?

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -48,47 +47,99 @@ const StyledTimeLine = styled((props: TimeLinePropsWithRef) => <Timeline {...pro
...
+            );
+          }
+        })}
+      </TreeItem>
+    );
+  }
+}
+
+function colorDot(eventsStates: { [key: string]: EventState } | undefined) {
+  let standby: Boolean = true;
```
koonpeng, line 78: This causes lint error
```
Don't use `Boolean` as a type. Use boolean instead
```

---

Which lead me to realize that the linter is not ran on the CI :sweat: . You can run it manually with `npm run lint`.

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -48,47 +47,99 @@ const StyledTimeLine = styled((props: TimeLinePropsWithRef) => <Timeline {...pro
...
+  for (const id in eventsStates) {
+    const event: EventState = eventsStates[id];
```
koonpeng, line 80: Since you are not using `id` in any other meaningful way, prefer `for...of` instead.

```ts
  for (const event of eventsStates) {
```

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -48,47 +47,99 @@ const StyledTimeLine = styled((props: TimeLinePropsWithRef) => <Timeline {...pro
...
+    if (event.status === 'underway') {
+      return 'success' as const;
```
koonpeng, line 82: I don't quite understand this cast, it seems redundant as `'success'` is already a valid return type (assuming the return is meant to be a string)

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -48,47 +47,99 @@ const StyledTimeLine = styled((props: TimeLinePropsWithRef) => <Timeline {...pro
...
-          </TimelineContent>
-        </TimelineItem>
-      ))}
+      {phases.map((phase, idx) =>
+        phase.events ? (
+          <TimelineItem key={phase.id}>
+            <TimelineOppositeContent color="text.secondary">
+              <Typography variant="overline" color="textSecondary" style={{ textAlign: 'justify' }}>
+                {phase.original_estimate_millis
+                  ? format(new Date(phase.original_estimate_millis * 1000), "hh:mm aaaaa'm'")
```
koonpeng, line 107: `Date` constructor takes a unix millis, so there should be no need to mutiply by 1000.

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -48,47 +47,99 @@ const StyledTimeLine = styled((props: TimeLinePropsWithRef) => <Timeline {...pro
...
+                  : null}
+              </Typography>
+            </TimelineOppositeContent>
+            <TimelineSeparator>
+              <TimelineDot
+                color={colorDot(phase.events)}
+                sx={{
+                  margin: '0.5px',
```
koonpeng, line 115: Prefer theme units, also it is weird to specify such a small margin.

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -48,47 +47,99 @@ const StyledTimeLine = styled((props: TimeLinePropsWithRef) => <Timeline {...pro
...
-                      label={event.name || 'unknown'}
-                    ></TreeItem>
-                  ))
-                : null}
-            </TreeView>
-          </TimelineContent>
-        </TimelineItem>
-      ))}
+      {phases.map((phase, idx) =>
+        phase.events ? (
```
koonpeng, line 102: Are you sure we should hide phases without events?

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -48,47 +47,99 @@ const StyledTimeLine = styled((props: TimeLinePropsWithRef) => <Timeline {...pro
...
+              </Typography>
+            </TimelineOppositeContent>
+            <TimelineSeparator>
+              <TimelineDot
+                color={colorDot(phase.events)}
+                sx={{
+                  margin: '0.5px',
+                }}
+              />
+              {idx < phases.length - 1 && <TimelineConnector />}
```
koonpeng, line 118: My understanding is that we should only render `TimelineConnector` when it is not the last phase, if so this needs to take into account phases which are hidden.

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -48,47 +47,99 @@ const StyledTimeLine = styled((props: TimeLinePropsWithRef) => <Timeline {...pro
...
+              ></TreeItem>
+            );
+          }
+        })}
+      </TreeItem>
+    );
+  }
+}
+
+function colorDot(eventsStates: { [key: string]: EventState } | undefined) {
```
koonpeng, line 77:
```suggestion
function colorDot(eventsStates: { [key: string]: EventState } | undefined): TimelineDotProps['color'] {
```

More specific return type to avoid confusions in the future.

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -48,47 +47,99 @@ const StyledTimeLine = styled((props: TimeLinePropsWithRef) => <Timeline {...pro
...
+                }}
+              />
+              {idx < phases.length - 1 && <TimelineConnector />}
+            </TimelineSeparator>
+            <TimelineContent>
+              <TreeView
+                defaultCollapseIcon={<ExpandMoreIcon />}
+                defaultExpandIcon={<ChevronRightIcon />}
+              >
+                {Object.values(phase.events).map((event) => (
```
koonpeng, line 125: I think we can use `NestedEvents` here, passing `phase.events` to it.

packages/dashboard/src/components/tasks/task-logs.tsx
```diff
@@ -1,33 +1,48 @@
...
       flex: '0 0 auto',
+      maxHeight: '95%',
+      overflow: 'auto',
     },
   }),
 );
 
 export function TaskLogs(props: TaskLogProps) {
   const { taskLog } = props;
+  const { taskState } = props;
```
koonpeng, line 29:
```suggestion
const { taskLog, taskState } = props;
```

You can also do it directly on the function params

```ts
export function TaskLogs({ taskLog, taskState }: TaskLogProps) {
```

packages/dashboard/src/components/tasks/task-logs.tsx
```diff
@@ -1,33 +1,48 @@
...
   const theme = useTheme();
   const phaseIds = taskLog.phases ? Object.keys(taskLog.phases) : [];
+
+  function MapEventColor(event: LogEntry) {
```
koonpeng, line 33: functions should be camelCase

packages/dashboard/src/components/tasks/task-logs.tsx
```diff
@@ -52,36 +70,50 @@ export function TaskLogs(props: TaskLogProps) {
                     <div
                       style={{
                         marginTop: theme.spacing(1),
-                        backgroundColor: theme.palette.success.light,
+                        backgroundColor: MapEventColor(event),
                         padding: theme.spacing(1),
+                        borderRadius: '6px',
```
koonpeng, line 75: prefer theme units

packages/dashboard/src/components/tasks/task-page.tsx
```diff
@@ -43,7 +43,7 @@ export function TaskPage() {
         undefined,
         11,
         page * 10,
-        '-unix_millis_start_time',
```
koonpeng, line 46: Why is this changed?

packages/dashboard/src/components/tasks/task-panel.tsx
```diff
@@ -164,8 +187,10 @@ export function TaskPanel({
       return [];
     }
     if (selectedTask) {
-      const logs = await tasksApi.getTaskLogTasksTaskIdLogGet(selectedTask.booking.id);
+      const logs = await tasksApi.getTaskLogTasksTaskIdLogGet(selectedTask.booking.id, '0,999999');
```
koonpeng, line 190: `999999` is Thu Jan 01 1970 00:16:39, should probably put a bigger number.

koonpeng (changes requested):

packages/react-components/lib/tasks/task-timeline.tsx
mayman99: It used to not accept the string type when it was part of the `dashboard package`, but now after moving it to react components and building it, it works without casting, no idea why. but I just removed the casting.

packages/react-components/lib/tasks/task-timeline.tsx
mayman99: Nope I'm note sure, it was not specified in the figma if we should or should not.
Do you think we should show them?

packages/dashboard/src/components/tasks/task-page.tsx
mayman99: https://osrf.slack.com/archives/C01ECP3GBTQ/p1641831025038800

cnboonhan:
Here are some hopefully useful console logs of the event messages being rendered: 
image

To reproduce:
```
 ros2 launch rmf_demos_ign office.launch.xml headless:=1 server_uri:=ws://localhost:8001 use_sim_time:=true finishing_request:=charge
RMF_SERVER_USE_SIM_TIME=true rmf_api_server
npm run start:react
ros2 run rmf_demos_tasks direct_patrol --robot tinyRobot1 -s supplies -f coe -n 3  -F tinyRobot --use_sim_time
```

Inspecting these values, i seem to be understanding the values in these fields as "duration for event to complete". In which case, this line might not be right https://github.com/open-rmf/rmf-web/blob/4864c15106213f5e10e7f0f26c39b96c0ff8f429/packages/dashboard/src/components/tasks/task-logs.tsx#L104 as its treating this "time-delta" as unix time.

packages/dashboard/src/components/tasks/task-page.tsx
koonpeng: Yeah, `unix_millis_start_time` should be working now. Having it undefined means unspecified sorting.

mxgrey:
> But the name "unix_millis_time" seems to suggest it should be in unix time.

Yes, `unix` means it's a time-point within the "UNIX Epoch". `millis` means it's a count of milliseconds.

> Inspecting these values, i seem to be understanding the values in these fields as "duration for event to complete".

Could you elaborate on what's giving you that impression? The `unix_millis_time` field in the log entries is meant to be a timestamp indicating the point in time that the log entry was created. If something has led you to assume differently, maybe we should change the way information is being presented.

packages/react-components/lib/tasks/task-timeline.tsx
koonpeng: I'm not sure what kind of phases will have no events, and if this will be a good heuristics to hide possibly uninteresting data from the user, maybe @mxgrey can give better insights.

cnboonhan:
my impression was formed just because the values are pretty small, I was expecting larger values for real time dates. I think this could be due to me running in sim time, which produces values that look small. I don't know if this explains why the time intervals between events are so great, though.

Unfortunately I am unable to inspect the full event logs if I am not running on sim time, while using simulation.

packages/react-components/lib/tasks/task-timeline.tsx
mayman99: so if we are hiding empty phases, this shouldn't be necessary, right?

packages/dashboard/src/components/tasks/task-page.tsx
mayman99: Cool, didn't know it was fixed already, reverting it then.

mxgrey:
Yeah, the small values are because of sim time.

The time intervals appear to be every ~10-15 seconds. That seems pretty normal for me. This log entry will only get issued when the robot replans its itinerary. I should change the log message to reflect that better.

mayman99:
@cnboonhan atm, it doesn't automatically renders new logs, so its intended (because we need to make a socketio request first to get new data, and this doesn't happen automatically at the moment)

packages/react-components/lib/tasks/task-timeline.tsx
mayman99: This small margin was the one I test that resulted with the Dot being connected to the separator to look like the figma figure.
I've removed it for now, I'll add it later while fixing the other small visual issues.

packages/react-components/lib/tasks/task-timeline.tsx
mayman99: I tried this but kept giving me a compilation error somehow

packages/react-components/lib/tasks/task-timeline.tsx
```diff
@@ -48,47 +47,99 @@ const StyledTimeLine = styled((props: TimeLinePropsWithRef) => <Timeline {...pro
...
+                key={eventsStates[child].name}
+                nodeId={`child-${id}`}
+                label={eventsStates[id].name}
+              ></TreeItem>
+            );
+          }
+        })}
+      </TreeItem>
+    );
+  }
```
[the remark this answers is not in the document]
mxgrey: This is fixed by 89b70abae369fca577405d74409d165b707f9a96

mayman99:
thanks @mxgrey for the cleanup, the timeline looks much nicer now!
I'm writing a small test for `task-logs` at the moment.

mxgrey (approved):
I've tested out the new dashboard, and it's generally working very nicely with the task system redesign changes.

We need the changes in this PR in order for the dashboard to work with the new task system. There may still be some loose threads to tie up, but we can save those for future PRs.

Issue #575, opened 2022-02-03
Pruning of react-component fields that no longer work in redesign_v2


cnboonhan:
I would like to propose the removal of the following components in the dashboard that no longer seem to be propagated correctly in `redesign_v2`:

Checking / unchecking these checkboxes do not seem to visually change anything
![image](https://user-images.githubusercontent.com/12628669/152300302-23707a59-2d96-47d3-950d-b466ade7b400.png)

Start Location and Destination are always empty here. Active Task Duration is also not populated:
![image](https://user-images.githubusercontent.com/12628669/152300440-d7f0b412-08fa-43f8-8cd2-824bb9b92133.png)

Location and Destination are undefined here. Task progress is also always at 11%:
![image](https://user-images.githubusercontent.com/12628669/152300647-9ee7bcae-1ad0-42f0-a426-a1ed558b6e46.png)

@koonpeng  @aaronchongth  @mxgrey ?

mxgrey:
How was progress being measured previously? I'm not sure there's any meaningful way to measure progress, tbh. Best case scenario would be to base it off of time estimates, but then we could see progress slide backwards due to delays.

koonpeng:
> How was progress being measured previously? I'm not sure there's any meaningful way to measure progress, tbh. Best case scenario would be to base it off of time estimates, but then we could see progress slide backwards due to delays.

iirc, it was using a super naive approach of `(current_time - start_time) / (est_end_time - start_time)`.

mxgrey:
I think we have enough information now to go back to that definition of progress, at least until we can think of something more meaningful.

cnboonhan:
closing this to focus on merging https://github.com/open-rmf/rmf-web/pull/581

PR #576, opened 2022-02-04
Prototype/tasks v2 prune robot panel

cnboonhan:
Removes non-useful fields and fixes data propagation for other fields, according to discussion https://github.com/open-rmf/rmf-web/issues/575.

Wanted to iterate once before cleaning up dangling code and references.

image

### DIFF
--- a/packages/react-components/lib/robots/robot-table.tsx
+++ b/packages/react-components/lib/robots/robot-table.tsx
@@ -13,7 +13,7 @@ import {
   Typography,
   styled,
 } from '@mui/material';
-import type { RobotState } from 'api-client';
+import type { RobotState, TaskState } from 'api-client';
 import { Refresh as RefreshIcon } from '@mui/icons-material';
 import React from 'react';
 
@@ -60,44 +60,11 @@ const StyledPaper = styled((props: PaperProps) => <Paper {...props} />)(({ theme
 
 interface RobotRowProps {
   robot: RobotState;
+  fetchSelectedTask?: (taskId: string) => Promise<TaskState | undefined>;
   onClick: React.MouseEventHandler<HTMLTableRowElement>;
 }
 
-const returnLocationCells = (robot: RobotState) => {
-  // const taskDescription = robot.tasks[0].summary.task_profile.description;
-  // switch (taskTypeToStr(taskDescription.task_type.type)) {
-  //   case 'Loop':
-  //     return (
-  //       <>
-  //         <TableCell>{taskDescription.loop.start_name}</TableCell>
-  //         <TableCell>{taskDescription.loop.finish_name}</TableCell>
-  //       </>
-  //     );
-  //   case 'Delivery':
-  //     return (
-  //       <>
-  //         <TableCell>{taskDescription.delivery.pickup_place_name}</TableCell>
-  //         <TableCell>{taskDescription.delivery.dropoff_place_name}</TableCell>
-  //       </>
-  //     );
-  //   case 'Clean':
-  //     return (
-  //       <>
-  //         <TableCell>-</TableCell>
-  //         <TableCell>{taskDescription.clean.start_waypoint}</TableCell>
-  //       </>
-  //     );
-  //   default:
-  return (
-    <>
-      <TableCell>-</TableCell>
-      <TableCell>-</TableCell>
-    </>
-  );
-  // }
-};
-
-function RobotRow({ robot, onClick }: RobotRowProps) {
+function RobotRow({ robot, fetchSelectedTask, onClick }: RobotRowProps) {
   const getRobotModeClass = (robotMode: string) => {
     switch (robotMode) {
       case 'emergency':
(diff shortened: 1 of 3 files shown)


packages/react-components/lib/robots/robot-table.tsx
```diff
@@ -60,44 +60,11 @@ const StyledPaper = styled((props: PaperProps) => <Paper {...props} />)(({ theme
 
 interface RobotRowProps {
   robot: RobotState;
+  fetchSelectedTask?: (taskId: string) => Promise<TaskState | undefined>;
```
koonpeng, line 63: I would prefer if `react-components` don't require fetching of data. This can lead of some usability problems, for example, imagine if the robot detail panel also requires fetching of data, then the robot page would fetch the same data multiple times, with possibly different results each time.

Instead, it would be better to wrap the `RobotState` and its current `TaskState`, hypothetically called `RobotData` (any better name which doesn't conflict with rmf model names?)

```tsx
interface RobotData {
  state: RobotState;
  current_task_state?: TaskState;
}
```

koonpeng (changes requested):
Can you also update the tests to check if the fields are rendered properly?

packages/react-components/lib/robots/robot-table.tsx
cnboonhan: there was a `VerboseRobot` that might no longer be used as is now, perhaps we can reuse that?

cnboonhan: i just realized the fetching of data already exists in robot-table and robot-panel: the fetchSelectedTask is defined in robot-page and passed into all the robot related react-components. Should i move fetchSelectedTask out of all the react-components, in that case?

koonpeng: > i just realized the fetching of data already exists in robot-table and robot-panel: the fetchSelectedTask is defined in robot-page and passed into all the robot related react-components. Should i move fetchSelectedTask out of all the react-components, in that case?

Fetch data will be needed at some point, but better to put it in dashboard. Components here should be re-usable and having them handle fetching of data puts an assumption on how the app manages data.

> there was a VerboseRobot that might no longer be used as is now, perhaps we can reuse that?

Naming things is always hard, to me, `VerboseRobot` doesn't exactly explain what it represents. On another note, what if suppose that the robot row needs `RobotState` and `TaskState`, and the robot detail also needs `TaskLog`, how should be name the interfaces?

One way to avoid this naming conflict is to not name them in the first place, but that will hurt readability, e.g.

```ts
export interface RobotRowProps {
  data: { state: RobotState, current_task: TaskState }
  ...
}

export interface RobotDetailProps {
  data: { state: RobotState, current_task_state: TaskState, current_task_log: TaskLog }
  ...
}
```

Another way is to drop the reliance on rmf models, e.g.

```ts
export interface RobotRowProps {
  name: string,
  active_task_duration: number,
  battery: number,
  ...
}
```

Theoricatically, this would be the most robust approach, unused properties will not cause needless re-renders (with the current approach, e.g. if the robot location changes, it will cause a needless re-render even though the result is the same since we are not showing the location) and things will not break when rmf models change (it will still break in dashboard, but at least react-components will not be affected).

cnboonhan: i went ahead and used "verboserobot" but it should be straightforward to refactor this name

cnboonhan:
I'm sort of stuck trying to get the state field in robot page to refresh automatically, I tried briefly to implement the socketio pull method for subscribing to tasks, but i think i was not doing it right. further, i think this socketio task subscription will have been done in a parallel PR by @mayman99 , so i think ill hold off that for now to avoid merge conflict.

koonpeng:
The dashboard tests are failing due to an empty test file, can you at least add a superficial test with a `FIXME` to do real testing?

mayman99:
Okay, sounds good, I was trying to fix a visual thing (when you expand the tree in #573 the next element shift to the left a bit), but I'm guessing we could leave this visual thing to later.

mayman99:
Then I'll will give the PR current state one more iteration and then ping you :+1: .

mayman99:
I can try to have a look then

mayman99:
So the reason why it doesn't compile, it is [the commit](https://github.com/open-rmf/rmf-web/commit/eb6da2869fe991a58212ea7184fd809b0d9d8f0f) of allowing multiple subscribers, I tried to adapt the dashboard to that change but I couldn't find a clean way for now.
To compile I just reverted the `unsubscribe` function in `api-client` to before [this commit](https://github.com/open-rmf/rmf-web/commit/eb6da2869fe991a58212ea7184fd809b0d9d8f0f) and it went through, I didn't push this change because maybe someone have an idea of how to solve the error 
for ref: https://github.com/open-rmf/rmf-web/blob/59b19f8443b6d08be0c15d40c34fbe2d8c4b7377/packages/dashboard/src/util/common-subscriptions.ts#L28-L36

mayman99:
Also I get the error `Unhandled Rejection (TypeError): Cannot set properties of undefined (setting '_leaflet_pos')` when I try to click on a robot in the robot panel page, it used to work last time I tested it a few days ago, so could be sth related to last merge

cnboonhan:
i'll take a look as well! so sorry was on PTO before. :eyes:

PR #577, opened 2022-02-08
support multiple subscriptions

koonpeng:
## What's new

Keep tracks of number of ongoing subscriptions to a room so that `unsubscribe` is emitted only when the last subscriber has unsubscribed.

## Self-checks

- [ ] I have prototyped this new feature (if necessary) on Figma 
- [ ] I'm familiar with and follow this [Typescript guideline](https://basarat.gitbook.io/typescript/styleguide)
- [ ] I added unit-tests for new components
- [ ] I tried testing edge cases
- [ ] I tested the behavior of the components that interact with the backend, with an e2e test

PR #581, opened 2022-02-15
Prototype/tasks v2 2 merge prune robot panel

cnboonhan:
This PR tracks a merge of various work into a single branch, and tracks the bugs that are still present that should be resolved before merging to `prototype/tasks_v2-2`. After that merge, the follow up target would be to merge into main.

Features of this branch to be merged into `prototype/tasks_v2-2`
* Includes robot panel edits to remove redundant / unused components: https://github.com/open-rmf/rmf-web/pull/576
* Includes event logs as described here: https://github.com/open-rmf/rmf-web/pull/573
* ~~Does *not* include bug fixes for multiple subscriptions as yet, which is present in main: https://github.com/open-rmf/rmf-web/pull/577~~ This has been reverted and includes the bugfix for multiple subscriptions

There are a few bugs/user "gotchas" as of now
* Sending delivery requests from the create-task panel fails as we cannot specify a payload. This could be resolved with another field in the create-task panel 
```
[full_control-12] [INFO] [1644943525.937784621] [tinyRobot_fleet_adapter]: [Bidder] Received Bidding notice for task_id [delivery.dispatch-0]
[full_control-12] [ERROR] [1644943525.938383205] [tinyRobot_fleet_adapter]: Received a request description for [delivery] with an invalid format. Error: At /dropoff/payload of "" - no subschema has succeeded, but one of them is required to validate
[full_control-12]
[full_control-12] Request:
[full_control-12] {
[full_control-12]   "dropoff": {
[full_control-12]     "handler": "coke_ingestor",
[full_control-12]     "payload": "",
[full_control-12]     "place": "hardware_2"
[full_control-12]   },
[full_control-12]   "pickup": {
[full_control-12]     "handler": "coke_dispenser",
[full_control-12]     "payload": "",
[full_control-12]     "place": "pantry"
[full_control-12]   }
[full_control-12] }
[rmf_task_dispatcher-10] [WARN] [1644943528.137506720] [rmf_dispatcher_node]: Dispatcher Bidding Result: task [delivery.dispatch-0] has no submissions during bidding. Dispatching failed, and the task will not be performed.
[rmf_task_dispatcher-10] [ERROR] [1644943528.137573326] [rmf_dispatcher_node]: No submission error[1]: {"category":"Invalid request format","code":5,"detail":"At /dropoff/payload of \"\" - no subschema has succeeded, but one of them is required to validate\n"}
[schedule_visualizer-5] [2022-02-16 00:45:28] [error] handle_read_frame error: websocketpp.transport:7 (End of File)
[schedule_visualizer-5] [INFO] [1644943528.470067827] [rmf_visualization_schedule_data_node]: [TrajectoryServer] Disconnected with a client
[schedule_visualizer-5] [INFO] [1644943530.202191853] [rmf_visualization_schedule_data_node]: [TrajectoryServer] Connected with a client
[full_control-12] [INFO] [1644943535.903235009] [tinyRobot_fleet_adapter]: Replanning for [tinyRobot/tinyRobot2] because of an interruption
[full_control-12] [INFO] [1644943535.904153263] [tinyRobot_fleet_adapter]: Replanning for [tinyRobot/tinyRobot1] because of an interruption
[gzserver-14] [INFO] [1644943535.909809612] [slotcar_tinyRobot2]: tinyRobot2 received a path request with 2 waypoints
[gzserver-14] [INFO] [1644943535.912172978] [slotcar_tinyRobot1]: tinyRobot1 received a path request with 2 waypoints

```

* Unsuccessful ( unassigned ) tasks end up on the task panel anyway, with an "unknown" status, rather than failed: 
image

* Running under `use_sim_time=true`, we need to manually specify the start time of each task on the create-task panel as 0. Otherwise the panel submits the task start time as unix time, which causes task to be queued far in the future
* Running under `use_sim_time=true`, the current approach to calculating progress is inaccurate
image

* Task cancellation doesn't work, but clicking the button in the task panel reports success. There might be underlying issues as well in the RMF backend, issue here: https://github.com/open-rmf/rmf_ros2/issues/175
* ~~Tests are borked~~ Tests are less borked by removing many of them, but need to fix them ASAP in a folllow up

Otherwise, submission of Loop Requests under `use_sim_time:=false` seems to be working well, seen in this video.
https://user-images.githubusercontent.com/12628669/154110886-747f68f0-d36d-4a8c-943f-aa3ac9fd23fc.mp4

Notably, these edits may not be ideal, but were necessary to handle some validation errors. Do advise if these are not the best solution:
https://github.com/open-rmf/rmf-web/pull/581/files#diff-f8cf7ad0d839656743c8f01b8df95728fd6921717927d8958b22ef11208d92c4L22-R23

For the following error:
```
line 127, in serialize_response
    raise ValidationError(errors, field.type_)
pydantic.error_wrappers.ValidationError: 2 validation errors for TaskDispatchResponseItem
response -> success
  field required (type=value_error.missing)
response -> state
  field required (type=value_error.missing)
```


koonpeng (changes requested):
This undos a previous PR which is needed to fix other bugs, the build error should be fixed now you can do undo the undo.

koonpeng:
> Running under use_sim_time=true, we need to manually specify the start time of each task on the create-task panel as 0. Otherwise the panel submits the task start time as unix time, which causes task to be queued far in the future

I believe that is the correct behavior. The client always submits requests in unix time, and rmf always processes requests in unix time as well, this is same regardless if it is using sim time or not. Tasks are queued far in the future because rmf's time starts at 0.

> Running under use_sim_time=true, the current approach to calculating progress is inaccurate

The calculation is borked because it uses the client's time, actually this is a problem even if sim time is not used because there is no guarantee that the client's and rmf's time is synced. It won't be accurate as long as the client is doing any calculation based on the time. We can have rmf provide the progress instead, which will always ensure a correct estimate, but it would only "tick" when a new task state comes in.

cnboonhan:
Summary of to-do after merging:

- [ ] Modify create-task panel to add sku, quantity inputs for Delivery Requests
- [ ] Unsuccessful dispatches should be displayed as "failed" on task-panel instead of "unknown"
- [ ] In the longer term, figure out a way make task progress calculations independent of the client time
- [ ] Make task cancellations work, after this possible bug is resolved: https://github.com/open-rmf/rmf_ros2/issues/175

PR #582, opened 2022-02-16
Prototype/tasks v2 2 merge to main

cnboonhan:
This PR tracks the fixes needed to be done before merging to main
 
Summary of to-do ( or ASAP after merging ):

- [x] Modify create-task panel to add sku, quantity inputs for Delivery Requests ( relevant: https://github.com/open-rmf/rmf-web/pull/581#issue-1138962123 , https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/schemas/event_description__payload_transfer.json )
- [x] ~~Unsuccessful dispatches should be displayed as "failed" on task-panel instead of "unknown"~~ (thanks @mayman99 ) it seems the status field is empty when there is an error in the submission of task ( for example, wrongly named waypoint ). Which then corresponds to "unknown". As long as "unknown" is understood to mean precisely the situation where something shouldn't have happened, it should be ok at least for now?
image
- [x] Make task cancellations work, after this possible bug is resolved: https://github.com/open-rmf/rmf_ros2/issues/175

To do:
- [x] In the longer term, figure out a way make task progress calculations, and task request timestamps, independent of the client time. This might involve modifying the api server to serve time ( relevant: https://github.com/open-rmf/rmf-web/pull/581#issuecomment-1041030478 )
- [x] Input validation to prevent invalid inputs from the create-task form. Task form might be a bit buggy as well.
- [x] Auto refresh on logs (edited)
- [x] Submit batch task using form. (edited)
  - Simple validation is done through ajv, rmf needs to provide concrete task request schemas in order to perform full validation.
- [ ] Fix lint errors in react-component tests
  - components used by reporting is causing some of the lint errors, fixing these will require updating reporting to work with taskv2 as well. Alternative is to remove reporting from this repo altogether and move it to it's own repo.
- [ ] Restore various tests that were commented out 
  - Tests regarding components in robot pages are done.
  - Components in tasks page are still missing tests.
  - e2e tests need to be updated as well.
- [ ] Fix bug where notification of failed submissions from other interfaces shows up on the dashboard.
- [ ] Check that no intended edits were lost at sea :ocean: 
- [ ] Clean up all the code that is no longer required


mayman99:
regarding point 2

>  Unsuccessful dispatches should be displayed as "failed" on task-panel instead of "unknown" (thanks @mayman99 ) it seems the status field is empty when there is an error in the submission of task ( for example, wrongly named waypoint ). Which then corresponds to "unknown". As long as "unknown" is understood to mean precisely the situation where something shouldn't have happened, it should be ok at least for now?

I'm thinking of adding  check in the `tasksTable`, and if we have a task that doesn't have a status we we prompt sth that says task submission failed, and not add it in the `tasksTable`. 

A problem I see with this is that if someone failed to submit a task via another interface, the dashboard will still get the `fail to submit` prompt as well.

cnboonhan:
https://user-images.githubusercontent.com/12628669/154262645-a0a0b324-b8dd-4083-8494-b46c245cda6c.mp4

In a rudimentary sense I think most of the moving parts are in place. I think we can consider how much to refine these parts according to the check list above, before calling it merge to main. @mxgrey @koonpeng @mayman99 ?

Appreciate testing it as well, probably missed stuff.

koonpeng:
missing tests aside, the reporting dashboard is now broken.

mayman99:
To avoid the failing to submit a task from the dashboard side, I propose we replace these TextFields 
https://github.com/open-rmf/rmf-web/blob/cf2ca3fa23a97bdbac8fae046bc9d831fe6401ba/packages/react-components/lib/tasks/create-task.tsx#L126
with select components instead, what do you guys think?

koonpeng:
@mayman99 I'm adding tests and fixing bugs found along the way for components in the robot page.

> A problem I see with this is that if someone failed to submit a task via another interface, the dashboard will still get the fail to submit prompt as well.
>
> Edit: I've talked with Grey and we think the problem is not a bad thing.

This probably doesn't matter much now but this could be a security loophole if we have authorization. The server shouldn't process messages with unknown ids, so I would consider this a rather major bug.

> To avoid the failing to submit a task from the dashboard side, I propose we replace these TextFields

We can't do that as we don't have the list of possible values, we can't do automatic input validation as well because rmf does not provide a way to discover the task description schemas yet.

koonpeng:
> Submit batch task using form.

Uploading task from file is now fixed, with simple validations using ajv. We can't perform more validation without hardcoding because rmf doesn't provide a way to discover the list of task descriptions yet.

> Input validation to prevent invalid inputs from the create-task form. Task form might be a bit buggy as well.

Relating to this, we can't do more validation for the same reasons above. If that is ok, then we can tick this as well, else we need to either hardcode all the current validations or wait till rmf has task discovery done.

aaronchongth:
> Uploading task from file is now fixed, with simple validations using ajv. We can't perform more validation without hardcoding because rmf doesn't provide a way to discover the list of task descriptions yet.

Thanks for handling this, @koonpeng. Yup, that sounds like a reasonable solution for now. I'll note down for task description discovery as something to be looked into next time. I've ticked it off :sparkles:

aaronchongth:
Thank you all for crunching this out with such short notice. Will be merging this, and relying on a [separate issue](https://github.com/open-rmf/rmf-web/issues/584) to keep track of clean-up and fixes.

PR #588, opened 2022-02-28
fix reporting and reporting server

mayman99:
Signed-off-by: Mohamad <mohamad@openrobotics.org>

Fixes the front end so it builds, not tested with task-v2 yet.


mayman99:
good news from my side as well, I got the reporting server to work locally, so now I can actually test reporting end-to-end (I'll probably be doing that later today as it works for me only on the other system)

mayman99:
lgtm, the example worked by following the steps.

example-deployment/k8s/base/rmf-server/rmf-server.yaml
```diff
@@ -80,6 +80,23 @@ spec:
...
+  labels:
+    app: rmf-server-ws
+    tier: app
+spec:
+  selector:
+    app: rmf-server
+    tier: app
+  ports:
+    - protocol: TCP
+      port: 8001
```
koonpeng, line 98: This is not needed anymore as the api-server no longer uses multiple ports.

example-deployment/k8s/base/rmf-server/rmf-server.yaml
```diff
@@ -107,6 +124,8 @@ spec:
           env:
             - name: RMF_API_SERVER_CONFIG
               value: /rmf-server-config/rmf_server_config.py
+            - name: RMW_IMPLEMENTATION
+              value: rmw_fastrtps_cpp
```
koonpeng, line 128: I recalled that this was added to workaround minikube misconfiguration, iirc the problem was because minikube isolates the networks so other dds wouldn't work if rmf is ran locally and api-server is ran on the cluster. It only works for fastrtps because it supports shared memory.

I think we should remove this as abusing shared memory features to workaround broken network configuration is wrong. iirc it is not possible to create a configuration that works for other dds without also putting rmf inside the cluster, and even if it is in the cluster, it might still not work because dds requires udp broadcast which might not be supported by minikube.

We can also consider writing some deployment instructions for rmf + kubernetes in the rmf meta repo and link to that here since deploying rmf is out of scope of this repo. It might also make sense to move this example deployment to the meta repo as deployment of rmf + rmf-web has become quite coupled together, making it hard to have separate docs without explaining the other.

example-deployment/k8s/example-full/rmf_server_config.py
```diff
@@ -4,7 +4,7 @@
 
 config = deepcopy(default_config)
 config["host"] = "0.0.0.0"
-config["port"] = 8000
+config["base_port"] = 8000
```
koonpeng, line 7: New api-server reverts to using `port`.

packages/reporting/src/index.tsx
```diff
@@ -2,10 +2,14 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import App from './components/app';
 import * as serviceWorker from './serviceWorker';
+import LocalizationProvider from '@mui/lab/LocalizationProvider';
+import AdapterDateFns from '@mui/lab/AdapterDateFns';
```
koonpeng, line 6: `DateTimePicker` has graduated from material lab and into "material-x". This will cause reporting to fail when we merge https://github.com/open-rmf/rmf-web/pull/615.

koonpeng (changes requested):

aaronchongth:
Reporting has been removed via https://github.com/open-rmf/rmf-web/pull/628. Example deployment has been moved out to https://github.com/open-rmf/rmf_deployment_template

PR #587, opened 2022-02-28
Clean up after merging prototype/task_v2-2

mayman99:
Signed-off-by: Mohamad <mohamad@openrobotics.org>

## What's new
 solving issue #584 

## Self-checks

- [ ] I have prototyped this new feature (if necessary) on Figma 
- [ ] I'm familiar with and follow this [Typescript guideline](https://basarat.gitbook.io/typescript/styleguide)
- [ ] I added unit-tests for new components
- [ ] I tried testing edge cases
- [ ] I tested the behavior of the components that interact with the backend, with an e2e test

PR #589, opened 2022-03-01
draft: remove reporting

koonpeng:
## What's new

Testing how it would look like if we were to remove reporting from this repo. The diffs here can be used as reference to migrate reporting to its own repo.

This also fixes react-components lint errors.

## Self-checks


mayman99:
> This also fixes react-components lint errors.

I propose we separate this from this PR.

> Testing how it would look like if we were to remove reporting from this repo. The diffs here can be used as reference to migrate reporting to its own repo.

I also propose we hold this off until #588 and #585 get merged, as they both change the reporting side.

aaronchongth:
Per VC, we should decouple fixing the linting errors from this PR, i.e. have https://github.com/open-rmf/rmf-web/pull/587 merged first too

aaronchongth:
Closing via https://github.com/open-rmf/rmf-web/pull/628

PR #597, opened 2022-03-29
add last update time for robot page, automate robot page update [Dashboard]

mayman99:
Signed-off-by: Mohamad <mohamad@openrobotics.org>

## Self-checks


youliangtan (approved):
I just tested it, the auto-refresh and "last update" work well! 

Btw, when testing with the office world, I noticed one issue. When `tinyrobot2` is moving with an ongoing task, the robot panel 's `Assigned Tasks` field is empty. Meanwhile, I can see the allocated task in the task panel. We can also address this separate issue in another ticket/pr. Thanks!


Screenshot from 2022-03-29 15-13-46
.

packages/dashboard/src/components/robots/robot-page.tsx
```diff
@@ -162,7 +168,7 @@ export function RobotPage() {
       setIngestors(ingestors);
       setFleets(fleets);
     })();
-  }, [rmfIngress]);
+  }, [rmfIngress, _triggerRender, selectedRobot]);
```
koonpeng, line 171: Adding these dependencies causes all data to be fetched every render, `selectRobot` is probably not needed as well.

packages/dashboard/src/components/robots/robot-page.tsx
```diff
@@ -154,6 +154,12 @@ export function RobotPage() {
           });
           newRobotStates[robot.name] = robot;
           robotStatesRef.current[robot.name] = robot;
+          if (selectedRobot && selectedRobot.name && robot.name) {
+            setSelectedRobot(robot);
+            if (robot.task_id) {
+              setSelectedTask(taskStatesRef.current[robot.task_id]);
+            }
+          }
```
koonpeng, line 162: `selectRobot` is already handled by the click handler, in what cases are the click handlers insufficient?

packages/react-components/lib/robots/robot-info.tsx
```diff
@@ -15,6 +15,16 @@ function getTaskStatusDisplay(assignedTask?: string, taskStatus?: string) {
...
+function calculateLastUpdateTime(updateTime: number): string {
+  if (lastUpdateTime === updateTime) {
+    return ((Date.now() - timeOfLastUpdate) / 1000).toFixed(1);
+  } else {
+    lastUpdateTime = updateTime;
+    timeOfLastUpdate = Date.now();
+    return '< 1';
+  }
+}
+
```
koonpeng, line 27: This should use data-fns to provide better relative time formatting.

packages/react-components/lib/robots/robot-info.tsx
```diff
@@ -36,9 +46,12 @@ export interface RobotInfoProps {
   taskStatus?: TaskStatus;
   taskProgress?: number;
   estFinishTime?: number;
+  updateTime?: number;
 }
 
 const finishedStatus: TaskStatus[] = ['failed', 'completed', 'skipped', 'killed', 'canceled'];
+let timeOfLastUpdate: number = Date.now();
+let lastUpdateTime: number = 0;
```
koonpeng, line 54: Avoid global singletons.

packages/react-components/lib/robots/robot-info.tsx
```diff
@@ -15,6 +15,16 @@ function getTaskStatusDisplay(assignedTask?: string, taskStatus?: string) {
   }
 }
 
+function calculateLastUpdateTime(updateTime: number): string {
```
aaronchongth, line 18: I propose having more lenient ranges like `< 5s ago`, `< 30s ago`, `> 1m ago`, `> 5m ago`, this will hopefully also reduce the amount of re-rendering needed.

packages/react-components/lib/robots/robot-info.tsx
```diff
@@ -55,6 +69,11 @@ export function RobotInfo({
     <StyledDiv>
       <Typography variant="h6" style={{ textAlign: 'center' }} gutterBottom>
         {robotName}
+        <Typography variant="caption" display="block">
+          {updateTime !== undefined
+            ? `Last update - ${calculateLastUpdateTime(updateTime).toLocaleString()}s ago`
```
aaronchongth, line 74: nit:
```suggestion
            ? `Last updated - ${calculateLastUpdateTime(updateTime).toLocaleString()}s ago`
```

koonpeng, line 74: This part is interesting, I couldn't find what `String.toLocaleString` do. If that is required to get proper string for display, should we be applying this to all UI elements?

koonpeng (dismissed):

aaronchongth (dismissed):
Thanks for whipping this solution up so quickly! It looks amazing!

While testing, I found my robot side panel keeps jumping back to `TinyRobot2` even after I select `TinyRobot1`, as shown in the video,

https://user-images.githubusercontent.com/5383623/160564847-bdfa4ea0-dbcb-4899-967e-caa59131b864.mp4

I have also started experiencing a lot of lag too, could this be due to the amount of re-rendering that is happening?

packages/react-components/lib/robots/robot-info.tsx
aaronchongth: oh! `data-fns` looks amazing! That would be awesome

packages/dashboard/src/components/robots/robot-page.tsx
mayman99: Yes, well the user need to click on a task to see the task-progress get updated.

packages/react-components/lib/robots/robot-info.tsx
mayman99: I found out why you had the issue, its a miss type from my side, thanks.
about the ranges, it is a good idea, so the update interval wouldn't be a constant one second, but a number that increases exponentially with time?

packages/dashboard/src/components/robots/robot-page.tsx
mayman99: Yes, that is the whole, idea, because the data are not fetched automatically, and the user need to refresh the page to see any update, in the table or the panel.

mayman99:
For now, unless there is a way to not depend on triggers to re-fetch data, the only way I can think of to make it faster it to remove the rerender trigger from triggering data fetch, and only do that when the user changes the robot selected.
I'm open to ideas.

packages/react-components/lib/robots/robot-info.tsx
aaronchongth: Per VC, this will be resolved when we look into using `data-fns` but for now a lower update rate has made it more consistent

packages/react-components/lib/robots/robot-info.tsx
```diff
@@ -15,6 +15,20 @@ function getTaskStatusDisplay(assignedTask?: string, taskStatus?: string) {
...
+  lastUpdateTime: number,
+  timeOfLastUpdate: number,
+): string {
+  if (lastUpdateTime === updateTime) {
+    return ((Date.now() - timeOfLastUpdate) / 1000).toFixed(1);
+  } else {
+    lastUpdateTime = updateTime;
+    timeOfLastUpdate = Date.now();
+    return '< 1';
+  }
```
youliangtan, line 29: With this, I was facing the issue of the latest update keeps showing "<1", despite the fleet_states are not updating from rmf.

It works after changing this  function block impl and changing it to a single line
```
  return ((Date.now() - updateTime) / 1000).toFixed(1);
```
I believe this is actually sufficient and we don't need `lastUpdateTime` and `timeOfLastUpdate` argument.

youliangtan (changes requested):

packages/react-components/lib/robots/robot-info.tsx
youliangtan: @mayman99   This feature with the above changes is tested and proved well in https://github.com/open-rmf/rmf-web/pull/603/files#r903270019.

Understand that this might be incorrect for `use_sim_time` sim scenario. One way to move forward is to merge this helpful feature in to `main`, while opening an issue to acknowledge this sim_time issue when running it on simulation.

mayman99: Sorry I just saw your comment right now, I agree :+1:

packages/react-components/lib/robots/robot-info.tsx
```diff
@@ -15,6 +15,27 @@ function getTaskStatusDisplay(assignedTask?: string, taskStatus?: string) {
   }
 }
 
+function calculateLastUpdateTime(
```
youliangtan, line 18: I tested it with the deployment setup, in realtime. this should be the equation. which just required one input arg of `updateTime`

```js
function calculateLastUpdateTime(
  updateTime: number,
): string {
    const lastUpdateMillis = Date.now() - updateTime;
    if (lastUpdateMillis < 1000) {
      return '< 1';
    } else {
      return (lastUpdateMillis / 1000.0).toFixed(1).toString();
    }
}
```

youliangtan (changes requested):
I tested it with a deployment setup with actual wall time. To test it, I killed the fleet_adapter after some keeping it running for a while. Unfortunately, the last update time will not increase overtime. Here to illustrate again, the definition of this lastupdate is when the fleetadapter last update the `fleetstate` to the `api-server`, therefore using just the `updateTime` arg is sufficient.

After changing it, i managed to get the right `lastupdate` value. However, this still requires me to manually refresh the page to observe the change in `Last update - 12.86s ago` statement. Hopefully we can also have this value update periodically too.

P/s: If you wish to test a demo example with realtime. feel free to run the 2 launch commands in this pending pr: https://github.com/open-rmf/fleet_adapter_ecobot/pull/1 . (kill the fleetadapter node after keeping it running for a short while)

packages/react-components/lib/robots/robot-info.tsx
```diff
@@ -47,14 +73,33 @@ export function RobotInfo({
...
     <StyledDiv>
       <Typography variant="h6" style={{ textAlign: 'center' }} gutterBottom>
         {robotName}
+        <Typography variant="caption" display="block">
+          {updateTime !== undefined &&
+          lastUpdateTime !== undefined &&
+          timeOfLastUpdate !== undefined &&
+          setLastUpdateTime &&
+          setTimeOfLastUpdate
+            ? `Last update - ${calculateLastUpdateTime(
```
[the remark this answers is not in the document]
aaronchongth: https://github.com/open-rmf/rmf-web/pull/597/commits/fb4f313a6f0ec1640553e7a6a4fe6e8a9971a780

packages/react-components/lib/robots/robot-info.tsx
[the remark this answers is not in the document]
aaronchongth: One of the reasons why we needed all the other fields was to ensure it worked in simulation, by calculating using time relative to the dashboard.

Your suggestion would work only in deployment, I believe, but in simulation `updateTime` would be unix epoch time and `Date.now() - updateTime` would be inaccurate.

aaronchongth:
@youliangtan @mayman99, I could not find a good solution to allow both sim and real time while displaying how long ago the lats update was. I've instead added a new column in the robot table, which displays the time itself. Let me know what you think.

image

youliangtan (approved):
Thanks. I think this new proposal of adding a "last update" column makes total sense. Looks good to me!

PR #598, opened 2022-03-29
CI: use matrix to test with older npm ver

koonpeng:
## Self-checks


aaronchongth (approved):
Thanks for adding this! There are still CIs not passing but looks like they were failing in a similar fashion on main as well, this can be merged then.

PR #609, opened 2022-05-09
api-server: maybe fix memory leak

koonpeng:
## What's new

This PR made 2 changes (one of them relatively major) which may fix the memory leak observed in long term usage.

1. Fix rxpy subscriptions getting leaked when client disconnects.
2. As previously discussed, it is possible that the leak might have something to do with uvicorn. Unfortunately the api-server uses a hack which heavily relies on the server being uvicorn to allow running 2 separate ASGI "instances" so it is not possible to use a different server. This PR removes the hack and instead exposes the internal gateway at `/_internal`. All endpoints are now on the same port, in order to ensure that the internal endpoint is not exposed to public, deployments need to do one of
    1. Set up a reverse proxy so that it only allows access to the internal endpoint from a specific network.
    2. Use an alternative ASGI server that supports filtering access based on the path, e.g. apache. More advanced servers can also bind to multipe ports, together with path filtering and url rewrite, it is possible to mimic the current behaviour.
    3. Alternatively, we can add support for "Client Credentials Flow" of oauth/oidc, in this auth method, all rmf nodes that connects to the api-server must have a client_id registered in the identity provider, they must connect with the identity provider to obtain an access token and pass this token to the api-server. The api-server can then be configured so that the internal endpoint can only be accessed with the correct tokens.

## Self-checks


aaronchongth:
Per VC, @youliangtan will help test this fix locally, as well as consider the 3 options needed for deployment.

packages/api-server/api_server/__main__.py
```diff
@@ -1,121 +1,6 @@
...
-        server_manager.shutdown()
-
-    signal.signal(signal.SIGINT, stop_servers)
-    signal.signal(signal.SIGTERM, stop_servers)
-
-    server_manager.run()
-
 
 if __name__ == "__main__":
-    main()
```
youliangtan, line 121: This `main()` fn is still needed, due to https://github.com/open-rmf/rmf-web/blob/987fdf3016a2ed0373b3312cb87f26052a2ce697/packages/api-server/setup.py#L36

packages/api-server/api_server/__main__.py
```diff
@@ -1,120 +1,10 @@
...
-    server_manager = ServerManager()
-
-    def stop_servers(_signum, _frame):
-        server_manager.shutdown()
-
-    signal.signal(signal.SIGINT, stop_servers)
-    signal.signal(signal.SIGTERM, stop_servers)
-
-    server_manager.run()
+    uvicorn.run(app)
```
youliangtan, line 7: Also will need to populate the config to the server
```
    uvicorn.run(
        app,
        host=app_config.host,
        port=app_config.port,
        root_path=app_config.public_url.path,
        log_level=app_config.log_level.lower(),
    )
```

youliangtan (changes requested):

packages/api-server/api_server/__main__.py
koonpeng: I just pushed a fix for this. Thanks for helping to test!

youliangtan (approved):
Thanks for the fix~ The commit on fixing https://github.com/open-rmf/rmf-web/issues/614#issuecomment-1150645529 solves the issue. This fix is really important for rmf deployment.

On the side note, i will use open a pr for the fix on the client side.

PR #613, opened 2022-05-27
Fix ros-translator CI test

aaronchongth:
Signed-off-by: Aaron Chong <aaronchongth@gmail.com>

## What's new

Fixes broken test in `ros-translator`, due to https://github.com/microsoft/TypeScript/issues/8186.

## Self-checks

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -3,9 +3,6 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 name = "pypi"
 
-[packages]
-jinja2 = "*"
-
 [dev-packages]
 black = "==21.12b0"
 isort = "~=5.8"
--- a/Pipfile
+++ b/Pipfile
@@ -4,6 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
+jinja2 = "*"
 
 [dev-packages]
 black = "==21.12b0"
--- a/packages/ros-translator/ros_translator/typescript/test/test.sh
+++ b/packages/ros-translator/ros_translator/typescript/test/test.sh
@@ -5,6 +5,6 @@ cd $(dirname $0)
 rm -rf out
 PIPENV_MAX_DEPTH=6 pipenv run python -m ros_translator -t=typescript -o=out ros_translator_test_msgs
 echo 'test build'
-tsc --noEmit -p "$(pwd)/tsconfig.json"
+tsc --noEmit
 echo 'ok'
 ts-node -T $(npx which jasmine) "$@"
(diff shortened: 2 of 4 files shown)


koonpeng:
I don't think this fixes anything, adding `|| echo done` simply turns an error return into a success.

aaronchongth:
yup, but from https://github.com/open-rmf/rmf-web/runs/6621350437?check_suite_focus=true, the error seems to occur from just the basic `npx tsc --noEmit` call. Is that really a test error?

It looks like the real test failure is `ImportError: cannot import name 'soft_unicode' from 'markupsafe'`, I'll look into it too

koonpeng:
It appears the problem is due to the changes in `npx` it now seems to always run in the package directory. Solution is to simply drop usage of npx in the typescript's `test.sh`, it is not needed if it is called as part of the main test script.

aaronchongth:
@koonpeng It looks like the issue in CI is related to the version of `markeupsafe` bumped up by https://github.com/open-rmf/rmf-web/pull/612 recently.

Looks a little similar to this https://github.com/aws/aws-sam-cli/issues/3661 and reverting the version to `2.0.1` seems to help. Do you think it'll be an issue to revert the version?

koonpeng:
That bump is because of a high severity security vulnerability, that particular library handles verifying jwt tokens so reverting will be quite dangerous.

packages/ros-translator/ros_translator/typescript/test/test.sh
```diff
@@ -5,6 +5,6 @@ cd $(dirname $0)
 rm -rf out
 PIPENV_MAX_DEPTH=6 pipenv run python -m ros_translator -t=typescript -o=out ros_translator_test_msgs
 echo 'test build'
-npx tsc --noEmit
```
koonpeng, line 8: Can you revert this? the `$(pwd)/tsconfig.json` shouldn't be needed since we already `cd` into the correct directory

Pipfile
```diff
@@ -4,6 +4,7 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
+jinja2 = "*"
```
koonpeng, line 7: `jinja2` is a dev package, we should try to categorize it correctly. Try manually moving this into `dev-packages` and do a `pipenv sync`.

koonpeng (changes requested):

packages/ros-translator/ros_translator/typescript/test/test.sh
aaronchongth: Done, https://github.com/open-rmf/rmf-web/pull/613/commits/02a8fdf600b13c130c56e4d4ead40464a166df12

koonpeng (approved):
Problem is because `ros-translator` has dep on jinja2~=2, so the change should be in `ros-translator`'s `setup.py`. I fixed that now, feel free to merge it whenever.

aaronchongth:
Can't believe that was missed out. Thank you!

PR #603, opened 2022-04-20
Demo rmf-web fixes

youliangtan:
Fixing frontend and backend: (Note: tick if the feature is merged into `main`)

- [x]  Full-screen RMF map. Commented out the infrastructure section (without lift and doors). TODO: make this collapsible
- [x]  Display "last update" time for robot_state, helpful to check if fleet adapter is running (https://github.com/open-rmf/rmf-web/pull/597)
- [x]  Cancel task and display of task itinerary (https://github.com/open-rmf/rmf-web/pull/620)
  - Make queued task cancelable, and grayout the cancel button when task is completed/killed/canceled 
  - Show task status for terminated tasks
- [x] Front end dashboard misc changes and fixes (https://github.com/open-rmf/rmf-web/pull/621)
  - Limit battery percentage decimals to (2)
  - set default traj length to 5 mins (prev was 1 min)
  - Fix incorrect "finish time" for task displayed in robot tab
  - Add `category` column to the task table on tasks tab. According to user feedback, this is useful.
  - Fix taskprogressbar incorrect calculation
- [x]  Fix task log `unix_millis_time` from 32 to 64 bit  (https://github.com/open-rmf/rmf-web/pull/618)
- [x]  Task Submission form: `Loop` (https://github.com/open-rmf/rmf-web/pull/622)
  - Update `Loop` Task form to accept single Location, since internally now it is a `Patrol` Task
- [ ]  Task Submission form: `Clean`
  - Update `Clean` Task form to be compatible with the new `perform action` composed clean task.
- [x]  Fix `api-server`'s `/robot_task` endpoint response body   (https://github.com/open-rmf/rmf-web/pull/619)
- [x]  Temp fix on front-end consistent sub unsub of socketio (https://github.com/open-rmf/rmf-web/pull/623)

PR #618, opened 2022-06-21
log time with 64 bit

youliangtan:
Originated from https://github.com/open-rmf/rmf-web/pull/603

Previously will get receive an error printout when the timestamp is more than 32bit `IntField` .


aaronchongth (approved):
Thanks for the fix!

PR #620, opened 2022-06-21
Fix/dashboard task panel display and cancellation for different task status

youliangtan:
Originated from https://github.com/open-rmf/rmf-web/pull/603

This PR introduces 2 features:
 - Enable cancellation for queued task. Previously the button will get gray-out for queued task
 - Show the "task itinerary" when the task is already completed/canceled. (The task details shown on the right when clicked on the task on task panel). It will not show up before this fix.
 - (Update) fix client not sending taskcancellation req to server in new windowing sys


aaronchongth (changes requested):
Thanks for putting together this fix! I found that I could not select a queued task (for the cancellation button to show up) on the robot table, if I was already selecting a task that is underway.

You can reproduce this issue by first launching `office_loop.launch.xml`, then once the first 2 tasks show up and is `underway` on the table, launch `office_conflict.launch.xml`. When the task state and log panels are empty, clicking on a queued state will show the `cancel` button. However once you select a task that is `underway`, you can't click onto a task that is `queued`.

aaronchongth (approved):
Just tested and it works perfectly now, thanks! LGTM

PR #622, opened 2022-06-22
accept single place for loop task form

youliangtan:
PR originated from: https://github.com/open-rmf/rmf-web/pull/603

With the new `patrol` task (previously known as `Loop`), the task form accepts a list of places, instead of a pair of `start` and `end` places. Thus this form is to make the second place optional, helping dispatch the robot to a single location. Also, with the new `patrol` task, the number of loops is changed to `rounds`

Screenshot from 2022-06-22 12-12-40

### DIFF
--- a/packages/react-components/lib/tasks/create-task.tsx
+++ b/packages/react-components/lib/tasks/create-task.tsx
@@ -1,5 +1,6 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
+import { NineKOutlined } from '@mui/icons-material';
 import {
   Autocomplete,
   Button,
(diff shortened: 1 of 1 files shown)


aaronchongth (changes requested):
Thanks for splitting this out @youliangtan, this is awesome!

I think this branch might still be missing some of the components related though, for example https://github.com/open-rmf/rmf-web/pull/603/files#diff-b303145e2e7e318221a93e9f20d4e8f50dab4b4aae04b4b94a8db382e7b406e4R457, the `optional` text field. Could you help take a look what other changes are needed for this task form?

youliangtan:
Thanks for noticing this, I missed it out in a separate commit.

koonpeng:
Making the 2nd location doesn't address that the patrol task can accept more than 2 locations. The most complete fix would be to make the form accept a list. As a quick fix we can also consider using a comma separated list of locations, we would have to change it to a freeform text input though (i.e. no autocomplete), then again, the way we had been getting the locations has always been a "hack" as taskv1 doesn't offer such apis so we are just grabbing all named waypoints (which may not be valid task locations). Taskv2 should provide an endpoint to list the valid locations but afaik that has not been implemented yet.

youliangtan:
@koonpeng valid point, I have updated the task submission form for loop (aka patrol) task, which now supports multiple places input.
ezgif com-gif-maker (1)

koonpeng (approved):
LGTM

koonpeng:
@youliangtan Noticed lint errors from this PR https://github.com/open-rmf/rmf-web/actions/runs/2587166568. Can you open another PR to fix it?

PR #623, opened 2022-06-22
feat: window system and many other fixes and improvments

koonpeng:
## What's new

* Finished the window system, there are 2 custom ui provided that allows the user to freely configure the layout.
* Properly supports auto refresh in robots and tasks app.
  * Removed auto refresh toggle in tasks app, it is mostly redundant, other components like robots app also does not have that button.
* rmf ingress now handles tracking of subscriptions with the help of rxjs.
  * This also allows many components to avoid unnecessary re-renders.
* Remove many outdated components that are no longer working or showing the wrong information. Including hotkeys, tooltip toggle, settings drawer, help drawer etc.
* Finish theming support. Includes 3 themes, material default, rmf-light and rmf-dark. Ideally, I would prefer to reduce it to 2 themes, but imo, the rmf-light is very bad so I added the material default.

https://user-images.githubusercontent.com/45189696/174951198-8749aef6-8311-4e7b-ab68-2cba80a20c09.mp4

## Self-checks


koonpeng:
> This is very awesome!!! I can imagine this customization is very helpful to users. Tested it locally and all works well. partying_face
One thing I noticed, on the robot-page and task-page, there are whitespaces on both sides. Will it be good if both robot&task pages can be consistent with the infra page, not having whitespaces on the sides?

These layouts are designed based on a 1440p screen, the problem is that the system does not support responsive layouts.

youliangtan:
Just drop in some things that I noticed during my tests:
- When a new task is created from the dashboard's task-page  (1st client), the new task will not appear automatically on another task page (2nd client). This requires the 2nd user to click on the "refresh button" to trigger the `/tasks` REST request. understand that this works previously because of constant polling, but that is also not ideal too.
- Similar to above, when a robot is newly added after loading up the robot page, the robot will not automatically get added to the page. This demands a F-5 refresh since a refresh button is not avail
- When selecting a robot, the map won't zoom and focus on the selected robot. This useful behavior is not available as compared to the previous version

koonpeng:
Thanks for helping to test. The first 2 points are expected, for tasks, we only listen for task and fleet states for the currently displayed items. For robots, rmf does not support notification when a new robot is added. For tasks, due to paging and sorting, it is possible for tasks to jump around if we implement that.

I think for tasks, what we can do is implement a new "Active Tasks" microapp, this microapp will show all active tasks without any paging completed or failed tasks will be immediately removed. For robots, we have to add the required events in rmf for it to work "properly".

koonpeng:
> Similar to above, when a robot is newly added after loading up the robot page, the robot will not automatically get added to the page. This demands a F-5 refresh since a refresh button is not avail

Looking at the code, new robots should show up as we are listening to fleet states. Can you confirm if it is a new robot being added to an existing fleet or a new fleeting being added?

aaronchongth:
Per VC,
* Clicking on the robots in the tables, the maps do not center on that robot at the moment.
* New tasks created by current client do not show up on other dashboard clients readily, unless the refresh button is clicked.

These 2 notes should not block this PR from merging, new feature requests would be opened.

youliangtan (approved):
Thanks! Feel free to merge this in :+1:

PR #624, opened 2022-06-28
deps: humble fixes

koonpeng:
## What's new

This updates the pipfile to use python 3.10, the consequences of this is that it will no longer work on ubuntu 20.04. `pipenv` does not support python version ranges so we can't support both versions at the same time.

Other alternatives workarounds:
* Remove the python version check, this will allow it to work for both ubuntu 20.04 and 22.04, but the issue with this is that pipenv will just use whatever python version it can find. This means that the resolution of dependencies will be different depending on the environment and since deps are locked into `pipfile.lock` and committed into git, it can create many many problems.

The only way I can think of that works for both ubuntu 20.04 and 22.04 is to maintain multiple branches. If we want to continue supporting galactic and below, we can branch off the current `main` and have `main` only support humble.

This is built on top of #623.
Requires https://github.com/open-rmf/rmf/pull/165.

## Self-checks


koonpeng:
closed in favor of #628

PR #627, opened 2022-07-06
runtime configure dashboard env vars

youliangtan:
Currently, when building the `dashboard.Dockerfile`, we will need to specify the URL during build time. When we wish have containers with different URLs, this requires us to build multiple images with their own respective URL. Hence, having a `public/env.js`, which can be edited during runtime enables us to share one single dashboard docker image.

This command can be executed during runtime
```bash
echo "$(cat <<EOM
window.ENV =
{
  PUBLIC_URL:"/dashboard",
  REACT_APP_TRAJECTORY_SERVER:"wss://${DOMAIN_URL}/trajectory",
  REACT_APP_RMF_SERVER:"https://${DOMAIN_URL}/rmf/api/v1",
  REACT_APP_AUTH_PROVIDER:"keycloak",
  REACT_APP_KEYCLOAK_CONFIG='{"realm": "rmf-web", "clientId": "dashboard", "url" : "https://${DOMAIN_URL}/auth"}',
}
EOM
)" > $ENV_CONFIG_FILE_PATH
```

the `ENV_CONFIG_FILE_PATH` is located in `dashboard/build/env.js`

Referred to https://stackoverflow.com/questions/51653931/react-configuration-file-for-post-deployment-settings/51663697#51663697

### DIFF
--- a/packages/dashboard/public/env.js
+++ b/packages/dashboard/public/env.js
@@ -1,0 +1,15 @@
+// public/env.js
+window.ENV = {
+  REACT_APP_TRAJECTORY_SERVER: 'ws://localhost:8006',
+  // REACT_APP_AUTH_PROVIDER: "keycloak",
+  // REACT_APP_KEYCLOAK_CONFIG: '{"realm": "rmf-web", "clientId": "dashboard", "url" : "https://localhost:8000/auth"}',
+  REACT_APP_RMF_SERVER: 'http://localhost:8000',
+  PUBLIC_URL: '/dashboard',
+};
+
+// ARG DOMAIN_URL="rmf-deployment-template.open-rmf.org"
+// ENV PUBLIC_URL="/dashboard"
+// ENV REACT_APP_TRAJECTORY_SERVER="wss://${DOMAIN_URL}/trajectory"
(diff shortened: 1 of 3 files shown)


koonpeng:
This "limitation" of the dashboard has come up before, after thinking about it, my conclusion is that we should just build the dashboard as the build system intended. Trying to circumvent it usually leads to several problems and inefficencies.

* Typescript is not aware of these variables so compilation fails (fixable with declaration files).
* The extra `env.js` requires an extra fetch before loading the page.
  * It is also not processed by the build system so cannot be cached.
  * Extra fetch can be avoided with http2.0 push, but requires web server to be configured correctly.
* It doesn't address the web server configuration and may also negatively affect other deployment methods.
  * In the example deployment, nginx is used to the serve the webapp, the url configuration needs to be set there as well.
  * In other deployments, particularly static server (github pages, cloudflare etc), there is no way to modify `env.js` at "run time", also it may not be possible to configure the web server to automatically push `env.js` to avoid the extra fetch.

If you only want to serve the same dashboard on multiple urls, it can be configured on the reverse proxy. If you want to serve different dashboards but don't want to build with different configurations, you can try the "placeholder url and sed at run time" approach. However, these workarounds should only be done at the deployment configuration level, I think we should just stick to building a static webapp as the build system is designed.

PR #632, opened 2022-07-25
api-server: fix fastio to work with newer uvicorn

koonpeng:
## What's new

Since upgrading uvicorn (probably from #617), you may get `tortoise.exceptions.ConfigurationError: No DB associated to model` errors. This is due to the app startup not being called at all, while I'm not sure of the root cause, switching the main asgi app to fastapi fixes it.

## Self-checks


youliangtan (approved):
tested, and confirm this fixes the issue. thanks!

PR #633, opened 2022-07-27
Add numpy to pipfile

xiyuoh:
This PR adds numpy as one of the packages to install. Without it I got a `ModuleNotFoundEror: No module named 'numpy'` and is unable to launch rmf-web

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -12,6 +12,7 @@ coverage = "~=5.5"
 api-server = {editable = true, path = "./packages/api-server"}
 requests = "~=2.25"
 asyncpg = "~=0.25.0"
+numpy = "~=1.23.1"
 websocket-client = "~=1.2.3"
 # reporting-server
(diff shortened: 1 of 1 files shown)


aaronchongth (approved):
Thanks for the fix!

koonpeng:
This issue came up a few times before, I don't think we ever got to the cause of it. `rmf-web` does not use numpy so we shouldn't add it to the Pipfile, the Pipfile is also only used for dev packages so this does not solve the problem in production.

numpy is used by some ros packages, for some reason that is not being installed.

aaronchongth:
I've also faced this issue in the past before when trying to use `rmf-web` for the first time :thinking: Is this issue plaguing any of our efforts with deployment using docker containers at the moment? That might be a good place to start debugging it

In the meantime, would you recommend reverting this change, or adding a comment saying that it is a stopgap before we find the real cause?

PR #634, opened 2022-08-11
safe check internal ws msg 'type' field

youliangtan:
For some instance when running rmf_demos, i will get this error:
```
    await dependant.call(**values)
  File "/usr/local/lib/python3.8/dist-packages/api_server/routes/internal.py", line 49, in rmf_gateway
    await process_msg(msg, fleet_repo)
  File "/usr/local/lib/python3.8/dist-packages/api_server/routes/internal.py", line 18, in process_msg
    payload_type: str = msg["type"]
TypeError: list indices must be integers or slices, not str
```

I noticed some client is sending in an msg data with empty array `[]`, which is not compatible to the rmf msg. This requires further investigation to fix this. Meanwhile, the api server should also be robust enough to pick up incompatible msg, thus a simple safe check to ensure 'type' exists in the msg field.

### DIFF
--- a/packages/api-server/api_server/routes/internal.py
+++ b/packages/api-server/api_server/routes/internal.py
@@ -16,11 +16,12 @@ task_repo = TaskRepository(user)
 
 async def process_msg(msg: Dict[str, Any], fleet_repo: FleetRepository) -> None:
     if "type" not in msg:
-        logger.debug("Ignoring message, 'type' must include in msg field")
+        logger.warn(msg)
+        logger.warn("Ignoring message, 'type' must include in msg field")
         return
     payload_type: str = msg["type"]
     if not isinstance(payload_type, str):
-        logger.error("error processing message, 'type' must be a string")
+        logger.warn("error processing message, 'type' must be a string")
         return
     logger.debug(msg)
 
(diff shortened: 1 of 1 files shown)


packages/api-server/api_server/routes/internal.py
```diff
@@ -15,6 +15,9 @@
 
 
 async def process_msg(msg: Dict[str, Any], fleet_repo: FleetRepository) -> None:
+    if "type" not in msg:
+        logger.debug("Ignoring message, 'type' must include in msg field")
```
koonpeng, line 19: This should be `logger.warn`. While we are at it, let's also change line 23 to `logger.warn` as well. I think warn is more suitable than error because the source of the problem is not with api server, it is just warning that it received something unexpected.

packages/api-server/api_server/routes/internal.py
```diff
@@ -15,6 +15,9 @@
...
+        return
     payload_type: str = msg["type"]
     if not isinstance(payload_type, str):
         logger.error("error processing message, 'type' must be a string")
```
koonpeng, line 23: Can you add `logger.warn(msg)` to this and the above `type` check? Having it print out the msg will be very helpful.

koonpeng (changes requested):

packages/api-server/api_server/routes/internal.py
youliangtan: Done! https://github.com/open-rmf/rmf-web/pull/634/commits/9ce9b7470f547514071eb16a525a64b9ae682c47

koonpeng (approved):

PR #641, opened 2022-09-14
Robot issues app

aaronchongth:
## What's new

Adds a micro app for displaying reported robot issues.

robot_issues-2022-09-14_15 45 04

A quick and dirty way to test it would be replacing https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/src/rmf_fleet_adapter/agv/FleetUpdateHandle.cpp#L882-L888, with some fake issues

```cpp
      // object issue
      nlohmann::json object_issue_msg;
      object_issue_msg["category"] = name + "object_category";
      nlohmann::json object;
      object["first_field"] = "first_value";
      object["second_field"] = "second_value";
      object_issue_msg["detail"] = std::move(object);
      issues_msg.push_back(std::move(object_issue_msg));
      // array issue
      nlohmann::json array_issue_msg;
      array_issue_msg["category"] = name + "string_category";
      array_issue_msg["detail"] = {"first_value", "second_value"};
      issues_msg.push_back(std::move(array_issue_msg));
      // string issue
      nlohmann::json string_issue_msg;
      string_issue_msg["category"] = name + "string_category";
      string_issue_msg["detail"] = "string_value";
      issues_msg.push_back(std::move(string_issue_msg));
```

## Self-checks

## Discussion

I have not been able to implement resetting all the open accordions when a new robot is selected. Any ideas?


koonpeng:
I think this could be part of robot info, what do you think?

aaronchongth:
I was playing around with that idea too. The next step is actually to include robot logs too, and I was hoping to avoid panels/spaces that will most of the time remain empty. 

What do you adding it as a clickable icon at the end of the robot table row, that opens up an overlay?

koonpeng:
I don't see any advantage of that over just having it on the robot info. If there is no issues, it would actually be important to indicate that. There should be an indicator at the robot table to quickly find robots with issues, but I don't think it needs any special click handler, it can just open the robot info as normal.

PR #647, opened 2022-10-04
Fix/ci

aaronchongth:
## What's new

Fixes broken CI for
- [x] nightly, dashboard-e2e
- [x] dashboard
- [x] api-server, for checking None issues, https://github.com/microsoft/pyright/issues/2007

Changing the `e2e` test to check for `coe_door` instead of `main_door` as the second pane sometimes gets stuck
image

## Self-checks

### DIFF
--- a/.github/actions/bootstrap/action.yml
+++ b/.github/actions/bootstrap/action.yml
@@ -19,7 +19,7 @@ runs:
         node-version: '16'
         cache: 'pnpm'
     - name: Install dependencies
-      run: pnpm install -w --filter ${{ inputs.package }}...
+      run: pnpm install -w --filter ${{ inputs.package }}... --no-frozen-lockfile
       shell: bash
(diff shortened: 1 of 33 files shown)


aaronchongth:
After a few days of attempting to fix `dashboard-e2e` I have concluded that it is incredibly flaky running on a github workflow :sweat_smile: After re-running it for 10+ times, it has finally passed. We might need to re-consider running it on github runners

mayman99 (approved):
lgtm
I remember when this started as test not finding a button and now this fixed a whole lot of stuff :D

aaronchongth:
There is still a high chance that `dashboard-e2e` is going to fail on the `main` branch, I think we might need to be a bit more mindful abuot that and run it locally to make sure thigns work. Otherwise we can try to migrate this workflow onto something else more robust than a github runner.

PR #649, opened 2022-10-18
Draft: react-leaflet upgrade to 3.2.5

mayman99:
## What's new
react-leaflet 3.2.5 is the latest version that doesn't depend on React v18

After a react upgrade to 18, we could upgrade once more to react-leaflet v4+

## Self-checks


mayman99:
This is still a WIP, will continue after ROSSon most probably

aaronchongth:
Per VC, we will be migrating to `react-three-fiber` instead of continuing with `react-leaflet`. I'll close this PR, feel free to keep the branch if you think it is still useful for the `react` upgrade

PR #651, opened 2022-10-21
Feature/implementation of datagrid in tasks section

Angatupyry:
## What's new
Adding [Datagrid](https://mui.com/x/react-data-grid/) instead of TaskTable component in TaskApp.

- Order by column (asc, desc)
- Filtering by Start time, End time and Category column (Server side)

[screen-capture.webm](https://user-images.githubusercontent.com/37310205/197220119-fc874cd0-47a1-463c-91d3-caae05da72a9.webm)

## Self-checks


mayman99 (approved):
lgtm

mayman99 (changes requested):
lgtm... just the CI thing and thins looks good to go

packages/dashboard/src/components/tasks/tasks-app.tsx
```diff
@@ -91,44 +99,54 @@ export const TasksApp = React.memo(
         setWorkcells(Object.keys(resourceManager.dispensers.dispensers));
       }, [resourceManager]);
 
+      const GET_LIMIT = 10;
```
aaronchongth, line 102: Could you add a `TODO` above this line to hopefully parameterize this value in our next set of fixes?

aaronchongth (changes requested):
This is awesome! Thanks for putting together this feature, I have a few nitpicks and some questions but generally it is looking real good.

While testing with simulation, I did run into issues where the filters overlay keeps jumping without allowing the user to give inputs to the filter. This doesn't happen with the data dump, and the frequency looks to be the same as incoming task state updates, could you help look into it?

https://user-images.githubusercontent.com/5383623/200347074-d839370f-7948-446e-ade7-ac55f9f7fdc2.mp4

packages/react-components/lib/tasks/task-table-datagrid.tsx
```diff
@@ -0,0 +1,579 @@
...
+    color: theme.palette.getContrastText(theme.palette.grey[300]),
+  },
+  [`& .${classes.taskUnknownCell}`]: {
+    backgroundColor: theme.palette.warning.main,
+    color: theme.palette.getContrastText(theme.palette.warning.main),
+  },
+}));
+
+export interface DefaultTableDataGridProps {
+  isLoading: boolean;
```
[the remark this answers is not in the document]
Angatupyry: I have set this variable as initial value to true. It will be useful when the api has some kind of delay and it will show a loading icon

packages/dashboard/src/components/tasks/tasks-app.tsx
Angatupyry: `const GET_LIMIT = 10; // TODO PARAMETERIZE THIS VARIABLE`
Something like that?

Angatupyry:
@aaronchongth I have changed the variable names according to your suggestions. They make a lot of sense. 
Regarding how the filters are working with your tests, could you describe for me the steps needed to reproduce that scenario?

packages/dashboard/src/components/tasks/tasks-app.tsx
aaronchongth: Something above it would work,
```
// TODO: parameterize this variable
const GET_LIMIT = 10;
```

aaronchongth:
Per VC, it looks like the task logs do not automatically get updated, without the task subscriptions, could you investigate if it is possible to workaround this issue, perhaps to pause updating the table while a filter is being configured.

You can try it out with,
```bash
# start api-server and dashboard, start office demo
ros2 launch rmf_demos_gz office.launch.xml server_uri:="ws://localhost:8000/_internal"

# start loop tasks
ros2 launch rmf_demos office_loop.launch.xml
```

Refreshing and clicking onto a task, we can see that the logs do not get updated. Clicking on the task again after a while, we see that lots of newer logs come in.

Angatupyry:
@aaronchongth The logs were also not being displayed in the main branch. In the video you can see that you have to click to see the updated logs.
[screen-capture.webm](https://user-images.githubusercontent.com/37310205/201959062-ba5507b0-026a-4671-b588-d6bf4123d78f.webm)

This is something we need to work on if we want to see the logs in real time.

aaronchongth (approved):
Just ran some tests and everything seems to be working as expected, thanks for the amazing implementation @Angatupyry! LGTM

PR #654, opened 2022-11-24
Server-side sorting

aaronchongth:
## What's new

Continuation of https://github.com/open-rmf/rmf-web/pull/651, implementing server-side sorting.

Incorporates changes from https://github.com/open-rmf/rmf-web/pull/538, since database migration will be required.

https://user-images.githubusercontent.com/5383623/203946045-035ad1e4-8c31-46df-bb8b-e8b3fecfdca5.mp4

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/tasks/tasks-app.tsx
+++ b/packages/dashboard/src/components/tasks/tasks-app.tsx
@@ -6,11 +6,12 @@ import React from 'react';
 import {
   CreateTaskForm,
   CreateTaskFormProps,
-  getPlaces,
-  Window,
-  TaskDataGridTable,
-  Tasks,
   FilterFields,
+  getPlaces,
+  SortFields,
+  Tasks,
+  TaskDataGridTable,
+  Window,
 } from 'react-components';
 import { AppControllerContext, ResourcesContext } from '../app-contexts';
 import { AppEvents } from '../app-events';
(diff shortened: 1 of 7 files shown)


mayman99:
Sorting by state and by assignee doesn't work as i excepted it to work, this one is by state
image
and this one is sorted by assignee 
image

packages/dashboard/src/components/tasks/tasks-app.tsx
```diff
@@ -105,6 +108,14 @@ export const TasksApp = React.memo(
           return;
         }
 
+        let orderBy: string = '-unix_millis_start_time';
```
mayman99, line 111: do we only support server sorting by time?

aaronchongth: This is the default value, which is sorting by descending start time. The subsequent lines change the value of `orderBy`

aaronchongth:
Thanks for the quick review, I took a look at it by loading the shared data dump, and I can recreate the issue. I think I may have messed something up when I dumped out my migrated database, I'm currently using live data from a simulation and everything seems to be working well, including `assigned_to` and `status`.

https://user-images.githubusercontent.com/5383623/205229538-cc275985-427d-43f8-90cf-d5301a6f6a46.mp4

aaronchongth:
@mayman99  Ah I figured out what's going on now! During the database migration, it only creates empty columns for `assigned_to` and `state`, without copying over the data from `data`, which is a json and originally those columns were populated from that json.

As of now, the reason why the data dump does not sort, is due to those empty database columns. We were essentially sorting empty spaces.

I'll see if the database migration supports copying data over as well.

aaronchongth:
Just managed to get it working! Will be sending over the modified database in slack, where all the fields are properly populated.

https://user-images.githubusercontent.com/5383623/205282157-f47db76c-d938-4705-a281-356e72062ea1.mp4

mayman99 (approved):

mayman99:
lgtm!

PR #672, opened 2023-02-06
api-server: scheduling backend

koonpeng:
## What's new

Initial work on a scheduler backend. 2 scheduler is considered while working on this

1. https://github.com/dbader/schedule
    * Minimum scheduling system with "human readable syntax".
1. https://github.com/agronholm/apscheduler
    * Full featured scheduling system with managed data store.

Option 1 is chosen in the end, mainly because option 2 does not support async programming (there is an ongoing rewrite for async on the master branch).

The following endpoints are added:

* `GET /scheduled_tasks` List all scheduled tasks, there are no querying capabilities atm.
* `GET /scheduled_tasks/{task_id}` Get a single scheduled task.
* `POST /scheduled_tasks` Create a new scheduled task.
* `DELETE /scheduled_tasks/{task_id}` Delete a scheduled task.

## Self-checks


packages/api-server/api_server/routes/tasks/scheduled_tasks.py
```diff
@@ -0,0 +1,115 @@
...
+class PostScheduledTaskRequest(BaseModel):
+    task_request: TaskRequest
+    schedules: list[
+        ttm.ScheduledTaskSchedulePydantic
+    ]  # a scheduled task can have multiple schedules
+
+
+async def schedule_task(task: ttm.ScheduledTask, task_repo: TaskRepository):
+    await task.fetch_related("schedules")
+    jobs = [x.to_job() for x in task.schedules]
```
aaronchongth, line 30: Maybe put this in a try except block since `to_job` raises an exception in the case of an unsupported period?

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
```diff
@@ -0,0 +1,115 @@
...
+
+    async def run():
+        await post_dispatch_task(req, task_repo)
+        task.last_ran = datetime.now()
+        await task.save()
+
+    for j in jobs:
+        j.do(lambda: asyncio.get_event_loop().create_task(run())).tag(f"task_{task.pk}")
+
+    # Job have a operator overload that sorts based on the next run
```
aaronchongth, line 44:
```suggestion
    # Job has an operator overload that sorts based on the next run
```

aaronchongth (changes requested):
Thanks for putting this together! Just a few nitpicks, but all looks good in general!

I just noticed the use of `pydantic_model_creator`, and I really like to have that as a single source of truth (tied to the version of `tortoise`) for most of our models that don't yet require any schema validation on RMF's C++ side. Do you think there would be any drawbacks for us if we continue using `pydantic_model_creator` for all our newer routes, without worrying much about syncing with `rmf_api_msgs`?

We can then avoid something like what we did with favorite tasks, [write schemas](https://github.com/open-rmf/rmf_api_msgs/pull/33), [implement feature](https://github.com/open-rmf/rmf-web/pull/676), and [handle all the API changes](https://github.com/open-rmf/rmf-web/pull/686).

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
koonpeng: `kp/scheduling-frontend` actually addressed this.

aaronchongth (approved):
Awesome! Thanks for the heads up, LGTM

PR #681, opened 2023-02-23
Fix react-components test failure on win10

aaronchongth:
## What's new

```
Chrome 110.0.0.0 (Windows 10) RobotInfo information renders correctly FAILED
	Expected function not to throw, but it threw TestingLibraryElementError: Unable to find an element with the text: 12/31/1969, 7:00:00 PM. This could be because the text is broken up by multiple elements. In this case, you can provide a function for your text matcher to make your matcher more flexible.
```

## Self-checks


packages/react-components/karma.conf.js
```diff
@@ -2,6 +2,8 @@
 
 process.env.CHROME_BIN = require('puppeteer').executablePath();
 
+process.env.TZ = 'Asia/Singapore';
```
koonpeng, line 5: I suspect this might be the thing that fixes the tests.

packages/react-components/lib/robots/robot-info.spec.tsx
```diff
@@ -19,7 +19,17 @@ describe('RobotInfo', () => {
...
-    expect(() => root.getByText(new Date(0).toLocaleString())).not.toThrow();
+    expect(() =>
+      root.getByText((content, node) => {
+        if (!node) {
+          return false;
+        }
+        const hasText = (node) => node.textContent === new Date(0).toLocaleString();
+        const nodeHasText = hasText(node);
+        const childrenDontHaveText = Array.from(node.children).every((child) => !hasText(child));
+        return nodeHasText && childrenDontHaveText;
```
koonpeng, line 30: Not too sure if we should be concerned about `childrenDontHaveText`, it seems abit too verbose to test for text, I would expect something like this to be built into testing library if such tests are the intended way to use it.

packages/react-components/karma.conf.js
aaronchongth: I was thinking the same thing, hence https://github.com/open-rmf/rmf-web/pull/681/commits/66a5338d6d3c22b129d4dc1e4a9daf82839cb4e8, however it still failed.

I kept this env variable in as it was mentioned as a general practice for CI.

koonpeng (commented):
I think 1 more thing worth trying is updating testing-library. If that doesn't work as well then I'm fine with this change.

aaronchongth:
> I think 1 more thing worth trying is updating testing-library. If that doesn't work as well then I'm fine with this change.

@koonpeng, ah unfortunately it looks like it still fails after updating `testing-library/*` and using our old tests. I've kept the updates in this PR and switched to the newer tests

Angatupyry (approved):
Certainly, as @koonpeng  mentioned the goal of the library is to help you write tests that resemble how a user would use your application, but test library also works with the DOM tree rendered. 
So... I think we are facing point 4 of this [link](https://polvara.me/posts/five-things-you-didnt-know-about-testing-library). 
So the proposed solution makes sense to me.

PR #686, opened 2023-03-06
Fix task_favorite model

Angatupyry:
Set name, description, and category fields as required.


aaronchongth (changes requested):
Thanks for getting back so quickly! I've only just noticed that our routes are `tasks/favorite_task`, I think it would be better if we kept the routes separate from `tasks`, I like the separation of concern that KP implemented for the scheduling backend, https://github.com/open-rmf/rmf-web/pull/672/files#diff-49343ae99efda11415c743d6fc2ccf51977e8ece61cb8a179369ccfc09de6fa5R92, using a prefix

So along with this refactor, perhaps we should have favorite tasks be it's own file under `routes/tasks/favorite_tasks.py`, and have its own route `/favorite_tasks`, this will help us a ton down the road when we need to iterate or test more

this is looking great! once this gets merged we can cherry pick this and https://github.com/open-rmf/rmf-web/pull/676 back into `main`

Angatupyry:
> Thanks for getting back so quickly! I've only just noticed that our routes are `tasks/favorite_task`, I think it would be better if we kept the routes separate from `tasks`, I like the separation of concern that KP implemented for the scheduling backend, https://github.com/open-rmf/rmf-web/pull/672/files#diff-49343ae99efda11415c743d6fc2ccf51977e8ece61cb8a179369ccfc09de6fa5R92, using a prefix
> 
> So along with this refactor, perhaps we should have favorite tasks be it's own file under `routes/tasks/favorite_tasks.py`, and have its own route `/favorite_tasks`, this will help us a ton down the road when we need to iterate or test more
> 
> this is looking great! once this gets merged we can cherry pick this and #676 back into `main`

Done!
Thanks for the quick review!

packages/api-server/api_server/repositories/tasks.py
```diff
@@ -24,6 +23,20 @@
...
+    name: str
+    unix_millis_earliest_start_time: datetime
+    priority: Dict
+    category: str
+    description: Dict
+    user: str
+
+
+class TaskFavoriteOut(TaskFavoriteIn):
+    unix_millis_earliest_start_time: int
```
aaronchongth, line 37: Are these being models being used somewhere? I can't seem to find any references to them

aaronchongth (changes requested):
Looks great! I think I may have found some stale unused code, let me know if it is used anywhere. But otherwise, LGTM!

aaronchongth (approved):
Thanks! LGTM!

PR #688, opened 2023-03-09
fix wrong version of `datamodel-codegen` being used

koonpeng:
## What's new

fix wrong version of `datamodel-codegen` being used. Also removed the check that rmf is not sourced, the check is added due to colcon's "include pollution bug/feature" which is now largely mitigated as rmf is using "isolated" include dir (e.g. `/opt/ros/humble/include/rmf_door_msgs/rmf_door_msgs`).

## Self-checks


aaronchongth (approved):
This works great! Thanks!

PR #689, opened 2023-03-13
Date column in Tasks section

Angatupyry:
### Date column has been added in taskstate
This field is saved when the user creates a new task. It allows us to know when a task was created.  
This field cannot be updated by the user.

- Related https://github.com/open-rmf/rmf_api_msgs/pull/34

Screenshot from 2023-03-13 16-38-07

## Self-checks


aaronchongth (changes requested):
Upon going through it in detail, I believe we might not need to update `rmf_api_msgs`, and the `pydantic` model as well, just an additional field to the tortoise model is sufficient.

It looks like we just need to modify https://github.com/open-rmf/rmf-web/blob/main/packages/api-server/api_server/repositories/tasks.py#L29-L46, where instead of `ttm.TaskState.update_or_create`, we can [get_or_none](https://tortoise.github.io/models.html#tortoise.models.Model.exists), if it is not none, we copy the existing state's `unix_millis_request_time`, and update the model with that same request time.

Otherwise, we create a new one with the current time stamp for the request time. We can be sure that the first state we get to save is right after the request has been dispatched, since it is saved in [this post call](https://github.com/open-rmf/rmf-web/blob/main/packages/api-server/api_server/routes/tasks/tasks.py#L147-L149) and [this post call](https://github.com/open-rmf/rmf-web/blob/main/packages/api-server/api_server/routes/tasks/tasks.py#L167-L169)

This approach should be used for the upcoming column of `task initiator` too, using the dashboard's current user. Let me know what you think about this approach

Angatupyry:
> Upon going through it in detail, I believe we might not need to update `rmf_api_msgs`, and the `pydantic` model as well, just an additional field to the tortoise model is sufficient.
> 
> It looks like we just need to modify https://github.com/open-rmf/rmf-web/blob/main/packages/api-server/api_server/repositories/tasks.py#L29-L46, where instead of `ttm.TaskState.update_or_create`, we can [get_or_none](https://tortoise.github.io/models.html#tortoise.models.Model.exists), if it is not none, we copy the existing state's `unix_millis_request_time`, and update the model with that same request time.
> 
> Otherwise, we create a new one with the current time stamp for the request time. We can be sure that the first state we get to save is right after the request has been dispatched, since it is saved in [this post call](https://github.com/open-rmf/rmf-web/blob/main/packages/api-server/api_server/routes/tasks/tasks.py#L147-L149) and [this post call](https://github.com/open-rmf/rmf-web/blob/main/packages/api-server/api_server/routes/tasks/tasks.py#L167-L169)
> 
> This approach should be used for the upcoming column of `task initiator` too, using the dashboard's current user. Let me know what you think about this approach

Hey, Aaron. I think your approach is ok and that's exactly what I'm trying to do here
```
async def save_task_state(self, task_state: TaskState) -> None:
        datetime_request_time = (
            datetime.fromtimestamp(task_state.unix_millis_request_time / 1000)
            if task_state.unix_millis_request_time
            else datetime.now()
        )
        task_state.unix_millis_request_time = (
            task_state.unix_millis_request_time
            if task_state.unix_millis_request_time
            else round(time.time() * 1000)
        )
        await ttm.TaskState.update_or_create(
            {
                "data": task_state.json(),
                "category": task_state.category.__root__
                if task_state.category
                else None,
                "assigned_to": task_state.assigned_to.name
                if task_state.assigned_to
                else None,
                "unix_millis_start_time": task_state.unix_millis_start_time
                and datetime.fromtimestamp(task_state.unix_millis_start_time / 1000),
                "unix_millis_finish_time": task_state.unix_millis_finish_time
                and datetime.fromtimestamp(task_state.unix_millis_finish_time / 1000),
                "status": task_state.status if task_state.status else None,
                "unix_millis_request_time": datetime_request_time,
            },
            id_=task_state.booking.id,
        )
```
But I also need to update the `data` field so that it can be displayed later in the dashboard. 
And regarding displaying in the datagrid, how could we access the field without it existing [here](https://github.com/open-rmf/rmf-web/blob/main/packages/api-server/api_server/models/rmf_api/task_state.py#L172)? The API client creates the interfaces depending on that class. 
I consider that to save the data in the database the tortoise field is enough but to show it on the dashboard I consider that we also need the field in [rmf_api](https://github.com/open-rmf/rmf-web/blob/main/packages/api-server/api_server/models/rmf_api/task_state.py#L172) (which does not necessarily have to depend on rmf_msgs_api and)

packages/api-server/api_server/models/rmf_api/task_state.py
```diff
@@ -181,6 +181,7 @@ class TaskState(BaseModel):
         None, description="Which agent (robot) is the task assigned to"
     )
     status: Optional[Status] = None
+    unix_millis_request_time: Optional[int] = None
```
aaronchongth, line 184: I think this should be in `Booking`, as it is part of the task information that does not change throughout the entire task and is related to the task request itself

aaronchongth: https://github.com/open-rmf/rmf_api_msgs/pull/34 would probably need to be updated as well, https://github.com/open-rmf/rmf_api_msgs/blob/main/rmf_api_msgs/schemas/task_state.json#L110

packages/api-server/api_server/repositories/tasks.py
```diff
@@ -27,6 +28,16 @@ def __init__(self, user: User):
         self.user = user
 
     async def save_task_state(self, task_state: TaskState) -> None:
+        datetime_request_time = (
+            datetime.fromtimestamp(task_state.unix_millis_request_time / 1000)
+            if task_state.unix_millis_request_time
+            else datetime.now()
+        )
```
aaronchongth, line 35: The task state is published by `rmf_fleet_adapter::TaskManager`, which at the moment, does not populate the `unix_millis_request_time` field, https://github.com/open-rmf/rmf_ros2/blob/main/rmf_fleet_adapter/src/rmf_fleet_adapter/TaskManager.cpp#L459, which means we will always be setting `task_state.unix_millis_request_time` as the current time. We will need to query the previous task state of this task ID, and obtain the `previous_task_state.unix_millis_request_time` to be used here instead

packages/api-server/api_server/repositories/tasks.py
```diff
@@ -27,6 +28,16 @@ def __init__(self, user: User):
...
+        task_state.unix_millis_request_time = (
+            task_state.unix_millis_request_time
+            if task_state.unix_millis_request_time
+            else round(time.time() * 1000)
+        )
```
aaronchongth, line 40: The above comment applies here as well, we will need to use `previous_task_state.unix_millis_request_time` instead

aaronchongth (changes requested):
Doh! Thanks for explaining, I totally forgot that we are using the `data` JSON field, we should fix that at some point, I'll add a backburner task for that.

I've left some comments and suggestions, let me know what you think

packages/api-server/api_server/repositories/tasks.py
Angatupyry: I completely agree with you. 
But justly the value of `task_state.unix_millis_request_time` will be the previous value if the method is "update", and if it is "create" it will come null. 
So that's the control I'm doing there. If `task_state.unix_millis_request_time` has a value then keep using that value otherwise use datetime.now()

packages/api-server/api_server/models/rmf_api/task_state.py
Angatupyry: Field added to booking according to suggestion!

packages/api-server/api_server/repositories/tasks.py
aaronchongth: yup, however we are still applying the keyword https://github.com/open-rmf/rmf-web/pull/689/files#diff-6e62b347b0bbff536cbf2a85cf360c53355a713c6f5bf02ef4aa1519b56db94bR55, which will always update with the values from `datetime_request_time`

request time will always increase whenever a new task state comes in,

image

you can verify it by adding these 2 lines above [here](https://github.com/open-rmf/rmf-web/blob/feature/add-unix-millis-request-time/packages/api-server/api_server/repositories/tasks.py#L41), 
```python
        print('datetime_request_time: {}'.format(datetime_request_time))
        print('in json: {}'.format(task_state.booking.unix_millis_request_time))
```

Angatupyry: You are right!
Thanks for the thorough review.

aaronchongth (approved):
Thanks for the quick rework! Tested and working great, LGTM

PR #690, opened 2023-03-17
Fix/request date export

aaronchongth:
## What's new

Related to https://github.com/open-rmf/rmf-web/pull/689

* Adds Date column to minimal export
* Refactored task repository `save_task_state`
  * creating a new task state model explicitly setting request time, if it is a new task
  * updating all other fields when task already exists, leaving out request time and ID

To check exports, go to download button and select minimal export

To check if the request time stays the same, add these lines after [here, outside of the else scope](https://github.com/open-rmf/rmf-web/pull/690/files#diff-6e62b347b0bbff536cbf2a85cf360c53355a713c6f5bf02ef4aa1519b56db94bR72)

```python
        # debug
        debug_task_state = await DbTaskState.get_or_none(id_=task_state.booking.id)
        if debug_task_state is None:
            print('oh no something went wrong')
        else:
            print('request_time: {}'.format(debug_task_state.unix_millis_request_time))
```

## Self-checks

### DIFF
--- a/packages/api-server/api_server/repositories/tasks.py
+++ b/packages/api-server/api_server/repositories/tasks.py
@@ -28,35 +28,59 @@ class TaskRepository:
         self.user = user
 
     async def save_task_state(self, task_state: TaskState) -> None:
-        previous_task_state = await DbTaskState.get_or_none(id_=task_state.booking.id)
-        datetime_request_time = datetime.now()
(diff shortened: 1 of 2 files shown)


Angatupyry:
LGTM!
I think it was something we needed when we started adding those fields that should not be updated. So, thanks for these changes!
The only question I have is whether we should separate the update and create methods in different endpoints, but let's discuss that at our next meeting.

PR #692, opened 2023-03-17
Feature/task initiator

Angatupyry:
## What's new
Related: https://github.com/open-rmf/rmf_api_msgs/pull/34

### Initiator column has been added in taskstate
This field is saved when the user creates a new task. It allows us to know who creates a task.
The user cannot update this field.

Screenshot from 2023-03-17 08-57-44

## Self-checks


aaronchongth (approved):
Sorry for the delay! This looks great and works well, export is working well too, thanks!

PR #694, opened 2023-03-21
Alerts rework

aaronchongth:
## What's new

[basic_look.webm](https://user-images.githubusercontent.com/5383623/233531494-f6f4a95b-9bfa-49c8-8fb3-ab96b65120c9.webm)

[acknowledge_dismiss_auto_close.webm](https://user-images.githubusercontent.com/5383623/233466054-5f1fca62-03d2-441f-8e30-65ddc618ed0e.webm)

### Backend for alerts
* generic models for alerts and details, to handle both task and robot alerts
* acknowledgement from users

Test with `localhost:8000/docs` after spinning up the server
* create alert
* get all alerts
* get alert via id
* acknowledge alert via id

### Frontend for alerts
* generic alert dialog react component
* task and robot alert handlers

### Expected behaviors
* Task alerts are derived from completion of tasks or if the logs contain errors
* Robot alerts are derived from having an error robot status, the issues under the robot state will be displayed
* If the robot which triggered an error has resolved the issue, without any user acknowledging it (solving it), the alert dialog will automatically go away (so as to not flood the front end with dialogs if the user does not use it for days)
* When a user acknowledges an alert, other user's acknowledgement button will be disabled, and it will mention which user has acknowledged it
* As time goes by, information on the alert dialog gets updated, for example robot issues, last updated time, etc
* Dismissing just closes the alert, which will not show up for this user again

### Next step
* Create a button that opens the list of alerts (both acknowledged or dismissed), like a notification list

## Self-checks


packages/dashboard/src/components/alert-store.tsx
```diff
@@ -17,90 +19,49 @@ export interface AlertToDisplay extends RobotWithTask {
...
+          };
+        });
+      } else if (alert.category === 'fleet' || alert.category === 'robot') {
+        setRobotAlerts((prev) => {
+          return {
+            ...prev,
+            [alert.id]: alert,
+          };
+        });
+      }
```
Angatupyry, line 46: We can use the Switch statement here to be more clear. 
```suggestion
      switch (alert.category) {
        case 'task':
          setTaskAlerts((prev) => ({ ...prev, [alert.id]: alert }));
          break;
        case 'fleet':
        case 'robot':
          setRobotAlerts((prev) => ({ ...prev, [alert.id]: alert }));
          break;
        default:
        // default case
      }

File: packages/dashboard/src/components/alert-store.tsx [hunk 2]
```diff
@@ -17,90 +19,49 @@ export interface AlertToDisplay extends RobotWithTask {
...
-      setFleets(
-        fleets.reduce<string[]>((acc, f) => {
-          if (f.name) {
-            acc.push(f.name);
-          }
-          return acc;
-        }, []),
-      );
+    const sub = rmf.alertObsStore.subscribe(async (alert) => {
+      if (alert.category === 'task') {
```
Angatupyry, line 32: Does it make sense for us to create a [Typescript enum](https://www.typescriptlang.org/docs/handbook/enums.html#string-enums) if the api client didn't provide one for [categories](https://github.com/open-rmf/rmf-web/pull/694/files#diff-ba1b98e67ee0daeb8df03b15ead4fd9cb68dc27678160ac40cec2aa24c3ef004R13-R17)?

Angatupyry (changes requested):
This looks really great, Aaron!
Thanks for working on it. 
I left some comments on the frontend. 
I don't have any comments on the backend. LGTM!

packages/react-components/lib/alert-dialog.tsx
```diff
@@ -1,100 +1,104 @@
...
 export const AlertDialog = React.memo((props: DialogAlertProps) => {
-  const classes = useStyles();
+  const useStyles = makeStyles((theme: Theme) =>
+    createStyles({
+      textField: {
+        background: theme.palette.background.default,
+        pointerEvents: 'none',
+      },
+    }),
+  );
```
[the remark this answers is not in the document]
aaronchongth: Thanks for sharing this! I never knew about it, fixed https://github.com/open-rmf/rmf-web/pull/694/commits/f92e50c56ef6ba2dff1da379aa0f24ef0de4a5f3

packages/dashboard/src/components/alert-store.tsx
aaronchongth: Yup, good idea. Ideally it would have come from the API, but alas since we are using the pydantic creator from tortoise, we don't get the luxury of handling enums from API like `TaskState.Status`. I've made the changes https://github.com/open-rmf/rmf-web/pull/694/commits/207bd8d94f7d6845ef7f2c28ab3843eb6824f1e1, and commented that it needs to adhere to the tortoise model

packages/dashboard/src/components/alert-store.tsx
aaronchongth: Good idea, this is much cleaner. https://github.com/open-rmf/rmf-web/pull/694/commits/7fddd62cdf8e9e2e69122a2005adcc8543a23f4e

Angatupyry (approved):
LGTM!

Issue #629, opened 2022-07-13
Good to have features in client-side dashboard


youliangtan:
Per VC,
* Clicking on the robots in the tables, the maps do not center on that robot at the moment.
* New tasks created by current client do not show up on other dashboard clients readily, unless the refresh button is clicked.

These 2 notes should not block this PR from merging, new feature requests would be opened.

_Originally posted by @aaronchongth in https://github.com/open-rmf/rmf-web/issues/623#issuecomment-1181457594_

koonpeng:
> Clicking on the robots in the tables, the maps do not center on that robot at the moment.
PR for this #630.

> New tasks created by current client do not show up on other dashboard clients readily, unless the refresh button is clicked.
This will require new endpoints in api-server, currently wip at branch `feat/active-tasks`.

PR #697, opened 2023-03-24
Task creation as big button accessible from any tab

Angatupyry:
## What's new

- Added button accessible from any tab and allow users to create tasks
- Remove "Add icon" from task section 

[chrome-capture-2023-2-24.webm](https://user-images.githubusercontent.com/37310205/227580297-51da11b8-8b8c-456e-9c21-295a65f21bc0.webm)

[chrome-capture-2023-2-24 (1).webm](https://user-images.githubusercontent.com/37310205/227580290-b151d4a5-c54a-4c4e-8ca2-175a66c8aa83.webm)

## Self-checks

### DIFF
--- a/packages/dashboard-e2e/tests/ui-interactions/submit-task.test.ts
+++ b/packages/dashboard-e2e/tests/ui-interactions/submit-task.test.ts
@@ -4,7 +4,7 @@ describe('submit task', () => {
   it('can submit loop task', async () => {
     const appBar = await getAppBar();
     await (await appBar.$('button[aria-label="Tasks"]')).click();
-    await (await $('button[aria-label="Create Task"]')).click();
+    await (await appBar.$('button[aria-label="new task"]')).click();
     await (await $('#task-type')).click();
     const getLoopOption = async () => {
       const options = await $$('[role=option]');
--- a/packages/dashboard/src/components/app-contexts.tsx
+++ b/packages/dashboard/src/components/app-contexts.tsx
@@ -13,15 +13,12 @@ export interface AppController {
   updateSettings: (settings: Settings) => void;
   showAlert: (severity: AlertProps['severity'], message: string, autoHideDuration?: number) => void;
   setExtraAppbarIcons: (node: React.ReactNode) => void;
+  forceTaskQueueTable: number;
+  updateTaskQueueTable: (forceTaskQueueTable: number) => void;
 }
 
 export interface Tooltips {
   showTooltips: boolean;
-}
-
-export interface RefreshTasksTable {
-  forceRefreshTask: number;
-  setForceRefreshTask: (forceRefreshTask: number) => void;
 }
 
 export const TooltipsContext = React.createContext<Tooltips>({
@@ -32,11 +29,8 @@ export const AppControllerContext = React.createContext<AppController>({
   updateSettings: () => {},
   showAlert: () => {},
   setExtraAppbarIcons: () => {},
-});
-
-export const RefreshTaskTableContext = React.createContext<RefreshTasksTable>({
-  forceRefreshTask: 0,
-  setForceRefreshTask: () => {},
+  forceTaskQueueTable: 0,
+  updateTaskQueueTable: () => {},
 });
 
 export const AppConfigContext = React.createContext<AppConfig>(appConfig);
--- a/packages/dashboard/src/components/app-contexts.tsx
+++ b/packages/dashboard/src/components/app-contexts.tsx
@@ -19,6 +19,11 @@ export interface Tooltips {
   showTooltips: boolean;
 }
 
+export interface RefreshTasksTable {
+  forceRefreshTask: number;
+  setForceRefreshTask: (forceRefreshTask: number) => void;
+}
+
 export const TooltipsContext = React.createContext<Tooltips>({
   showTooltips: true,
 });
@@ -27,6 +32,11 @@ export const AppControllerContext = React.createContext<AppController>({
   updateSettings: () => {},
   showAlert: () => {},
   setExtraAppbarIcons: () => {},
+});
+
+export const RefreshTaskTableContext = React.createContext<RefreshTasksTable>({
+  forceRefreshTask: 0,
+  setForceRefreshTask: () => {},
 });
 
 export const AppConfigContext = React.createContext<AppConfig>(appConfig);
--- a/packages/dashboard/src/components/app-contexts.tsx
+++ b/packages/dashboard/src/components/app-contexts.tsx
@@ -13,8 +13,6 @@ export interface AppController {
   updateSettings: (settings: Settings) => void;
   showAlert: (severity: AlertProps['severity'], message: string, autoHideDuration?: number) => void;
   setExtraAppbarIcons: (node: React.ReactNode) => void;
-  forceTaskQueueTable: number;
-  updateTaskQueueTable: (forceTaskQueueTable: number) => void;
 }
 
 export interface Tooltips {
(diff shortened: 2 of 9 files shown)


aaronchongth:
Per VC, modify e2e test to create task from Robot tab and Task tab, to verify that we can create tasks from different parts of the dashboard.

packages/dashboard/src/components/app-contexts.tsx
```diff
@@ -19,6 +19,11 @@ export interface Tooltips {
   showTooltips: boolean;
 }
 
+export interface RefreshTasksTable {
+  forceRefreshTask: number;
+  setForceRefreshTask: (forceRefreshTask: number) => void;
+}
+
```
aaronchongth, line 26: What do you think about adding these into `AppController`, instead of creating a new context? We can name it `updateTaskQueueTable`, similar to `updateSettings` too

let me know if there's something I might have missed that would make it not feasible

packages/dashboard-e2e/tests/ui-interactions/submit-task.test.ts
```diff
@@ -4,7 +4,7 @@ describe('submit task', () => {
   it('can submit loop task', async () => {
     const appBar = await getAppBar();
     await (await appBar.$('button[aria-label="Tasks"]')).click();
-    await (await $('button[aria-label="Create Task"]')).click();
+    await (await appBar.$('button[aria-label="new task"]')).click();
```
aaronchongth, line 7: thanks! could you add an additional e2e test that requests a new task from the `Robot` tab, to verify that we can make new requests from anywhere in the dashboard

aaronchongth (changes requested):
Thanks for the huge refactor! This looks awesome. I have a few comments and questions regarding the contexts, but otherwise it's looking good!

The refresh/update mechanism is sweet, it fixes https://github.com/open-rmf/rmf-web/issues/629. But alas what I meant during our meeting was regarding auto-refreshing the task queue table while idling automatically, perhaps every 5 seconds or something. I created a new ticket for it, https://github.com/open-rmf/rmf-web/issues/699, don't worry about it in this PR.

packages/dashboard/src/components/app-contexts.tsx
Angatupyry: This suggestion makes total sense. 
Allow us to avoid adding another context provider in app-base.tsx
Thanks, Aaron!

packages/dashboard-e2e/tests/ui-interactions/submit-task.test.ts
Angatupyry: Oh, man. I'm sorry, I forgot that we mentioned it per VC. 
I've added another file to test it to make it more organized.

aaronchongth (changes requested):
Thanks for making the changes so quickly! Overall the component migration looks good!

I'm sorry I didn't think of this sooner, but I believe we should use `AppEvents` for triggering a refresh on the task queue table instead of the app controller context.

For some reference, we can check out how `AppEvents.robotSelect` is used to trigger map centering or populating `RobotInfo`, as an example. I think we can continue your approach with using a counter in `AppEvents`, maybe called `AppEvents.refreshTaskQueueTableCounter`, subscribe to it, and refresh the table based on that, while calling `AppEvents.refreshTaskQueueTableCounter.next(...)`, from creating a task

Let me know what you think!

Angatupyry:
> Thanks for making the changes so quickly! Overall the component migration looks good!
> 
> I'm sorry I didn't think of this sooner, but I believe we should use `AppEvents` for triggering a refresh on the task queue table instead of the app controller context.
> 
> For some reference, we can check out how [`AppEvents.robotSelect`](https://github.com/open-rmf/rmf-web/blob/main/packages/dashboard/src/components/app-events.ts#L9) is used to trigger map centering or populating `RobotInfo`, as an example. I think we can continue your approach with using a counter in `AppEvents`, maybe called `AppEvents.refreshTaskQueueTableCounter`, subscribe to it, and refresh the table based on that, while calling `AppEvents.refreshTaskQueueTableCounter.next(...)`, from creating a task
> 
> Let me know what you think!

This suggestion was really good. 
Definitely a better alternative. 
Please take a look at the new changes. 
Thanks, Aaron!

packages/dashboard/src/components/appbar.tsx
```diff
@@ -105,6 +124,11 @@ export const AppBar = React.memo(({ extraToolbarItems }: AppBarProps): React.Rea
   const safeAsync = useAsync();
   const [brandingIconPath, setBrandingIconPath] = React.useState<string>('');
   const [settingsAnchor, setSettingsAnchor] = React.useState<HTMLElement | null>(null);
+  const [openCreateTaskForm, setOpenCreateTaskForm] = React.useState(false);
+  const [placeNames, setPlaceNames] = React.useState<string[]>([]);
+  const [workcells, setWorkcells] = React.useState<string[]>();
+  const [favoritesTasks, setFavoritesTasks] = React.useState<TaskFavorite[]>([]);
+  const [currentQueueTaskTableValue, setCurrentQueueTaskTableValue] = React.useState(0);
```
aaronchongth, line 131:
```suggestion
  const [refreshTaskQueueTableCount, setRefreshTaskQueueTableCount] = React.useState(0);
```

packages/dashboard/src/components/app-events.ts
```diff
@@ -8,4 +8,5 @@ export const AppEvents = {
   ingestorSelect: new Subject<Ingestor | null>(),
   robotSelect: new Subject<[fleetName: string, robotName: string] | null>(),
   taskSelect: new Subject<TaskState | null>(),
+  refreshTaskQueueTableCounter: new Subject<number>(),
```
aaronchongth, line 11:
```suggestion
  refreshTaskQueueTableCount: new Subject<number>(),
```

packages/dashboard/src/components/appbar.tsx
```diff
@@ -116,13 +140,133 @@ export const AppBar = React.memo(({ extraToolbarItems }: AppBarProps): React.Rea
...
+      if (!favoriteTask.id) {
+        throw new Error('Id is needed');
+      }
+
+      await rmf.tasksApi.deleteFavoriteTaskFavoriteTasksFavoriteTaskIdDelete(favoriteTask.id);
+      AppEvents.refreshTaskQueueTableCounter.next(currentQueueTaskTableValue + 1);
+    },
+    [rmf, currentQueueTaskTableValue],
+  );
+
```
aaronchongth, line 267: nit: extra line here

aaronchongth (changes requested):
Thanks for making the changes! Everything looks good and works fine, except for a few minor name change suggestions.

I like the use of regions! Did you manually add it in, or was it generated using a vscode extension?

Angatupyry:
> Thanks for making the changes! Everything looks good and works fine, except for a few minor name change suggestions.
> 
> I like the use of regions! Did you manually add it in, or was it generated using a vscode extension?

"Region" comes from C#. I was using "Visual Studio Enterprise" and that allowed it as an inherent thing. When I started working with JavaScript and started using VSCode I looked it up and it didn't have "regions" built in naturally. The way to add it is manually and as a comment. I didn't like that very much and got angry for a few moments because I liked the way the regions ordered the piece of code. 
After some time I understood that one has to be tidy enough in its code and it is not necessary to fill with regions for someone else to understand it, however for some cases, as is this PR, I see it pertinent to add for better organization of sections in the code.

aaronchongth (approved):
Thanks! LGTM

PR #700, opened 2023-03-30
scheduling frontend

koonpeng:
## What's new

UI for creating scheduled task and rendering it in a calendar. Also includes changes in the backend to support querying scheduled tasks.

## Self-checks


aaronchongth (changes requested):
This looks awesome! Thanks for handling this massive feature. I noticed a few things that need modifications on regarding the frontend, that I thought would be best to mention first. I will then continue to review the code while you can check out my comments.

Most of these comments will be based on what our figma design, I'll DM a link over shortly.

* The scheduling feature should be a second sub tab, under `Tasks`, called `Calendar`, while our original table will be under the first sub tab, called `Queue`. Users will need to select the sub tabs to change the view between our task queue table and the calendar view of scheduled tasks
* For task creation, it looks like daily scheduling is turned on by default, unless a user deselects all days or checks `Now`. Scheduling would ideally be a decision made only at the end of creating a task (whether to schedule or send it out once), which should open up a separate dialog that contains the components you've written so far for selecting days-of-the-week, and future features of selecting ending dates, etc. This will help keep the task creation form less cluttered and easy for us to add more features next.

Let me know if these modifications is possible with our current implementation, thanks!

Angatupyry:
The api-server error is because pydantic warning `Access to a protected member _id of a client class (protected-access)`
Why we called _id instead of id [here](https://github.com/open-rmf/rmf-web/pull/700/files#diff-853e7ec9f0edce76e47ec2a45754ae2c7d8584700ac7b6bf79132a87bd034d40L48-R51)  ?

aaronchongth:
@Angatupyry, thank you so much for the help! I can verify that the deletion bug has been fixed.

> This has O(n) complexity, which implies that the execution time when deleting schedules is proportional to the size of the input. It doesn't seem very inefficient to me unless we are considering a large amount of data.

Yup, I'm willing to live with that. We can further optimize this when we implement end-date for schedules, and have to handle schedules that have already past.

> The api-server error is because pydantic warning Access to a protected member _id of a client class (protected-access)
Why we called _id instead of id [here](https://github.com/open-rmf/rmf-web/pull/700/files#diff-853e7ec9f0edce76e47ec2a45754ae2c7d8584700ac7b6bf79132a87bd034d40L48-R51) ?

Gotcha! I was actually referring to a stalling error, but it looks like it may have already been resolved, perhaps it was a github runner issue. The member is called `_id` to be considered a protected member of the `Model`, and hence will the interface will not include the member, https://github.com/open-rmf/rmf-web/blob/kp/scheduling-frontend/packages/api-client/lib/openapi/api.ts#L299-L330

I've added a getter to access the protected member instead, this resolves the linting error

aaronchongth (approved):
I'm liking how this is looking! Thanks for all the help so far @koonpeng and @Angatupyry! LGTM

@Angatupyry can you do another round of review for this? Probably not a good idea for me to approve and just merge what I wrote :rofl:

Angatupyry:
> I'm liking how this is looking! Thanks for all the help so far @koonpeng and @Angatupyry! LGTM
> 
> @Angatupyry can you do another round of review for this? Probably not a good idea for me to approve and just merge what I wrote 🤣

haha! Thanks, Aaron! 
I will!

Angatupyry (approved):
LGTM!

Angatupyry:
@aaronchongth 
image
The only thing I could say is that maybe it doesn't make much sense that the start date could be less than the current date. 
Perhaps it could be restricted as a minimum acceptable to today. 
But it is a detail that does not affect functionality.

aaronchongth:
Yeah I understand what you mean. Well in the spirit of mimicking a modern calendar system, we should also allow handling of schedules in the past then. Thanks for the work and review from all! Merging it! :boom:

PR #703, opened 2023-04-17
Show up task summary when user clic task row

Angatupyry:
## What's new
Popup dialog to show task summary

- Removed microapps (Task logs and Task details) from the top right of the task section

- Added a task summary with the most relevant data
- Task details and Task logs are now displayed by clicking a button in the summary component.

[chrome-capture-2023-3-17.webm](https://user-images.githubusercontent.com/37310205/232598385-1737d969-1b9c-4d8a-932e-d352cc1446a3.webm)

## Self-checks


packages/dashboard/src/components/tasks/task-inspector.tsx
```diff
@@ -0,0 +1,243 @@
...
+import { AppEvents } from '../app-events';
+import { RmfAppContext } from '../rmf-app';
+import { TaskInfo } from 'react-components';
+import { UserProfileContext } from 'rmf-auth';
+import { Enforcer } from '../permissions';
+import { TaskLogs } from './task-logs';
+
+export interface TableDataGridState {
+  task: TaskState | null;
+  open: boolean;
```
aaronchongth, line 25: Similar to the comment in `task-summary`, we can probably do away with this prop, by setting a state that is default true,
```
const [isOpen, setIsOpen] = React.useState(true);
```
and gets set to false in the `onClose` dialog prop

packages/dashboard/src/components/tasks/task-inspector.tsx
```diff
@@ -0,0 +1,243 @@
...
+}
+
+export function TaskInspector({ task, open, onClose }: TableDataGridState): JSX.Element {
+  const theme = useTheme();
+  const rmf = React.useContext(RmfAppContext);
+  const appController = React.useContext(AppControllerContext);
+
+  const [taskState, setTaskState] = React.useState<TaskState | null>(null);
+  const [taskLogs, setTaskLogs] = React.useState<TaskEventLog | null>(null);
+  const [waypoint, setWaypoint] = React.useState({
```
aaronchongth, line 36: I think we should separate this into `startWaypoint` and `finishWaypoint`, since we are not encapsulating that much information in this object

packages/dashboard/src/components/tasks/task-inspector.tsx
```diff
@@ -0,0 +1,243 @@
...
+          selectedTask ? rmf.getTaskStateObs(selectedTask.booking.id) : of(null),
+        ),
+      )
+      .subscribe(setTaskState);
+    return () => sub.unsubscribe();
+  }, [rmf]);
+
+  React.useEffect(() => {
+    AppEvents.taskSelect.next(task);
+  }, [task]);
```
aaronchongth, line 118: Instead of setting up this workaround chain of `AppEvents.taskSelect`, I think we can make use of https://github.com/open-rmf/rmf-web/blob/main/packages/dashboard/src/components/rmf-app/rmf-ingress.ts#L201, subscribing directly to the ID of the `task` prop, and retrieving the logs using the same API call you used

packages/react-components/lib/tasks/task-info.tsx
```diff
@@ -26,108 +26,25 @@ function InfoValue({ children }: React.PropsWithChildren<unknown>) {
...
   return (
     <StyledDiv>
       <Typography variant="h6" style={{ textAlign: 'center' }} gutterBottom>
-        {task.booking.id}
+        {title ? title : task.booking.id}
       </Typography>
       <Divider />
       <div style={{ marginBottom: theme.spacing(1) }}></div>
       <InfoLine>
         <span>State:</span>
```
aaronchongth, line 45:
```suggestion
        <span>Status:</span>
```

packages/react-components/lib/tasks/task-info.tsx
```diff
@@ -26,108 +26,25 @@ function InfoValue({ children }: React.PropsWithChildren<unknown>) {
...
         <InfoValue>{task.status || 'unknown'}</InfoValue>
       </InfoLine>
```
aaronchongth, line 47: Referring back to figma, I believe we may have shuffled up some of the intended fields, `Start waypoint` and `Finish waypoint` should be below this `Status`/`State` field within the `TaskInfo` components. Migrating the waypoint extraction logic down to this component will probably make `TaskInspector` much cleaner too

packages/dashboard/src/components/tasks/task-inspector.tsx
```diff
@@ -0,0 +1,243 @@
...
+              </DialogTitle>
+            </Grid>
+            <Grid container>
+              <Grid item xs={2} ml={7}>
+                <Typography>Status:</Typography>
+              </Grid>
+              <Grid item xs={3}>
+                <Typography align="right">{task?.status}</Typography>
+              </Grid>
+            </Grid>
```
aaronchongth, line 172: This is probably redundant, since we have the status listed under `TaskInfo`

packages/dashboard/src/components/tasks/task-inspector.tsx
```diff
@@ -0,0 +1,243 @@
...
+            <Grid container>
+              <Grid item xs={2} ml={7}>
+                <Typography>Finish Waypoint:</Typography>
+              </Grid>
+              <Grid item xs={3}>
+                <Typography align="right">{waypoint.finish}</Typography>
+              </Grid>
+            </Grid>
+          </Grid>
```
aaronchongth, line 189: As mentioned in a comment below, we should migrate these fields down to `TaskInfo`, based on our initial designs on figma

packages/dashboard/src/components/tasks/task-inspector.tsx
```diff
@@ -0,0 +1,243 @@
...
+          onClose={onClose}
+          aria-labelledby="alert-dialog-title"
+          aria-describedby="alert-dialog-description"
+          fullWidth
+          maxWidth="md"
+        >
+          <Grid container spacing={2}>
+            <Grid item xs={12}>
+              <DialogTitle id="scroll-dialog-title" align="center">
+                {task?.booking.id}
```
aaronchongth, line 162:
```suggestion
                Task: {task?.booking.id}
```

aaronchongth (changes requested):
This looks amazing! Thanks for putting this together, @Angatupyry, it's really coming together now.

I've some review comments and some suggestions that may help us further clean up the implementation. Let me know what you think!

packages/dashboard/src/components/tasks/task-summay.tsx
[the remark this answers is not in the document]
Angatupyry: > `task-summary.tsx`

Thanks for reviewing this typo!

packages/dashboard/src/components/tasks/task-summay.tsx
```diff
@@ -0,0 +1,182 @@
...
+    textField: {
+      background: theme.palette.background.default,
+      '&:hover': {
+        backgroundColor: theme.palette.background.default,
+      },
+    },
+  }),
+);
+
+const getTaskProgress = (task: TaskState | null): number => {
```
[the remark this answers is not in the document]
Angatupyry: > We should get the task progress using the estimated remaining time, like this implementation https://github.com/open-rmf/rmf-web/blob/main/packages/dashboard/src/components/robots/robot-info-app.tsx#L45-L58

@aaronchongth do you agree with show progress value only if we can calculate it or if it is undefined should we show a message like "Can't calculate task progress"?

packages/dashboard/src/components/tasks/task-summay.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const getTaskMessage = (task: TaskState | null) => {
+    if (!task || !task.phases || !task.active) {
+      return "Can't get task message";
+    }
+
+    if (Object.values(task.phases)[task.active - 1]) {
+      const message = Object.values(task.phases)[task.active - 1].detail;
+
+      return message ? message : "Can't get task message";
+    }
```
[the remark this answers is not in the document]
Angatupyry: > Looks like we might be missing an else case here

You're right!
Thanks!

packages/dashboard/src/components/tasks/task-summay.tsx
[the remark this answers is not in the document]
aaronchongth: I think we can skip the progress bar component if we can't calculate it. A console log message mentioning this is the case would be nice

aaronchongth:
Sorry I forgot to mention this in my main review comment.

I noticed that the `TaskSummary` dialog is not responsive as the task changes, while `TaskInspector` is. It would be nice if we could make the `TaskSummary` subscribe to the task state using `getTaskStateObs` as well, and have the information (colors, phase information, etc) updated over time.

packages/dashboard/src/components/tasks/task-inspector.tsx
Angatupyry: Same answer as above in task summary

packages/react-components/lib/tasks/task-info.tsx
Angatupyry: Perfect!
Nice suggestion!

packages/dashboard/src/components/tasks/task-inspector.tsx
Angatupyry: @aaronchongth the solution I did now make sense? Please correct me if my solution is not in line with your suggestion.

Angatupyry:
> Sorry I forgot to mention this in my main review comment.
> 
> I noticed that the `TaskSummary` dialog is not responsive as the task changes, while `TaskInspector` is. It would be nice if we could make the `TaskSummary` subscribe to the task state using `getTaskStateObs` as well, and have the information (colors, phase information, etc) updated over time.

Please take a look at the solution I did

packages/dashboard/src/components/tasks/task-summay.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+  const getTaskMessage = (task: TaskState | null) => {
+    if (!task || !task.phases || !task.active) {
+      return "Can't get task message";
+    }
+
+    if (Object.values(task.phases)[task.active - 1]) {
+      const message = Object.values(task.phases)[task.active - 1].detail;
+
+      return message ? message : "Can't get task message";
```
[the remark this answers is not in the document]
Angatupyry: I returned it without casting it to string because I would first like to know your opinion about the image I showed you above. Regardless of that, please see if this solution makes sense.

packages/dashboard/src/components/tasks/task-summay.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+    default:
+      return base.palette.background.default;
+  }
+};
+
+export interface TaskSummaryProps {
+  stopShowing: () => void;
+  task: TaskState | null;
+  show: boolean;
```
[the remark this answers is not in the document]
aaronchongth: Hmm, but we will not require multiple `TaskSummary` dialogs to exist at the same time, so this flag doesn't do much, since we are already passing in a callback to be called during `onClose`, we can lump in any closing or toggling logic in that callback.

I'll suggest in a new set of review comments how this can work. Let me know what you think.

packages/dashboard/src/components/tasks/task-summary.tsx
```diff
@@ -0,0 +1,209 @@
...
+    default:
+      return base.palette.background.default;
+  }
+};
+
+export interface TaskSummaryProps {
+  onClose: () => void;
+  task: TaskState | null;
+  show: boolean;
+}
```
aaronchongth, line 72:
```suggestion
export interface TaskSummaryProps {
  onClose: () => void;
  task: TaskState | null;
}
```

packages/dashboard/src/components/tasks/task-summary.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+export const TaskSummary = React.memo((props: TaskSummaryProps) => {
+  const classes = useStyles();
+  const rmf = React.useContext(RmfAppContext);
+
+  const { onClose, show, task } = props;
+
+  const [openTaskDetailsLogs, setOpenTaskDetailsLogs] = React.useState(false);
+  const [taskState, setTaskState] = React.useState<TaskState | null>(null);
```
aaronchongth, line 81:
```suggestion
  const [taskState, setTaskState] = React.useState<TaskState | null>(null);
  const [isOpen, setIsOpen] = React.useState(true);
```

packages/dashboard/src/components/tasks/task-summary.tsx
```diff
@@ -0,0 +1,209 @@
...
+  return (
+    <Dialog
+      PaperProps={{
+        style: {
+          backgroundColor: setTaskDialogColor(taskState?.status),
+          boxShadow: 'none',
+        },
+      }}
+      open={show}
+      onClose={onClose}
```
aaronchongth, line 177:
```suggestion
      open={isOpen}
      onClose={() => {
        setIsOpen(false);
        onClose();
      }}
```

packages/dashboard/src/components/tasks/tasks-app.tsx
```diff
@@ -247,6 +253,13 @@ export const TasksApp = React.memo(
             </Grid>
           </Grid>
           <input type="file" style={{ display: 'none' }} ref={uploadFileInputRef} />
+          {openTaskSummary && (
+            <TaskSummary
+              onClose={() => setOpenTaskSummary(!openTaskSummary)}
+              task={selectedTask}
+              show={openTaskSummary}
```
aaronchongth, line 260:
```suggestion
              onClose={() => setOpenTaskSummary(false)}
              task={selectedTask}
```

packages/dashboard/src/components/tasks/task-inspector.tsx
```diff
@@ -0,0 +1,175 @@
...
+import { TaskInfo } from 'react-components';
+import { UserProfileContext } from 'rmf-auth';
+import { Enforcer } from '../permissions';
+import { TaskLogs } from './task-logs';
+
+export interface TableDataGridState {
+  task: TaskState | null;
+  open: boolean;
+  onClose: () => void;
+}
```
aaronchongth, line 26:
```suggestion
export interface TableDataGridState {
  task: TaskState | null;
  onClose: () => void;
}
```

packages/dashboard/src/components/tasks/task-inspector.tsx
```diff
@@ -0,0 +1,175 @@
...
+
+export function TaskInspector({ task, open, onClose }: TableDataGridState): JSX.Element {
+  const theme = useTheme();
+  const rmf = React.useContext(RmfAppContext);
+  const appController = React.useContext(AppControllerContext);
+
+  const [taskState, setTaskState] = React.useState<TaskState | null>(null);
+  const [taskLogs, setTaskLogs] = React.useState<TaskEventLog | null>(null);
+
```
aaronchongth, line 35:
```suggestion
  const [taskLogs, setTaskLogs] = React.useState<TaskEventLog | null>(null);
  const [isOpen, setIsOpen] = React.useState(true);

```

packages/dashboard/src/components/tasks/task-inspector.tsx
```diff
@@ -0,0 +1,175 @@
...
+      appController.showAlert('error', `Failed to cancel task: ${(e as Error).message}`);
+    }
+  }, [appController, taskState, rmf]);
+
+  return (
+    <>
+      <Grid container wrap="nowrap" direction="column" height="100%">
+        <Dialog
+          open={open}
+          onClose={onClose}
```
aaronchongth, line 109:
```suggestion
          open={isOpen}
          onClose={() => {
            setIsOpen(false);
            onClose();
          }}
```

packages/dashboard/src/components/tasks/task-summary.tsx
```diff
@@ -0,0 +1,209 @@
...
+          autoFocus
+        >
+          Inspect
+        </Button>
+      </DialogActions>
+      {openTaskDetailsLogs && (
+        <TaskInspector
+          task={taskState}
+          open={openTaskDetailsLogs}
+          onClose={() => setOpenTaskDetailsLogs(!openTaskDetailsLogs)}
```
aaronchongth, line 204:
```suggestion
          task={taskState}
          onClose={() => setOpenTaskDetailsLogs(false)}
```

aaronchongth (commented):
As mentioned, this set of suggestions will remove the boolean prop and handle opening and closing.

packages/dashboard/src/components/tasks/task-summay.tsx
```diff
@@ -0,0 +1,182 @@
...
+}
+
+export const TaskSummary = React.memo((props: TaskSummaryProps) => {
+  const classes = useStyles();
+
+  const { stopShowing, show, task } = props;
+
+  const [openTaskDetailsLogs, setOpenTaskDetailsLogs] = React.useState(false);
+
+  const getTaskMessage = (task: TaskState | null) => {
```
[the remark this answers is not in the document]
aaronchongth: True, I agree. Since we don't expect that much information from `Detail`. Let's pretty-print it with 2 spaces then

packages/dashboard/src/components/tasks/task-summary.tsx
```diff
@@ -0,0 +1,209 @@
...
+        ? Object.values(task.phases)[task.active - 1].category
+        : ''
+    }`;
+  };
+
+  const returnDialogContent = () => {
+    const contents = [
+      {
+        title: 'ID',
+        value: taskState ? taskState.booking.id : '',
```
aaronchongth, line 138:
```suggestion
        value: taskState ? taskState.booking.id : 'Invalid task state.',
```

packages/dashboard/src/components/tasks/task-summary.tsx
```diff
@@ -0,0 +1,209 @@
...
+      },
+      {
+        title: 'Current phase',
+        value: getTaskPhaseDetails(taskState),
```
aaronchongth, line 142:
```suggestion
        value: taskState ? getTaskPhaseDetails(taskState) : 'Invalid task state.',
```
We can probably follow your lead a few lines above and clean up the logic using a ternary

packages/dashboard/src/components/tasks/task-summary.tsx
```diff
@@ -0,0 +1,209 @@
...
+    if (!rmf || !task) {
+      return;
+    }
+    const sub = rmf
+      .getTaskStateObs(task.booking.id)
+      .subscribe((subscribedTask) => setTaskState(subscribedTask));
+    return () => sub.unsubscribe();
+  }, [rmf, task]);
+
+  const getTaskPhaseDetails = (task: TaskState | null) => {
```
aaronchongth, line 112:
```suggestion
  const getTaskPhaseDetails = (task: TaskState) => {
```
We can use non-null task states here now! :partying_face: Our error messages here can include the ID's safely now

packages/dashboard/src/components/tasks/task-summary.tsx
```diff
@@ -0,0 +1,209 @@
...
+    if (!task || !task.phases || !task.active) {
+      return 'Failed to retrieve current task phase';
+    }
+
+    if (!Object.values(task.phases)[task.active - 1]) {
+      return 'Failed to retrieve current task phase';
+    }
+
+    const message = Object.values(task.phases)[task.active - 1].detail;
```
aaronchongth, line 121:
```suggestion
    const message = Object.values(task.phases)[task.active - 1]?.detail;
```
We can further clean up the logic here

packages/dashboard/src/components/tasks/task-summary.tsx
```diff
@@ -0,0 +1,209 @@
...
+
+    if (message) {
+      return message;
+    }
+
+    return `Failed to retrieve current task phase details of id ${task.booking.id} and category ${
+      Object.values(task.phases)[task.active - 1].category
+        ? Object.values(task.phases)[task.active - 1].category
+        : ''
+    }`;
```
aaronchongth, line 131:
```suggestion
    const categoryString = Object.values(task.phases)[task.active - 1]?.category
      ? ` category ${Object.values(task.phases)[task.active - 1].category}`
      : '';
    return `Failed to retrieve current task phase details of id ${task.booking.id}${categoryString}`
```
We can really go crazy with ternaries here

packages/dashboard/src/components/tasks/task-summay.tsx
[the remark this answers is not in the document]
aaronchongth: Just left some further comments! Resolving this one, thanks!

packages/dashboard/src/components/tasks/task-inspector.tsx
```diff
@@ -0,0 +1,175 @@
...
+    }
+    const sub = rmf
+      .getTaskStateObs(task.booking.id)
+      .subscribe((subscribedTask) => setTaskState(subscribedTask));
+    return () => sub.unsubscribe();
+  }, [rmf, task]);
+
+  React.useEffect(() => {
+    AppEvents.taskSelect.next(task);
+  }, [task]);
```
aaronchongth, line 74: I don't think we need these lines anymore

aaronchongth (changes requested):
Just a few more nitpicks! But all is looking freaking amazing now, thanks for iterating on this!

packages/dashboard/src/components/tasks/task-inspector.tsx
aaronchongth: your changes were exactly what I was considering, thanks for implementing it so quickly! Resolved! :partying_face:

packages/dashboard/src/components/tasks/task-summay.tsx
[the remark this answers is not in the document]
Angatupyry: I tried it doing this `message.toString().replace(/\s+/g, '   ');` but the result wasn't significant.  
Perhaps we can make the TextField higher with `maxRows=2`

packages/dashboard/src/components/tasks/task-inspector.tsx
Angatupyry: You're right!

Angatupyry:
> As mentioned, this set of suggestions will remove the boolean prop and handle opening and closing.

Oh, man! I misunderstood. I thought you meant removing even the onClose props. 
Thanks for taking the time to write this suggestion. This is definitely another alternative.

aaronchongth (approved):
Thanks for the hard work! This looks so good now! LGTM

PR #706, opened 2023-04-27
Show up robot summary when user click robot row

Angatupyry:
## What's new
Popup dialog to show robot summary

- Removed micro app (Robot info) from the top right of the robot section

- Added a robot summary with the most relevant data

- Robot summary popup has a button to show  Task Inspector

- The map is shown on the right

[chrome-capture-2023-3-27.webm](https://user-images.githubusercontent.com/37310205/234994779-71149660-9c90-4277-8da5-fede07776b4e.webm)

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/robots/robot-summary.tsx
+++ b/packages/dashboard/src/components/robots/robot-summary.tsx
@@ -230,7 +230,8 @@ export const RobotSummary = React.memo(({ onClose, robot }: RobotSummaryProps) =
       maxWidth="sm"
     >
       <Grid container mb={1} alignItems="center" spacing={1}>
-        <Grid item xs={10}>
+        <Grid item xs={2}></Grid>
+        <Grid item xs={8}>
           <DialogTitle align="center">Robot Summary: {robotState?.name}</DialogTitle>
         </Grid>
         <Grid item xs={2}>
--- a/packages/dashboard/src/components/robots/robot-summary.tsx
+++ b/packages/dashboard/src/components/robots/robot-summary.tsx
@@ -172,11 +172,11 @@ export const RobotSummary = React.memo(({ onClose, robot }: RobotSummaryProps) =
   const returnDialogContent = () => {
     const contents = [
       {
-        title: 'Assigned Tasks',
+        title: 'Assigned tasks',
         value: taskState ? taskState.booking.id : 'No task',
       },
       {
-        title: 'Est. End Time',
+        title: 'Est. end time',
         value: taskState?.unix_millis_finish_time
           ? `${new Date(taskState?.unix_millis_finish_time).toLocaleString()}`
           : '-',
@@ -237,7 +237,7 @@ export const RobotSummary = React.memo(({ onClose, robot }: RobotSummaryProps) =
       <Grid container mb={1} alignItems="center" spacing={1}>
         <Grid item xs={2}></Grid>
         <Grid item xs={8}>
-          <DialogTitle align="center">Robot Summary: {robotState?.name}</DialogTitle>
+          <DialogTitle align="center">Robot summary: {robotState?.name}</DialogTitle>
         </Grid>
         <Grid item xs={2}>
           <Grid container justifyContent="flex-end">
@@ -253,10 +253,10 @@ export const RobotSummary = React.memo(({ onClose, robot }: RobotSummaryProps) =
       <Divider />
       {taskProgress && (
         <>
-          <Typography variant="h6" fontWeight="bold" align="center">
-            Task Progress
+          <Typography variant="body2" fontWeight="bold" ml={3} mt={1}>
+            Task progress
           </Typography>
-          <Box sx={{ width: '90%', ml: 3 }}>
+          <Box sx={{ width: '95%', ml: 3 }}>
             <LinearProgressWithLabel value={taskProgress * 100} />
           </Box>
         </>
--- a/packages/dashboard/src/components/robots/robot-summary.tsx
+++ b/packages/dashboard/src/components/robots/robot-summary.tsx
@@ -20,10 +20,18 @@ import { RobotTableData, base } from 'react-components';
 import { RobotState, Status2, TaskState } from 'api-client';
 import { EMPTY, combineLatest, mergeMap, of } from 'rxjs';
 import { TaskInspector } from '../tasks/task-inspector';
-import BatteryFullIcon from '@mui/icons-material/BatteryFull';
-import Battery60Icon from '@mui/icons-material/Battery60';
-import Battery20Icon from '@mui/icons-material/Battery20';
-import BatteryChargingFullIcon from '@mui/icons-material/BatteryChargingFull';
+import {
+  Battery0Bar,
+  Battery1Bar,
+  Battery2Bar,
+  Battery3Bar,
+  Battery4Bar,
+  Battery5Bar,
+  Battery6Bar,
+  BatteryFull,
+  BatteryChargingFull,
+  BatteryUnknown,
+} from '@mui/icons-material';
 
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
@@ -75,24 +83,28 @@ interface RobotSummaryProps {
 
 const showBatteryIcon = (robot: RobotState, robotBattery: number) => {
   if (robot.status === Status2.Charging) {
-    return <BatteryChargingFullIcon />;
+    return <BatteryChargingFull />;
   }
 
-  const batteryIcons: Record<string, JSX.Element> = {
-    '0': <Battery20Icon />,
-    '40': <Battery20Icon />,
-    '80': <Battery60Icon />,
-    '100': <BatteryFullIcon />,
+  const batteryIcons: Record<number, JSX.Element> = {
+    0: <Battery0Bar />,
+    16: <Battery1Bar />,
+    32: <Battery2Bar />,
+    48: <Battery3Bar />,
+    64: <Battery4Bar />,
+    80: <Battery5Bar />,
+    96: <Battery6Bar />,
+    100: <BatteryFull />,
   };
 
-  for (let i = 0; i < Object.keys(batteryIcons).length; i++) {
-    const level = Object.keys(batteryIcons)[i];
-    if (robotBattery <= parseInt(level)) {
+  for (const level in batteryIcons) {
+    if (robotBattery >= parseInt(level)) {
+      continue;
+    } else {
       return batteryIcons[level];
     }
   }
-
-  return null;
+  return <BatteryUnknown />;
 };
 
 export const RobotSummary = React.memo(({ onClose, robot }: RobotSummaryProps) => {
--- a/packages/dashboard/src/components/robots/robot-summary.tsx
+++ b/packages/dashboard/src/components/robots/robot-summary.tsx
@@ -1,0 +1,199 @@
+import {
+  Box,
+  Button,
+  Dialog,
+  DialogActions,
+  DialogContent,
+  DialogTitle,
+  Divider,
+  LinearProgress,
+  LinearProgressProps,
+  TextField,
+  Theme,
+  Typography,
+} from '@mui/material';
+import { makeStyles, createStyles } from '@mui/styles';
+import React from 'react';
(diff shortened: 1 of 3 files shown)


packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,253 @@
...
+        value: taskState?.unix_millis_finish_time
+          ? `${new Date(taskState?.unix_millis_finish_time).toLocaleString()}`
+          : '-',
+      },
+    ];
+
+    if (taskState) {
+      contents.push(
+        {
+          title: 'Location',
```
aaronchongth, line 164: Since we are not inferring exact location from here, I'd say we should probably name it something like `Navigation start`

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,253 @@
...
+          value: location,
+        },
+        {
+          title: 'Destination',
```
aaronchongth, line 168: Probably `Navigation destination` then

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,253 @@
...
+        taskState.estimate_millis /
+          (taskState.unix_millis_finish_time - taskState.unix_millis_start_time),
+      1,
+    );
+  }, [taskState]);
+
+  React.useEffect(() => {
+    if (!taskState || !taskState.phases || !taskState.active) {
+      setLocation('Failed to retrieve current location');
+      setDestination('Failed to retrieve robot destination');
```
aaronchongth, line 126: In the event that the robot is not performing any tasks, or the tasks don't involve navigation, I think it would be more appropriate to just set these 2 fields to `-`

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,253 @@
...
+      let match;
+      const waypoints = [];
+
+      // Iterate over all matches found by the regular expression
+      while ((match = regex.exec(message.toString()))) {
+        waypoints.push(match[1]);
+      }
+      setLocation(waypoints[0]);
+      setDestination(waypoints[1]);
+    }
```
aaronchongth, line 144: I think we are missing the else case here, where we really fail to retrieve the navigation start and destination. In the else case we should probably still display these fields as `-`, but use console errors to notify that we are unable to retrieve the information needed.

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,253 @@
...
+  }, [taskState]);
+
+  const returnDialogContent = () => {
+    const contents = [
+      {
+        title: 'Assigned Tasks',
```
aaronchongth, line 150:
```suggestion
        title: 'Assigned tasks',
```
Since field titles can sometimes be a few words long, it'll make it easier to read if we stick to capitalizing only the first letter, instead of all the first letters of each word

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,253 @@
...
+        value: taskState ? taskState.booking.id : 'No task',
+      },
+      {
+        title: 'Est. End Time',
```
aaronchongth, line 154:
```suggestion
        title: 'Est. end time',
```

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,253 @@
...
+      onClose={() => {
+        setIsOpen(false);
+        onClose();
+      }}
+      fullWidth
+      maxWidth="sm"
+    >
+      <Grid container mb={1} alignItems="center" spacing={1}>
+        <Grid item xs={10}>
+          <DialogTitle align="center">Robot Summary: {robotState?.name}</DialogTitle>
```
aaronchongth, line 214:
```suggestion
          <DialogTitle align="center">Robot summary: {robotState?.name}</DialogTitle>
```

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,253 @@
...
+            }%`}</Typography>
+            <Battery80Icon />
+          </Grid>
+        </Grid>
+      </Grid>
+      <Divider />
+      {taskProgress && (
+        <>
+          <Typography variant="h6" fontWeight="bold" align="center">
+            Task Progress
```
aaronchongth, line 229:
```suggestion
            Task progress
```

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,253 @@
...
+          </Typography>
+          <Box sx={{ width: '90%', ml: 3 }}>
```
aaronchongth, line 231:
```suggestion
          <Box sx={{ width: '95%', ml: 3 }}>
```
Thanks for narrowing this, never realized how weird it looked. Just tweaked it and found that 95% looked more natural. I'll leave it up to you to decide.

image

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,253 @@
...
+              robotState?.battery ? robotState?.battery * 100 : 0
+            }%`}</Typography>
+            <Battery80Icon />
+          </Grid>
+        </Grid>
+      </Grid>
+      <Divider />
+      {taskProgress && (
+        <>
+          <Typography variant="h6" fontWeight="bold" align="center">
```
aaronchongth, line 228:
```suggestion
          <Typography variant="body2" fontWeight="bold" ml={3} mt={1}>
```
more for aesthetics, I think centering `Task progress` looks a little odd with the dialog title already centered. I would suggest to reduce its size and shift it to the left, similar to how the other fields' titles are placed


image

aaronchongth (changes requested):
This looks super good! Thanks for tackling this, I think we've really cleaned up the user interactions and reduced a lot of wasted space on the dashboard :sparkles: I've left some comments below

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,253 @@
...
+      open={isOpen}
+      onClose={() => {
+        setIsOpen(false);
+        onClose();
+      }}
+      fullWidth
+      maxWidth="sm"
+    >
+      <Grid container mb={1} alignItems="center" spacing={1}>
+        <Grid item xs={10}>
```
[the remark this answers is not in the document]
Angatupyry: Yea! b253453cdd6fa63aedbd2c34f0815785e7ecc56a

packages/dashboard/src/components/robots/robot-summary.tsx
Angatupyry: Done! 5c7929f75e267afa7e53069cde2af45337555aea

packages/dashboard/src/components/robots/robot-summary.tsx
Angatupyry: Thanks for the suggestion!

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,281 @@
...
+export const RobotSummary = React.memo(({ onClose, robot }: RobotSummaryProps) => {
+  const classes = useStyles();
+  const rmf = React.useContext(RmfAppContext);
+
+  const [isOpen, setIsOpen] = React.useState(true);
+  const [robotState, setRobotState] = React.useState<RobotState | null>(null);
+  const [taskState, setTaskState] = React.useState<TaskState | null>(null);
+  const [openTaskDetailsLogs, setOpenTaskDetailsLogs] = React.useState(false);
+  const [navigationStart, setNavigationStart] = React.useState('-');
+  const [navigationDestination, setNavigationDestination] = React.useState('-');
```
aaronchongth, line 102: Sorry for pingpong-ing on this, I just realized it would be difficult to maintain if the default values `-` are set in multiple locations, could we use a `string | null` for these states, and set it to null by default, as well as when we fail to retrieve the navigation starts and destinations.

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,281 @@
...
+      1.0 -
+        taskState.estimate_millis /
+          (taskState.unix_millis_finish_time - taskState.unix_millis_start_time),
+      1,
+    );
+  }, [taskState]);
+
+  React.useEffect(() => {
+    if (!taskState || !taskState.phases || !taskState.active) {
+      setNavigationStart('-');
```
aaronchongth, line 145:
```suggestion
      setNavigationStart(null);
```
as mentioned above, something like this.

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,281 @@
...
+          ? `${new Date(taskState?.unix_millis_finish_time).toLocaleString()}`
+          : '-',
+      },
+    ];
+
+    if (taskState) {
+      contents.push(
+        {
+          title: 'Navigation start',
+          value: navigationStart,
```
aaronchongth, line 190: Only here will we then check if it is null, if it is, we give it `-`, probably using a ternary

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,281 @@
...
+      while ((match = regex.exec(message.toString()))) {
+        waypoints.push(match[1]);
+      }
+
+      setNavigationStart(waypoints[0]);
+      setNavigationDestination(waypoints[1]);
+    } else {
+      setNavigationStart('-');
+      setNavigationDestination('-');
+      console.log('Failed to retrieve robot navigation');
```
aaronchongth, line 168:
```suggestion
      console.error('Failed to retrieve robot\'s current navigation start and destination.');
```

aaronchongth (changes requested):
Last few things, thanks for iterating so far. Otherwise LGTM!

packages/dashboard/src/components/robots/robot-summary.tsx
Angatupyry: Nice!

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  const batteryIcons: Record<string, JSX.Element> = {
+    '100': <BatteryFullIcon />,
+    '80': <Battery60Icon />,
+    '40': <Battery20Icon />,
+    '0': <Battery20Icon />,
+  };
+
+  const key = Object.keys(batteryIcons).find((level) => robotBattery <= parseInt(level));
```
aaronchongth, line 88: I just realized that this might not achieve what we want, since the robot battery being 40% will also fall under `<=(parseInt('100'))`. I believe it is working at the moment coincidentally as keys is sorted alphabetically, and we are at the mercy of the hash function.

Would it be beneficial to loop over the record instead, picking the lowest possible case? In the previous example, choosing 40 instead of 100, even though battery is <= to both

packages/dashboard/src/components/robots/robot-summary.tsx
```diff
@@ -0,0 +1,281 @@
...
+    >
+      <Grid container mb={1} alignItems="center" spacing={1}>
+        <Grid item xs={2}></Grid>
+        <Grid item xs={8}>
+          <DialogTitle align="center">Robot summary: {robotState?.name}</DialogTitle>
+        </Grid>
+        <Grid item xs={2}>
+          <Grid container justifyContent="flex-end">
+            <Typography variant="subtitle1">{`${
+              robotState?.battery ? robotState?.battery * 100 : 0
```
aaronchongth, line 245: We will need to clip decimal points, 
image

Angatupyry:
> @Angatupyry I made some suggestions via #707, let me know what you think

Thanks!!! 
Done!

aaronchongth (approved):
LGTM, thanks for all the heavy lifting!

PR #707, opened 2023-05-05
Using all battery bar icons with unknown and empty

aaronchongth:
## What's new

I just noticed that the series of battery icons that we were using didn't cover empty battery levels :facepalm: I played around with the icons and tried to use integers instead of strings as keys. 

image
 
I had concerns that we are assuming the keys will always be sorted from low to high numbers, but I couldn't find a better solution. I think we can stick to the concept you had

## Self-checks


Angatupyry:
Thank you very much for looking into this matter and proposing your solution. 
It should always be ordered from lowest to highest although I agree that assuming that doesn't sound 100% safe but I didn't find a better solution either.

PR #708, opened 2023-05-08
System overview tab

Angatupyry:
## What's new

- Lift microapp removed from the map tab and added in the robot tab.
- Doors microapp has been removed from the map tab and displayed in the robot tab.
- The robot tab is now called System overview.
- Lift and Door microapp have been changed to tables.

[lift-2023-4-8.webm](https://user-images.githubusercontent.com/37310205/236914865-94c06305-5b78-47e2-b323-3f1ad2ffce2f.webm)
[door-state-2023-4-8.webm](https://user-images.githubusercontent.com/37310205/236914878-922af33d-c27d-4876-ab36-1c507fbb716c.webm)

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/appbar.tsx
+++ b/packages/dashboard/src/components/appbar.tsx
@@ -272,9 +272,9 @@ export const AppBar = React.memo(({ extraToolbarItems }: AppBarProps): React.Rea
         <LogoButton src={brandingIconPath} alt="logo" sx={{ width: logoSize }} />
         <NavigationBar value={tabValue}>
           <AppBarTab
-            label="Infrastructure"
+            label="Map"
             value="infrastructure"
-            aria-label="Infrastructure"
+            aria-label="Map"
             onTabClick={() => history.push(DashboardRoute)}
           />
           <AppBarTab
--- a/packages/dashboard/src/components/appbar.tsx
+++ b/packages/dashboard/src/components/appbar.tsx
@@ -278,9 +278,9 @@ export const AppBar = React.memo(({ extraToolbarItems }: AppBarProps): React.Rea
             onTabClick={() => history.push(DashboardRoute)}
           />
           <AppBarTab
-            label="Robots"
+            label="System Overview"
             value="robots"
-            aria-label="Robots"
+            aria-label="System Overview"
             onTabClick={() => history.push(RobotsRoute)}
           />
           <AppBarTab
--- a/packages/dashboard/src/components/doors-app.tsx
+++ b/packages/dashboard/src/components/doors-app.tsx
@@ -66,6 +66,7 @@ const DoorCard = ({ children, ...otherProps }: DoorCardProps) => {
       <TableCell>{doorTypeToString(otherProps.type)}</TableCell>
       <TableCell sx={labelStyle}>
         <Typography
+          data-testid="door-state"
           component="p"
           sx={{
             fontWeight: 'bold',
@@ -79,6 +80,7 @@ const DoorCard = ({ children, ...otherProps }: DoorCardProps) => {
         <Button
           variant="contained"
           size="small"
+          aria-label="open"
           sx={{ marginRight: 2 }}
           onClick={() =>
             rmf?.doorsApi.postDoorRequestDoorsDoorNameRequestPost(otherProps.name, {
@@ -91,6 +93,7 @@ const DoorCard = ({ children, ...otherProps }: DoorCardProps) => {
         <Button
           variant="contained"
           size="small"
+          aria-label="close"
           onClick={() =>
             rmf?.doorsApi.postDoorRequestDoorsDoorNameRequestPost(otherProps.name, {
               mode: RmfDoorMode.MODE_CLOSED,
--- a/packages/dashboard/src/components/doors-app.tsx
+++ b/packages/dashboard/src/components/doors-app.tsx
@@ -1,20 +1,27 @@
-import { CardActions, Grid } from '@mui/material';
+import { Button, SxProps, useTheme } from '@mui/material';
 import { BuildingMap, DoorState } from 'api-client';
 import React from 'react';
 import {
-  DoorCard as BaseDoorCard,
   DoorCardProps as BaseDoorCardProps,
-  DoorControls,
+  doorModeToString,
+  doorTypeToString,
 } from 'react-components';
-import { DoorMode as RmfDoorMode } from 'rmf-models';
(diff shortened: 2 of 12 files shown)


packages/dashboard/src/components/lifts-app.tsx
```diff
@@ -25,62 +38,108 @@ const LiftCard = ({ children, ...otherProps }: LiftCardProps) => {
...
+          <TableCell>Current floor</TableCell>
+          <TableCell>Destination floor</TableCell>
+          <TableCell>Door state</TableCell>
+          <TableCell></TableCell>
+        </TableRow>
+      </TableHead>
+      <TableBody>
+        {buildingMap &&
+          buildingMap.lifts.map((lift) => {
+            return <LiftCard key={lift.name} name={lift.name} lift={lift} sx={{ width: 200 }} />;
```
aaronchongth, line 140: Ah I see what's happening here. Since we are no longer using the `LiftCard` from `react-components`, and just using `LiftCardProp` with omitted fields, I would suggest we set up something new with the proper names, like `LiftTableRow` and `LiftTableRowProp`, and we should be able to stop using `LiftCard` altogether.

We can keep the implementation of `LiftTableRow` in `lifts-app` since it is tied to the implementation of the app which is a table.

packages/react-components/lib/lifts/lift-card.tsx
```diff
@@ -13,8 +13,10 @@ import {
 import React from 'react';
 import { LiftState } from 'rmf-models';
 import { doorStateToString, motionStateToString } from './lift-utils';
+import { Lift } from 'api-client';
 
 export interface LiftCardProps extends CardProps {
+  lift: Lift;
```
aaronchongth, line 19: As per my comment on top, once we have `LiftTableRow` internally in `lifts-app`, we won't need to modify or use these props anymore

packages/dashboard/src/components/doors-app.tsx
```diff
@@ -43,36 +120,31 @@ export const DoorsApp = createMicroApp('Doors', () => {
...
+          <TableCell>Type</TableCell>
+          <TableCell>Door State</TableCell>
+          <TableCell></TableCell>
+        </TableRow>
+      </TableHead>
+      <TableBody>
+        {buildingMap &&
+          buildingMap.levels.flatMap((level) =>
+            level.doors.map((door) => (
+              <DoorCard
```
aaronchongth, line 137: Similar to my comment on lift cards, we will stop using `DoorCard` and we should move to something new implemented here in `doors-app`, like `DoorTableRow` and `DoorTableProp`, and we should be able to remove that dependency now.

packages/dashboard/src/components/appbar.tsx
```diff
@@ -278,9 +278,9 @@ export const AppBar = React.memo(({ extraToolbarItems }: AppBarProps): React.Rea
             onTabClick={() => history.push(DashboardRoute)}
           />
           <AppBarTab
-            label="Robots"
+            label="System Overview"
             value="robots"
-            aria-label="Robots"
+            aria-label="System Overview"
```
aaronchongth, line 283: We should change the name of the `Infrastructure` tab to `Map` as well

aaronchongth (changes requested):
Thanks for putting this together! I've some comments for the implementation, and I'll leave a small comment regarding the UI here,

Can we have a slightly larger spacing between OPEN/CLOSED and the arrows?
Screenshot from 2023-05-15 19-33-32

Angatupyry:
> Can we have a slightly larger spacing between OPEN/CLOSED and the arrows?

Screenshot from 2023-05-17 13-39-53
What about now?

packages/dashboard/src/components/doors-app.tsx
```diff
@@ -1,32 +1,109 @@
...
-    <BaseDoorCard mode={doorState ? doorState.current_mode.value : -1} {...otherProps}>
-      {children}
-    </BaseDoorCard>
+    <TableRow key={doorName}>
+      <TableCell>{doorName}</TableCell>
+      <TableCell>{levelName}</TableCell>
+      <TableCell>{doorTypeToString(doorType)}</TableCell>
+      <TableCell sx={labelStyle}>
+        <Typography
+          data-testid="door-state"
```
aaronchongth, line 69: nit: let's use this opportunity to sanitize double quotes from our code base, only use single-quotes per typescript guidelines

packages/dashboard/src/components/lifts-app.tsx
```diff
@@ -1,86 +1,139 @@
...
-      doorState={liftState?.door_state}
-      currentFloor={liftState?.current_floor}
-      destinationFloor={liftState?.destination_floor}
-      {...otherProps}
-    >
-      {children}
-    </BaseLiftCard>
+    <TableRow key={lift.name}>
+      <TableCell>{lift.name}</TableCell>
+      <TableCell>{liftState?.current_floor || '?'}</TableCell>
```
aaronchongth, line 70:
```suggestion
      <TableCell>{liftState?.current_floor || 'N/A'}</TableCell>
```

packages/dashboard/src/components/lifts-app.tsx
```diff
@@ -1,86 +1,139 @@
...
+      <TableCell>{liftState?.destination_floor || 'Unknown'}</TableCell>
```
aaronchongth, line 71:
```suggestion
      <TableCell>{liftState?.destination_floor || 'N/A'}</TableCell>
```

aaronchongth (changes requested):
Thanks for making those changes! I have a few smaller nitpicks that I didn't catch last time, sorry about that. Since we are making lots of changes nowadays, I'd like to use this opportunity to handle linting more, like double-quotes to single-quotes, so if you see any of it in this PR, feel free to make the changes.

I was planning to purge the usage of `unknown` from the dashboard too, I believe `N/A` is more appropriate, since the information is not available, instead of not parseable

packages/dashboard/src/components/doors-app.tsx
Angatupyry: I left some comments on google chat about our eslint configuration currently. 
Maybe we can delve into this in a specific PR or it may be the possibility to define conventions. Let me know what you think.

aaronchongth: sure that sounds like a good idea, it'll be more a maintenance task and should not be prioritized.

aaronchongth (approved):
Thanks for getting this large shuffling done! LGTM!

PR #709, opened 2023-05-19
Show Robot summary after clic robot icon

Angatupyry:
## What's new
- Click map icon for robot state overlay

[chrome-capture-2023-4-19.webm](https://github.com/open-rmf/rmf-web/assets/37310205/d11245a7-2929-4aa8-b4af-bae3b95cb61e)

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/map-app.tsx
+++ b/packages/dashboard/src/components/map-app.tsx
@@ -16,6 +16,7 @@ import {
   LMap,
   loadAffineImage,
   Place,
+  RobotTableData,
   TrajectoryTimeControl,
 } from 'react-components';
 import { AttributionControl, ImageOverlay, LayersControl, Pane } from 'react-leaflet';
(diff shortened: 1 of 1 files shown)


Angatupyry:
@aaronchongth I found this commented code https://github.com/open-rmf/rmf-web/blob/main/packages/dashboard/src/components/map-app.tsx#LL181C1-L206C51
I don't really like commented code because it does not help the programmer understand why it is there. Unless there is a comment explaining why it is necessary to keep that code. 
I propose to delete it, if in the future we need to do something with that code we have it saved in the git history. 
But maybe you know better than me about this so let me know.

aaronchongth (approved):
LGTM!

Regarding the commented code, feel free to remove it. We have no intention to support viewing of negotiations at the moment, this can be handled in the future, but in the meantime it is definitely a good idea to keep the code clean.

PR #718, opened 2023-06-13
Op. Mode column in Lifts

Angatupyry:
## What's new

- Op. Mode column added

Screenshot from 2023-06-13 13-36-14

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/lifts-app.tsx
+++ b/packages/dashboard/src/components/lifts-app.tsx
@@ -3,6 +3,7 @@ import React from 'react';
 import { LiftTableData, LiftDataGridTable } from 'react-components';
 import { createMicroApp } from './micro-app';
 import { RmfAppContext } from './rmf-app';
+import { getApiErrorMessage } from './utils';
 
 export const LiftsApp = createMicroApp('Lifts', () => {
   const rmf = React.useContext(RmfAppContext);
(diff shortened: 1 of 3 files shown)


packages/dashboard/src/components/lifts-app.tsx
```diff
@@ -25,33 +26,40 @@ export const LiftsApp = createMicroApp('Lifts', () => {
...
+        const { health_status } = data;
+        const sub = rmf.getLiftStateObs(lift.name).subscribe((liftState) => {
+          setLiftTableData((prev) => {
+            return {
+              ...prev,
+              [lift.name]: [
+                {
+                  index: i,
+                  name: lift.name,
+                  opMode: health_status ? health_status : 'N/A',
```
aaronchongth, line 41: I think we'll need to remap the text, https://github.com/open-rmf/rmf-web/blob/main/packages/api-server/api_server/models/health.py#L10, from Healthy, Unhealthy, Dead to ONLINE, UNSTABLE, OFFLINE, with a similar styling with the door status (colored font, uppercase)

aaronchongth (changes requested):
Thanks for getting this out so quickly! I left one review comment which will apply to doors too

packages/dashboard/src/components/lifts-app.tsx
aaronchongth: image

Angatupyry: Nice!
Done here 7bf9c9673c38fd08fa38d14f1e5e81b2b178a5b0

aaronchongth (approved):
thanks! LGTM

PR #721, opened 2023-06-20
Feature/acknowledged by showed in task logs

Angatupyry:
## What's new

- Save `acknowledged by` in logs as a new value of phases
- Save `acknowledged by` in task state as a new value of phases
- Those are displayed in the last log with a warning color

[chrome-capture-2023-5-20.webm](https://github.com/open-rmf/rmf-web/assets/37310205/eca885f9-47b1-4f4b-905d-cd507f29f79b)

[chrome-capture-2023-5-20 (2).webm](https://github.com/open-rmf/rmf-web/assets/37310205/e070d68a-3889-4bda-95e5-0c06348370a4)

## Self-checks

### DIFF
--- a/packages/api-server/api_server/repositories/alerts.py
+++ b/packages/api-server/api_server/repositories/alerts.py
@@ -84,9 +84,12 @@ class AlertRepository:
         user = User(username="__rmf_internal__", is_admin=True)
         task_repo = TaskRepository(user)
         # Save in logs who was the user that acknowledged the task
-        await task_repo.save_log_acknowledged_task_completion(
-            alert.id, self.user.username, unix_millis_acknowledged_time
-        )
+        try:
+            await task_repo.save_log_acknowledged_task_completion(
+                alert.id, self.user.username, unix_millis_acknowledged_time
+            )
+        except Exception as e:
+            raise RuntimeError(f"Error in save_log_acknowledged_task_completion {e}")
 
         await alert.delete()
         ack_alert_pydantic = await ttm.AlertPydantic.from_tortoise_orm(ack_alert)
--- a/packages/api-server/api_server/repositories/alerts.py
+++ b/packages/api-server/api_server/repositories/alerts.py
@@ -11,8 +11,11 @@ from api_server.repositories.tasks import TaskRepository
 
 
 class AlertRepository:
-    def __init__(self, user: User):
+    def __init__(self, user: User, task_repo: TaskRepository = None):
         self.user = user
+        self.task_repo = (
+            task_repo if task_repo is not None else TaskRepository(self.user)
+        )
 
     async def get_all_alerts(self) -> List[ttm.AlertPydantic]:
         alerts = await ttm.Alert.all()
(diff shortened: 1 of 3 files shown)


packages/api-server/api_server/repositories/alerts.py
```diff
@@ -71,13 +72,23 @@ async def acknowledge_alert(self, alert_id: str) -> Optional[ttm.AlertPydantic]:
...
+                "unix_millis_acknowledged_time": unix_millis_acknowledged_time,
             }
         )
         await ack_alert.save()
+
+        user = User(username="__rmf_internal__", is_admin=True)
+        task_repo = TaskRepository(user)
+
+        # Save in logs who was the user that acknowledged the task
+        await task_repo.save_log_acknowledged_task_completion(
```
aaronchongth, line 88:
```suggestion
        await self.task_repo.save_log_acknowledged_task_completion(
```
with the suggestion in the main review comment, we can remove the construction of `task_repo` in the few lines before this, and just use `self.task_repo`

aaronchongth, line 88: Since we are raising exceptions in `save_log_acknowledged_task_completion`, perhaps we should add this within a `try`

aaronchongth (changes requested):
So far so good! It works as expected, just have a couple of comments.

Perhaps we should pass in a `TaskRepository` instance instead of creating one every time we acknowledge an alert. If it is not passed in, we can use the user of this alert repository to create a task repository too

```
class AlertRepository:
    def __init__(self, user: User, task_repo: TaskRepository = None):
        self.user = user
        self.task_repo = task_repo if task_repo is not None else TaskRepository(self.user)
```

packages/api-server/api_server/repositories/alerts.py
Angatupyry: Oh, man! Thanks so much for catching it. 
Unnecessary construction. 
Done 8b334f4cc27f7722a55acd08af8a57bee260ce03

Angatupyry: `Self` here is `AlertRepository` and I don't have access to `task_repo`. What I could do is `self.user` but I make the instance of `User` to make sure it is admin.

Angatupyry: Done a8bedf464550510432964db1749f71ac745d7aa6 !

aaronchongth: oh! `self.task_repo` would refer to the `TaskRepository` that is passed in during construction, take a look at the example code block I provided in the main review comment

```
class AlertRepository:
    def __init__(self, user: User, task_repo: TaskRepository = None):
        self.user = user
        self.task_repo = task_repo if task_repo is not None else TaskRepository(self.user)
```

aaronchongth: `task_repo` can be passed into `alert_repo` here, https://github.com/open-rmf/rmf-web/blob/feature/acknowledged_by-showed-in-task-logs/packages/api-server/api_server/routes/internal.py#L15

aaronchongth (changes requested):
Take a look at my follow up comments, let me know what you think!

packages/api-server/api_server/repositories/alerts.py
Angatupyry: Gotcha!
Sorry for my misunderstanding! 
Done here! c2cd6e7fb0cabd00c586ce3a9cdf820a56508c6a

Angatupyry:
@aaronchongth I also added [those](https://github.com/open-rmf/rmf-web/pull/721/files#diff-280f9b31b90f5884145ae93730c04f0a2abca8c007b7caed3f3e16e465f53f2bR102-R105) lines, please take a look

aaronchongth (approved):
Cool, thanks for handling the dependencies too! LGTM

PR #722, opened 2023-06-21
Changes in the robot table

Angatupyry:
## What's new

- Cell background color removed
- Text color depending on status was added. 
- Level column added
- Remove the decimal value from the battery column and add a percentage symbol at the end. 
- `Robot name` column changed by `Name`
- Display Robot name in the first column

Screenshot from 2023-06-21 14-29-48

## Self-checks


aaronchongth (approved):
Thanks! This is working very well!

PR #727, opened 2023-07-13
Migration to react 18

Angatupyry:
### React 18. 
- Some advantages: https://react.dev/blog/2022/03/29/react-v18

### What changed?
Many dependencies had to change due to incompatibility with React 18. 
The most relevant change was the routing handling. 
It was upgraded to react router 6. 

The most relevant change was to add `/*` to the paths that have children, in our case, only `/admin` in order to render the child paths. 

#### Error before adding the /* path 

```
You rendered descendant <Routes (or called `useRoutes()`) at "/" (under <Route path="/">) 
but the parent route path has no trailing "*". This means if you navigate deeper, 
the parent won't match anymore and therefore the child routes will never render.

Please change the parent <Route path="/"> to <Route path="*">.
```

You can test for this error by removing `/*` from line 17 of the file `/util/url.ts` of the dashboard package

The other changes are `react router v6` enhancements or different package names that serve the same purpose as the previous version. You can look at some of them [here](https://reactrouter.com/en/main/upgrading/v5#upgrade-to-react-router-v6
)

### Other dependencies changed were. 
- testing-library/dom
- testing-library/jest-dom
- testing-library/react
- testing-library/user-event

Many test files were changed due to incompatibility with react 18. 

#### Some of the most relevant changes were:

**Obs:** In react-component it was not possible to upgrade to the latest `testing library` version due to incompatibility with `webpack` and `karma`. I consider something not relevant to address now. Therefore:

- This code was added to avoid the alert 
```
Warning: An update to ComponentName inside a test was not wrapped in act(...).
ocasionada por incompatibilidad de webpack y testing library. 

```
```tsx
const originalError = console.error;
beforeAll(() => {
  // this is just a little hack to silence
  // Warning: An update to ComponentName inside a test was not wrapped in act(...).
  console.error = (...args) => {
    if (/Warning.*not wrapped in act/.test(args[0])) {
      return;
    }
    originalError.call(console, ...args);
  };
});
```

- Use fireEvent instead of userEvent in many cases because by default fireEvent is wrapped inside act function and this is useful when the user does some action and this action will cause component updates and re-render. (https://kentcdodds.com/blog/common-mistakes-with-react-testing-library)

- This is a relatively major change, so please test all possible cases, (create tasks, create alerts, create schedules). Especially the interactions of the routes in admin. 

## Self-checks


aaronchongth (commented):
Thanks for the enormous effort with the migration! While testing, I started changing a number of things and thought it would be best if I just branched off your work with my suggesions, https://github.com/open-rmf/rmf-web/pull/729. let me know what you think.

Other than the items I listed in https://github.com/open-rmf/rmf-web/pull/729, LGTM!

aaronchongth (approved):
LGTM!

PR #731, opened 2023-08-01
Feature/display adhooc or scheduled

Angatupyry:
## What's new

- Display on task queue whether task was scheduled or ad-hoc
- For now, automatic robot tasks do not have a distinct icon due we have not a way to differential it from real users since the name is the ROS 2 node name

chrome-capture-2023-6-31

## Self-checks

### DIFF
--- a/packages/react-components/lib/tasks/task-table-datagrid.tsx
+++ b/packages/react-components/lib/tasks/task-table-datagrid.tsx
@@ -11,9 +11,10 @@ import {
   GridFilterModel,
   GridSortModel,
 } from '@mui/x-data-grid';
-import { styled, TextField } from '@mui/material';
+import { styled, TextField, Stack, Typography } from '@mui/material';
 import * as React from 'react';
 import { TaskState, Status } from 'api-client';
+import { InsertInvitation as ScheduleIcon, Person as UserIcon } from '@mui/icons-material/';
 
 const classes = {
   taskActiveCell: 'MuiDataGrid-cell-active-cell',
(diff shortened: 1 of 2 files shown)


packages/react-components/lib/tasks/task-table-datagrid.tsx
```diff
@@ -134,11 +135,24 @@ export function TaskDataGridTable({
...
+        return (
+          <Stack direction="row" alignItems="center" gap={1}>
+            {cellValues.row.booking.requester &&
+            cellValues.row.booking.requester.includes('scheduled') ? (
+              <ScheduleIcon />
+            ) : (
+              <UserIcon />
+            )}
+            <Typography variant="body1">
+              {cellValues.row.booking.requester ? cellValues.row.booking.requester : 'unknown'}
```
aaronchongth, line 150: We can probably just display `stub` instead of `stub__scheduled` here, the different icon should be a good enough indicator

packages/react-components/lib/tasks/task-table-datagrid.tsx
```diff
@@ -134,11 +135,24 @@ export function TaskDataGridTable({
...
       editable: false,
-      valueGetter: (params: GridValueGetterParams) =>
-        params.row.booking.requester ? params.row.booking.requester : 'unknown',
-      flex: 1,
+      renderCell: (cellValues) => {
+        return (
+          <Stack direction="row" alignItems="center" gap={1}>
+            {cellValues.row.booking.requester &&
+            cellValues.row.booking.requester.includes('scheduled') ? (
+              <ScheduleIcon />
```
aaronchongth, line 145: Can we add a tooltip to these icons, such that hovering over them tells us what the icons mean. We can have the tooltips display `User scheduled`, as well as `User submitted` for the case of `UserIcon`.

aaronchongth (changes requested):
These look great! Thanks for adjusting the width of the table too. I left some minor comments on how it looks, but the implementation works great!

packages/react-components/lib/tasks/task-table-datagrid.tsx
Angatupyry: Sure!
Done 17b513d8de25bf91f8fe4136401a830095bdf5fe

aaronchongth (approved):
LGTM! Thanks!

PR #733, opened 2023-08-08
Feature/beacons frontend

aaronchongth:
## What's new

* Beacons micro-app
* Added missing declarations for the beacons subscription backend
* Updated api-client with beacon related routes

https://github.com/open-rmf/rmf-web/assets/5383623/4d718bd9-9d42-4905-8d3a-1861898fbe0c

# Testing

Either
* modify `dashboard/src/components/robots/robots-workspace.tsx` to show the beacons micro app
* or just use a custom tab to display the microapp

## Self-checks


packages/react-components/lib/beacons/beacon-table-datagrid.tsx
```diff
@@ -0,0 +1,132 @@
+import { ApiServerModelsTortoiseModelsBeaconsBeaconStateLeaf as BeaconState } from 'api-client';
+import { DataGrid, GridColDef, GridValueGetterParams, GridCellParams } from '@mui/x-data-grid';
+import { Box, SxProps, Typography, useTheme } from '@mui/material';
+import * as React from 'react';
```
Angatupyry, line 4: nit: Is there any reason to do this instead of how we do it in the rest of the files (import React from 'react';)?
In fact, we could start by doing import {useState, useEffect, ...} from React to be more specific.

Angatupyry (approved):
Thanks for doing this, Aaron! 
This code follows the way that we have designed Doors and Lifts tables so I have not any big comments. 
I left only one nit comment. 
All clean and looks really nice!

Angatupyry:
@aaronchongth Oh, btw, I don't see the Beacons table in the System Overview tab, is there a reason for that?
Screenshot from 2023-08-08 14-52-14

packages/react-components/lib/beacons/beacon-table-datagrid.tsx
aaronchongth: Looks like I copied over from doors and lifts too much, thanks for catching that! https://github.com/open-rmf/rmf-web/pull/733/commits/7cd343db68680249abbc18a01ca85faf46667d94 fixes it as well as for doors and lifts

aaronchongth:
> @aaronchongth Oh, btw, I don't see the Beacons table in the System Overview tab, is there a reason for that? Screenshot from 2023-08-08 14-52-14

@Angatupyry sorry I forgot to answer this, beacons will be rather specific hardware that is not used in any of our demos or use cases for now, so we'll only use them in projects or when we do have beacons in our simulation demos. For projects that use them, I expect the layout will be different anyway

PR #735, opened 2023-08-11
Deletes all existing building map entries if the id does not match the current served map, before saving

aaronchongth:
## What's new

Ensures that the map saved in db is always the one that is currently being served by the building map server.

This can be tested by starting an api-server instance, and changing the building map server,

```
# start api-server in a terminal
pnpm start

# serve office map, Ctrl+C, serve airport terminal map, repeat
ros2 run rmf_building_map_tools building_map_server WORKSPACE_DIR/install/rmf_demos_maps/share/rmf_demos_maps/office/office.building.yaml
ros2 run rmf_building_map_tools building_map_server WORKSPACE_DIR/install/rmf_demos_maps/share/rmf_demos_maps/airport_terminal/airport_terminal.building.yaml
```

## Self-checks

### DIFF
--- a/packages/api-server/api_server/models/building_map.py
+++ b/packages/api-server/api_server/models/building_map.py
@@ -21,7 +21,7 @@ class BuildingMap(rmf_building_map_msgs.BuildingMap):
 
     async def save(self) -> None:
         existing_maps = await ttm.BuildingMap.all()
-        for map in existing_maps:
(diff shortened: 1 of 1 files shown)


akash-roboticist (approved):
Since other data is not affected in the db, I think this looks fine and will minimize issues due to map updates during commissioning phase.
Thanks Aaron!

Issue #734, opened 2023-08-08
Migrate react leaflet to react three fiber.


Angatupyry:
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

React Three Fiber offers a powerful alternative for rendering maps. 
By transitioning from React Leaflet to React Three Fiber, we gain the ability to build immersive environments with multi-level scaling. This means that we can seamlessly navigate and visualize objects at various levels of detail, providing a more natural and comprehensive understanding of geographical spaces. 

### Done

- Show robots.
- Show places as cube.
- Show places' names as text.
- Show RobotInformation when clicking a robot. 
- Show Trajectory. 

### In progress

- [x] Which object would represent the robot? For now, it is this [object](https://app.gazebosim.org/OpenRobotics/fuel/models/Hatchback) 
    - Aaron >> robots, we can stick to what we have in the past, have a default drawn circle if no logo or 3d asset is provided through dashboard resources
- [x] Draw an elevator (like a square) 
- [x] Show robots' names 
- [X] Show dispensers 
     - [X]  Which object would represent the dispensers? For now, it is a cube
     - A circular shape is shown according to the comment below.
- [X] Show ingestor 
     - [X]  Which object would represent the ingestor? For now, it is a cube
     - A circular shape is shown according to the comment below.
- [x] Fix the position of places when the level changes. Currently, the place names and cubes (represents places) look good on the first level but when switching to level 2 they are one floor down. 

- [x] Scale of waypoints and objects. 
- [ ] Calculate the scale correctly. 
     - Currently, the scale for both objects and places is a hardcoded Vector3 and, although they look good at all levels, it is not good practice. 
- [ ] Correctly calculate the z-value of the trajectory. 
    - Currently, it has a hardcoded value (4) 

### Fix (Details in videos)
- [x] Zoom. 
- [x] Automatic zoom when changing levels. 
    - The elevation is set to a standard value

[chrome-capture-2023-7-8.webm](https://github.com/open-rmf/rmf-web/assets/37310205/76f4b7cc-44fb-4f43-84eb-c9767edea328)
	
[chrome-capture-2023-7-8 (1).webm](https://github.com/open-rmf/rmf-web/assets/37310205/5507b205-a50a-49f9-b1b1-39e091698f3a)

aaronchongth:
Per VC,
[...]

PR #737, opened 2023-08-17
Fix panning rotation issue

aaronchongth:
## What's new

Related to https://github.com/open-rmf/rmf-web/issues/734

* Use orthographic camera by default
* Target of orthographic camera set to 1km below origin, this fixes the re-orientation when we pan
* Changing left click to panning
* Disabling all other mouse interations (rotation and dolly)
* Using a fixed zoom of 20 for now, this should be configurable later on

## Self-checks


Angatupyry (approved):

Angatupyry:
This is awesome!!!
Thanks, @aaronchongth !

PR #745, opened 2023-08-24
Fixing missing handler delivery task description

aaronchongth:
## What's new

Fixes to broken delivery task request
* re-introduced the `handler` field for pickup and dropoff
* reverted back to `getPlace` function call, but changed the implementation to return information for all types of named places
* create-task now accepts a `Record<string, string>` for pickup and dropoff points, as it maps the name of the waypoint to the name of the handler
* checks for valid pickup and dropoff waypoint names and handler names before triggering change, to prevent crashes in the backend
* fixed a minor typo where modifying dropoff location changes pickup locations too

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/appbar.tsx
+++ b/packages/dashboard/src/components/appbar.tsx
@@ -35,7 +35,7 @@ import {
   AppBarTab,
   CreateTaskForm,
   CreateTaskFormProps,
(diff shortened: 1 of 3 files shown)


Angatupyry (approved):
LGTM!
Thanks, Aaron!

PR #753, opened 2023-08-28
App bar changes

Angatupyry:
## What's new

- Increase the font size on the title bar 
- Change New Task button size

## Self-checks


aaronchongth (changes requested):
Thanks for the help! It looks like the project's logo size is 406 x 124, which makes the appbar thicker than usual. For this deployment branch, I think we can make our default logo the same height, and maximize our font size based on that. Let me know if you think that's a good idea

edit: feel free to merge in from `deploy/hammer` too, to get https://github.com/open-rmf/rmf-web/pull/755 in, so the CI can run normally

aaronchongth (approved):
Thanks for resizing these items! I've added some changes based on your work for some other components too

PR #757, opened 2023-08-30
Feature/time warning

aaronchongth:
## What's new

Adds warn time to task creation

image

image

image

image

## How to test

* Run in office world
* select delivery task, with `Cart RFID` as `coke`, select pickup and dropoff
* check warn time, and select a warn time of before year 1970 (since we are using simulation time)
* submit
* a warning should immediately appear, that can be acknowledged or dismissed per normal
* acknowledging will append a log to the task log

## Self-checks


Angatupyry (approved):
I have looked at the code and it looks fine. Some "if's" and small things could be done differently but it wouldn't change the performance. It looks fine to me. Maybe we can think of it as another category when we are not running.

LGTM!
Thanks for this hard work, Aaron!!

PR #765, opened 2023-09-08
Feature/scheduling month day

aaronchongth:
## What's new

* Display month and day view
* Stop auto refresh when viewing schedule tab

## Self-checks

https://github.com/open-rmf/rmf-web/assets/5383623/9ab85577-6788-43d4-be4c-e11f79993363

### DIFF
--- a/packages/api-server/api_server/routes/tasks/scheduled_tasks.py
+++ b/packages/api-server/api_server/routes/tasks/scheduled_tasks.py
@@ -26,7 +26,12 @@ class PostScheduledTaskRequest(BaseModel):
 
 
 def datetime_to_date_format(date: datetime) -> str:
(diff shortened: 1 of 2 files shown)


Angatupyry (changes requested):
This is an excellent approach, Aaron!
Thanks for this effort!

I added a commit to fix the bug in the delete schedule when the month has two digits. 
Also, when you delete an event, it is not refreshing unless you switch tabs. 
Adding this
`setRefreshTaskAppCount((oldValue) => ++oldValue);`
 in the function of the event deletion may fix that (line 441)

Angatupyry (approved):
Excellent approach!
I tested it and it works correctly
LGTM!

PR #768, opened 2023-09-12
Feature/delivery alerts

aaronchongth:
## What's new

- [x]  `delivery_alerts` backend
- [x] frontend

## Assumptions
* When a delivery alert has tier `error`, there are no recover modes for that task, and this feature assumes it has been cancelled before emitting this alert.

## How it looks like

Pardon the long video, here it displays triggering the delivery alerts using the API server,
* warning on a task
* error on the same task, overwrites the warning (warning disappears)
* refreshing the dashboard still retrieves the current unresolved delivery alert
* closing an error delivery alert is considered resolving it (won't show up anymore)
* different action buttons to change the `action` field, before the alert is allowed to be closed

https://github.com/open-rmf/rmf-web/assets/5383623/228b9fbe-c9ce-466e-9daf-861e0a260417

## Self-checks


packages/dashboard/src/components/delivery-alert-store.tsx
```diff
@@ -0,0 +1,539 @@
...
+        // the delivery alerts that are currently present and have not been
+        // responded to.
+        if (!alert.task_id) {
+          filteredAlertsMap[alert.id] = { deliveryAlert: alert, taskState: undefined };
+          continue;
+        }
+
+        // Tasks that have not been encountered before will go into
+        // taskIdToAlertsMap first, as an error will supercede a warning.
+        if (!Object.keys(taskIdToAlertsMap).includes(alert.task_id)) {
```
Angatupyry, line 346: Since it is an if-else, the variable has a local scope, so I think we can use const and add it in the try. That makes the code readable because it assumes that the variable will be from the state of the api response. 

Also, I think the else has to follow the same structure and add the logic inside the try and not outside. 
```suggestion
if (!Object.keys(taskIdToAlertsMap).includes(alert.task_id)) {
  try {
    const state = (await rmf.tasksApi.getTaskStateTasksTaskIdStateGet(alert.task_id)).data;

    taskIdToAlertsMap[alert.task_id] = {
      deliveryAlert: alert,
      taskState: state,
    };
  } catch (error) {
    console.error(`Failed to fetch task state for ${alert.task_id} for delivery alert ${alert.id}`);
  }
} else if (alert.tier === 'error') {
  try {
    const state = (await rmf.tasksApi.getTaskStateTasksTaskIdStateGet(alert.task_id)).data;

    taskIdToAlertsMap[alert.task_id] = {
      deliveryAlert: alert,
      taskState: state,
    };
  } catch (error) {
    console.error(`Failed to fetch task state for ${alert.task_id} for delivery alert ${alert.id}`);
  }
}
```

Angatupyry: Besides, since we are doing 
```
taskIdToAlertsMap[alert.task_id] = {
            deliveryAlert: alert,
            taskState: state,
          };

```
Maybe in the if statement we can do something like

`if (!taskIdToAlertsMap[alert.task_id]) `

packages/dashboard/src/components/delivery-alert-store.tsx
```diff
@@ -0,0 +1,539 @@
...
+          continue;
+        }
+        const alertData = taskIdToAlertsMap[taskId];
+        filteredAlertsMap[alertData.deliveryAlert.id] = alertData;
+      }
+      setAlerts(filteredAlertsMap);
+    })();
+
+    const sub = rmf.deliveryAlertObsStore.subscribe(async (deliveryAlert) => {
+      let state: TaskState | undefined = undefined;
```
Angatupyry, line 389: Same suggestion as above. 
const instead of let and `filterAndPushDeliveryAlert(deliveryAlert, state);` function inside the try.

packages/dashboard/src/components/delivery-alert-store.tsx
```diff
@@ -0,0 +1,539 @@
...
+    return () => sub.unsubscribe();
+  }, [rmf]);
+
+  const onCancel = React.useCallback<Required<DeliveryWarningDialogProps>['onCancel']>(
+    async (delivery_alert_id, task_id) => {
+      try {
+        if (!rmf) {
+          throw new Error('tasks and delivery alert api not available');
+        }
+        await rmf.tasksApi?.postCancelTaskTasksCancelTaskPost({
```
Angatupyry, line 408: Maybe the '?' is not necessary

packages/dashboard/src/components/delivery-alert-store.tsx
```diff
@@ -0,0 +1,539 @@
...
+          type: 'cancel_task_request',
+          task_id: task_id,
+        });
+        await rmf.deliveryAlertsApi?.updateDeliveryAlertActionDeliveryAlertsDeliveryAlertIdActionPost(
```
Angatupyry, line 412: Maybe the '?' is not necessary

packages/dashboard/src/components/delivery-alert-store.tsx
```diff
@@ -0,0 +1,539 @@
...
+    [rmf, appController],
+  );
+
+  const onResume = React.useCallback<Required<DeliveryWarningDialogProps>['onResume']>(
+    async (delivery_alert_id, task_id) => {
+      try {
+        if (!rmf) {
+          throw new Error('delivery alert api not available');
+        }
+        await rmf.deliveryAlertsApi?.updateDeliveryAlertActionDeliveryAlertsDeliveryAlertIdActionPost(
```
Angatupyry, line 458: Maybe the '?' is not necessary

packages/dashboard/src/components/delivery-alert-store.tsx
```diff
@@ -0,0 +1,539 @@
...
+  // Closing on an error requires the additional step of setting the action
+  // to cancelled, as we want to ensure this error delivery dialog does not show
+  // up anymore when the dashboard is refreshed.
+  const onErrorCloseCancel = React.useCallback<Required<DeliveryErrorDialogProps>['onClose']>(
+    async (delivery_alert_id) => {
+      try {
+        if (!rmf) {
+          throw new Error('delivery alert api not available');
+        }
+        await rmf.deliveryAlertsApi?.updateDeliveryAlertActionDeliveryAlertsDeliveryAlertIdActionPost(
```
Angatupyry, line 487: Maybe the '?' is not necessary

Angatupyry (changes requested):
Thanks for this effort, Aaron!
This was really fast!

I left you minimal suggestions that you can consider.

packages/dashboard/src/components/delivery-alert-store.tsx
aaronchongth: https://github.com/open-rmf/rmf-web/pull/768/commits/83666bba825ec114cc679179300587c135db958a

packages/dashboard/src/components/delivery-alert-store.tsx
aaronchongth: https://github.com/open-rmf/rmf-web/pull/768/commits/83666bba825ec114cc679179300587c135db958a

packages/dashboard/src/components/delivery-alert-store.tsx
aaronchongth: https://github.com/open-rmf/rmf-web/pull/768/commits/83666bba825ec114cc679179300587c135db958a

packages/dashboard/src/components/delivery-alert-store.tsx
aaronchongth: https://github.com/open-rmf/rmf-web/pull/768/commits/83666bba825ec114cc679179300587c135db958a

packages/dashboard/src/components/delivery-alert-store.tsx
aaronchongth: Changed if statement in https://github.com/open-rmf/rmf-web/pull/768/commits/83666bba825ec114cc679179300587c135db958a

aaronchongth:
Thanks for the review @Angatupyry! I've further simplified the implementation, https://github.com/open-rmf/rmf-web/pull/768/commits/b082bdfc4792d06faa629806cda9fb82791b4ae1, as we are generating delivery alert IDs using timestamps, we can find out which ones are the newest.

This resolves a bug that I found, where newer warnings don't overwrite older warnings if they are not resolved.

xiyuoh:
Some comments/questions after testing onsite:
- Should `Cancel delivery` remain as an option? It is currently blacked out
wrong_cart
- There shouldn't be an `Override` option for Missing carts.
missing_cart_override
- Probably not related to this PR, but once I've selected `System Overview`, I can't select other tabs anymore

Angatupyry (approved):
LGTM!

aaronchongth:
@xiyuoh per DM, since updating action and cancelling task is in the same try except scope, the action is not updated if the task cancellation fails, while the button still becomes disabled. Disabling the button should come from feedback on whether the action has been updated instead.

aaronchongth:
Merging this to keep the diff small, will apply fixes if needed afterwards.

PR #775, opened 2023-09-19
Migration to React three fiber

Angatupyry:
## Migration to React three fiber

- Remove all react leaflet component 
- Using react three fiber instead. 

Pending:

- [ ] Labels overlapping. 
- [ ] Some hardcode z and scales values 
- [ ] Test from new react components
- [ ] It still happening that something the image doesn't load correctly and the frontend crash. I couldn't reproduce it properly yet.  
- [ ] Issue related #734 

[Screencast from 09-19-2023 07:41:42 PM.webm](https://github.com/open-rmf/rmf-web/assets/37310205/53f902aa-6ecf-41f0-8b78-0d57a912e21c)

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/map-app.tsx
+++ b/packages/dashboard/src/components/map-app.tsx
@@ -32,7 +32,7 @@ import { RobotSummary } from './robots/robot-summary';
 import { Box3, TextureLoader, Vector3 } from 'three';
 import { Canvas, useLoader } from '@react-three/fiber';
 import { Line } from '@react-three/drei';
-import { CameraControl, LayersController } from './three-fiber';
+import { CameraControl, LayersController, updateZoom } from './three-fiber';
 import { Lifts, Door, RobotThree } from './three-fiber';
 
 type FleetState = ApiServerModelsRmfApiFleetStateFleetState;
@@ -388,8 +388,8 @@ export const MapApp = styled(
               buildingMap.levels.find((l: Level) => l.name === value) || buildingMap.levels[0],
             );
           }}
-          handleZoomIn={() => AppEvents.zoom.next(zoom + 1)}
-          handleZoomOut={() => AppEvents.zoom.next(zoom - 1)}
+          handleZoomIn={() => AppEvents.zoom.next(updateZoom(zoom, 2))}
+          handleZoomOut={() => AppEvents.zoom.next(updateZoom(zoom, -2))}
         />
         <Canvas
           onCreated={({ camera }) => {
--- a/packages/dashboard/src/components/map-app.tsx
+++ b/packages/dashboard/src/components/map-app.tsx
@@ -223,30 +223,11 @@ export const MapApp = styled(
     }, [rmf]);
 
     const [imageUrl, setImageUrl] = React.useState<string | null>(null);
-    const [center, setCenter] = React.useState<L.LatLngTuple>([0, 0]);
     const [zoom, setZoom] = React.useState<number>(DEFAULT_ZOOM_LEVEL);
 
     React.useEffect(() => {
       const sub = AppEvents.zoom.subscribe((currentValue) => {
         setZoom(currentValue ?? DEFAULT_ZOOM_LEVEL);
-      });
-      return () => sub.unsubscribe();
-    }, []);
-
-    React.useEffect(() => {
-      const sub = AppEvents.mapCenter.subscribe((currentValue) => {
-        setCenter((prev) => {
-          const newCenter: L.LatLngTuple = [...currentValue];
-          // react-leaftlet does not properly update state when the previous LatLng is the same,
-          // even when a new array is passed.
-          if (prev[0] === newCenter[0]) {
-            newCenter[0] += 0.00001;
-          }
-          if (prev[1] === newCenter[1]) {
-            newCenter[1] += 0.00001;
-          }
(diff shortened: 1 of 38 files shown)


packages/react-components/lib/map/three-fiber/image-maker.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export interface ImageThreeProps {
+  level: Level;
+  imageUrl: string;
+}
+
+export const ReactThreeFiberImageMaker = ({ level, imageUrl }: ImageThreeProps): JSX.Element => {
+  let texture: Texture | undefined = undefined;
+  try {
+    texture = useLoader(TextureLoader, imageUrl);
```
aaronchongth, line 14: This is a tricky check failure (sorry I introduced this), however I'm unsure how we can be sure the `imageUrl` is accessible in the first place, these were some of the things I found, https://www.freecodecamp.org/news/how-to-validate-urls-in-javascript/.

aaronchongth (changes requested):
Thanks for helping with this huge migration! I think we're in pretty good shape now. There might be small issues here and there that pops up, but we can fix them later on.

I'll move this to a reviewable so CI can be run on it. Once we address all the check failures and CI passes, I think we are good.

packages/react-components/lib/map/three-fiber/image-maker.tsx
Angatupyry: It is currently working properly, could you try it?

packages/react-components/lib/map/three-fiber/shape-three-rendering.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+interface ShapeThreeRenderingProps {
+  position: [number, number, number];
+  color: string;
+  text?: string;
+  circleShape: boolean;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export const debounce = <T extends (...args: any[]) => void>(
```
aaronchongth, line 12: it looks like we're not using this function anywhere else at the moment, is it ok if we use
```
export const debounce = (callback: () => void, delay: number): (() => void) => {...}
```
This way we can avoid the eslint disabling

packages/dashboard/src/components/three-fiber/camera-control.tsx
```diff
@@ -0,0 +1,63 @@
...
+    // const handleWheel = (event: WheelEvent) => {
+    //   const delta = event.deltaY;
+    //   const sensitivity = 0.1;
+
+    //   setInternalZoom((prevZoom) => {
+    //     const newZoom = prevZoom - delta * sensitivity;
+    //     return Math.max(controls.minZoom, Math.min(controls.maxZoom, newZoom));
+    //   });
+
+    //   updateExternalStateDebounced(internalZoom);
```
aaronchongth, line 45: I see that `debounce` is used here as well, I think when we implement it later on, we can just re-write it with a different callback instead of using a template. (if we still need it in the later implementation)

aaronchongth (changes requested):
Just a comment left regarding linting, otherwise LGTM

packages/dashboard/src/components/three-fiber/camera-control.tsx
Angatupyry: Yes, definitely.
Do you suggest removing the commented code?

packages/react-components/lib/map/three-fiber/shape-three-rendering.tsx
Angatupyry: Done 87b60ccd91cea03f780ba75f79d229b47a1edc8e

packages/dashboard/src/components/three-fiber/camera-control.tsx
aaronchongth: oh sorry, I didn't mention what the action item was. I think it is ok as it is for now, just commented to make sure we understand what to do with debouncing later on

packages/dashboard/src/components/map-app.tsx
```diff
@@ -41,25 +42,31 @@ const debug = Debug('MapApp');
...
+      Waypoints: false,
+      Dispensers: false,
+      Ingestors: false,
+      Lifts: false,
+      Doors: false,
+      Trajectories: false,
+      Robots: false,
+      Labels: false,
+      PickupLabel: false,
+      DropoffLabel: false,
```
aaronchongth, line 68:
```suggestion
      Waypoint labels: false,
      Pickup point labels: false,
      Dropoff point labels: false,
```

My suggestion is not syntactically correct, but  the idea is to have spaces in the layer names.

aaronchongth (changes requested):
Thank you! Looks good, just a few tweaks and naming suggestions left

packages/dashboard/src/components/map-app.tsx
Angatupyry: Done 5a39c12c876182611593353d365cb7a307ca6e89 !
I separate the labels in case we have more in the future. For now, I think they are just doors. 

Then the waypoint labels and the dropoff and pickup labels.

aaronchongth (approved):
I think we're in a good state! LGMT, thanks for this huge effort!

PR #784, opened 2023-09-22
Support offline doc generation

aaronchongth:
## What's new

* Scripts for generating docs included as static files
* copies scripts into static directory before runs

## Self-checks


koonpeng (commented):
The offline swagger bundles should be included in the python package, might makes sense to just put them in `static/` instead of copying them over when starting. However, iirc we mapped the static path to `run/static`, so we need to rename the current "static" to "cache" or something (or serve the bundles from another path).

So we will have
* A `run` directory that is created at runtime, that will store "dynamic" data. This should be safe to delete, should also be configurable.
* A `static` directory that is bundled together with the api server. Should never be deleted, not configurable.

packages/api-server/sqlite_local_config.py
```diff
@@ -8,7 +8,7 @@
 config.update(
     {
         "db_url": f"sqlite://{run_dir}/db.sqlite3",
-        "static_directory": f"{run_dir}/static",  # The directory where static files should be stored.
+        "static_directory": "static",  # The directory where static files should be stored.
```
koonpeng, line 11: This will mix docs bundles with run time files, we should keep them separated. The docs bundle should be part of the python package and the run time path should be configurable.

koonpeng (commented):
The docs bundle should be added to the python package (need to be specified in `setup.py`).

aaronchongth:
I've refactored our static file usage to be cache instead, most of them were name changes. The real static files to be used are placed in the hardcoded `static` directory, that is not configurable.

packages/api-server/setup.py
```diff
@@ -38,5 +38,8 @@
             "rmf_api_server=api_server.__main__:main",
         ],
     },
+    package_data={
+        "static": ["*"],
+    },
```
koonpeng, line 43: the files are not exported, can test by running `pnpm run prepack`. According to the docs the format should be 
```python3
package_data={"mypkg": ["*.txt", "*.rst"]}
```
https://setuptools.pypa.io/en/latest/userguide/datafiles.html

packages/api-server/api_server/app.py
```diff
@@ -66,22 +66,23 @@ async def on_sio_connect(sid: str, _environ: dict, auth: Optional[dict] = None):
     allow_methods=["*"],
     allow_headers=["*"],
 )
-os.makedirs(app_config.static_directory, exist_ok=True)
+
 app.mount(
     "/static",
-    StaticFiles(directory=app_config.static_directory),
+    StaticFiles(directory="static"),
```
koonpeng, line 72:
```suggestion
    StaticFiles(directory=f"{__dir__}/static"), # may not be correct, check with the output `whl`
```
The directory need to be relative to the package's dir (e.g. when release in ros it will be some form of `/opt/ros/iron/lib/python-packages/...`.

packages/api-server/setup.py
aaronchongth: yeah that's a good idea, I was having trouble with `setup.py` to handle those files properly, they are in the `api_server` package and behaving now

packages/api-server/api_server/app.py
aaronchongth: Yup good call, using the absolute path of `app.py` works

aaronchongth:
Just tested without network access and it looks good so far

koonpeng (approved):
LGTM!

PR #791, opened 2023-10-06
Fix - schedule single event

### LINKED ISSUE #790: [Bug]: Editing or deleting single events not working as expected
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### OS and version

Ubuntu 22.04

### Open-RMF installation type

Binaries

### Other Open-RMF installation methods

_No response_

### Open-RMF version or commit hash

0.0.1-1jammy.20230920.014853

### ROS distribution

Humble

### ROS installation type

Binaries

### Other ROS installation methods

_No response_

### Package or library, if applicable

_No response_

### Description of the bug

Editing single events creates a new event, without deleting the previous event.
This might be due to `except` not working as expected.

Deleting single events, somehow deletes the event the day before.

A few possible explanations
* Time zone differences, which I'm experiencing
* Time zone differences due to a VPN I'm using
* The `except` field has been broken a few PRs ago

### Steps to reproduce the bug

1. RMF office simulation with dashboard and api-server
2. create a random task for schedule
3. edit single instance
4. delete single instance

https://github.com/open-rmf/rmf-web/assets/5383623/acbade70-4088-4118-bebf-46daedab6445

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional information or screenshots

_No response_

Angatupyry:
## What's new

Fix #790 

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/tasks/task-schedule.tsx
+++ b/packages/dashboard/src/components/tasks/task-schedule.tsx
@@ -179,7 +179,7 @@ export const TaskSchedule = () => {
 
       let exceptDate: string | undefined = undefined;
       if (eventScope === EventScopes.CURRENT) {
-        exceptDate = exceptDateRef.current.toISOString();
+        exceptDate = schedule.at.toISOString();
       }
 
       await Promise.all(
--- a/packages/dashboard/src/components/tasks/task-schedule.tsx
+++ b/packages/dashboard/src/components/tasks/task-schedule.tsx
@@ -179,7 +179,7 @@ export const TaskSchedule = () => {
 
       let exceptDate: string | undefined = undefined;
       if (eventScope === EventScopes.CURRENT) {
-        exceptDate = schedule.at.toISOString();
+        exceptDate = exceptDateRef.current.toISOString();
       }
 
       await Promise.all(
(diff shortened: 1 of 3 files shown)


packages/dashboard/src/components/tasks/task-schedule.tsx
```diff
@@ -179,7 +179,7 @@ export const TaskSchedule = () => {
 
       let exceptDate: string | undefined = undefined;
       if (eventScope === EventScopes.CURRENT) {
-        exceptDate = exceptDateRef.current.toISOString();
+        exceptDate = schedule.at.toISOString();
```
aaronchongth, line 182: I believe this change is not needed, as it will set `exceptDate` to today on the schedule's date, instead of the one selected for that event. You'll see it bugging out if you try to modify or delete an event next week, instead of today or tomorrow

aaronchongth (changes requested):
Thanks for debugging this, just left a small change that's needed

packages/dashboard/src/components/tasks/task-schedule.tsx
Angatupyry: You're right!
Thanks!

aaronchongth (approved):
LGTM

PR #802, opened 2023-10-12
Retain camera position after changing tabs

Angatupyry:
## What's new

[chrome-capture-2023-9-12 (1).webm](https://github.com/open-rmf/rmf-web/assets/37310205/761e187f-1173-4e4c-a6e2-0e4f81eb5c22)

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/three-fiber/camera-control.tsx
+++ b/packages/dashboard/src/components/three-fiber/camera-control.tsx
@@ -15,7 +15,6 @@ interface CameraControlProps {
 export const CameraControl: React.FC<CameraControlProps> = ({ zoom }) => {
   const { camera, gl } = useThree();
   const controlsRef = useRef<OrbitControls | null>(null);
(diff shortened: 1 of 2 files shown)


packages/dashboard/src/components/three-fiber/camera-control.tsx
```diff
@@ -50,7 +52,14 @@ export const CameraControl: React.FC<CameraControlProps> = ({ zoom }) => {
...
     };
-  }, [camera]);
+  }, [camera, cameraPosition]);
+
+  useEffect(() => {
+    const sub = AppEvents.cameraPosition.subscribe((currentValue) => {
+      setCameraPosition(currentValue);
+    });
+    return () => sub.unsubscribe();
+  }, []);
```
aaronchongth, line 62: This effect is not needed too

aaronchongth (changes requested):
Looks clean, thanks! I think there is no need for some of the statements listed below, since we are already updating the app state.

aaronchongth (approved):
Thanks for the quick iteration! Cheers

Issue #787, opened 2023-09-28
[Bug]: Doors in System Overview not populated properly


aaronchongth:
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### OS and version

Ubuntu 22.04

### Open-RMF installation type

Binaries

### Open-RMF version or commit hash

release-humble-230810

### ROS distribution

Humble

### ROS installation type

Binaries

### Description of the bug

With more doors, the doors-app in System Overview requires pagination and it looks like the pagination is not working properly.

https://github.com/open-rmf/rmf-web/assets/5383623/99b23ee6-9ca3-49c5-8743-4a58a4971f08

### Steps to reproduce the bug

Launch hotel world and dashboard.

PR #803, opened 2023-10-12
Fix/doors pagination

aaronchongth:
## What's new

* Fixes https://github.com/open-rmf/rmf-web/issues/787
* Sort names by default for robots, doors and lifts
* Turn off sorting for unrelated fields for doors (as it causes random order when fields have the same value and new states keep coming in)

## Self-checks


Angatupyry (approved):
LGTM!
Thanks, Aaron!

PR #809, opened 2023-10-17
Using icon to display robot

Angatupyry:
## What's new

- Display robot icons as images in react three fiber. 
- Make waypoints square smaller. 
- Make waypoints transparent

## Self-checks

### DIFF
--- a/packages/dashboard/src/managers/resource-manager-robots.ts
+++ b/packages/dashboard/src/managers/resource-manager-robots.ts
@@ -5,6 +5,7 @@ const debug = Debug('ResourceManager');
 export interface RobotResource {
   icons: Record<string, string>; // Record<ModelName|FleetName, IconPath>
   places: Record<string, string[]>; // Record<Places, Dispensers[]>
+  scale: number;
 }
 
 export class RobotResourceManager {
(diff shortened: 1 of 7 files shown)


aaronchongth (changes requested):
Thanks for putting this together, @Angatupyry! Unfortunately I don't see the scale reflected properly when I pull in assets from rmf_demos_dashboard_resources, it is currently scaling according to `DEFAULT_ROBOT_SCALE`.

I've been testing with the dashboard resources from https://github.com/open-rmf/rmf_demos/pull/196, let me know if my json is malformed

packages/dashboard/src/managers/resource-manager-robots.ts
```diff
@@ -53,6 +54,28 @@ export class RobotResourceManager {
...
+    if (!this.robots[fleetName].hasOwnProperty('scale')) {
+      debug(
+        `failed to load scale for "${fleetName}/${robotModel}" (fleet/model does not have an scale)`,
+      );
+      return null;
+    }
+
+    const robotScale = this.robots[fleetName].scale;
+
+    if (!!robotModel && robotScale.hasOwnProperty(robotModel)) {
```
aaronchongth, line 72:
```suggestion
    if (!!robotModel) {
```
We won't need to check for this `robotModel` property

aaronchongth (changes requested):
I think I found the issue!

edit: With the deployment issues we faced, could you help test in Microsoft Edge and without network connectivity too?

aaronchongth (approved):
Thanks for fixing it! LGTM

PR #819, opened 2023-10-26
Creation of e2e test for favorite tasks

Angatupyry:
## What's new

- e2e test for favorite task added

## Self-checks


aaronchongth (changes requested):
Thanks! Can we also test these scenarios
* open create-task dialog again, clicking on the previously saved favorite task, the fields are populated properly
* delete favorite task
* stretch goal, fix https://github.com/open-rmf/rmf-web/issues/820, and make sure we can't submit with the same name

aaronchongth (approved):
Thanks for setting this up! I've verified that it works locally, so I extended the timeout to 5 minutes to see if it is a latency issue on the side of the runner

PR #824, opened 2023-11-02
Display robot according to its level correctly 

Angatupyry:
## What's new

- Fix the problem that the robot kept appearing on the wrong floor after going up with the elevator to the next floor. 

[chrome-capture-2023-10-2.webm](https://github.com/open-rmf/rmf-web/assets/37310205/249753d5-6920-4d1c-af8c-5df9f88ce63a)

[chrome-capture-2023-10-2 (1).webm](https://github.com/open-rmf/rmf-web/assets/37310205/5f757103-eda6-468c-94b5-5d8e32c2bfa2)

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/map-app.tsx
+++ b/packages/dashboard/src/components/map-app.tsx
@@ -291,6 +291,7 @@ export const MapApp = styled(
             ?.filter(
               (r) =>
                 fleetState.robots &&
+                r in currentLevelOfRobots &&
                 currentLevelOfRobots[r] === currentLevel.name &&
                 `${fleetState.name}/${r}` in robotsStore,
             )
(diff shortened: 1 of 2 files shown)


packages/dashboard/src/components/map-app.tsx
```diff
@@ -316,6 +319,11 @@ export const MapApp = styled(
             return;
           }
           Object.entries(fleetState.robots).forEach(([robotName, robotState]) => {
+            setCurrentLevelOfRobots((prevState) => ({
+              ...prevState,
+              [robotName]: robotState.location?.map || 'L1',
```
aaronchongth, line 324: We should not set a default since there might be maps that do not have `L1` at all, I would suggest just not adding the robot into this record if the location is not available, since that is an issue that needs some attention. 

So in the case where it was already in the record, and somehow its location is missing, we should remove it from this record, and not render it. Probably should log an error too

packages/dashboard/src/components/map-app.tsx
```diff
@@ -288,14 +291,14 @@ export const MapApp = styled(
             ?.filter(
               (r) =>
                 fleetState.robots &&
-                fleetState.robots[r].location?.map === currentLevel.name &&
+                currentLevelOfRobots[r] === currentLevel.name &&
```
aaronchongth, line 294:
```suggestion
                r in currentLevelOfRobots &&
                currentLevelOfRobots[r] === currentLevel.name &&
```

aaronchongth (changes requested):

aaronchongth (approved):
LGTM

PR #830, opened 2023-11-10
Add query filter for delivery alerts

Angatupyry:
## What's new

- Field filters are added
- A new endpoint is created

[chrome-capture-2023-10-10.webm](https://github.com/open-rmf/rmf-web/assets/37310205/217440aa-4dce-4ef2-b4ba-53a2c042e2f2)

## Self-checks

### DIFF
--- a/packages/api-client/lib/openapi/api.ts
+++ b/packages/api-client/lib/openapi/api.ts
@@ -5918,6 +5918,70 @@ export const DeliveryAlertsApiAxiosParamCreator = function (configuration?: Conf
     },
     /**
      *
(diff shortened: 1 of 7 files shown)


aaronchongth (changes requested):
It works great! Thanks for adding this query endpoint.

Could you extend this PR to also
* re-generate the `api-client`with this new endpoint
* clean up this block, by querying only delivery alerts with action `waiting`, https://github.com/open-rmf/rmf-web/blob/deploy/hammer/packages/dashboard/src/components/delivery-alert-store.tsx#L418-L477

PR #840, opened 2023-11-24
Rearrange schedule view, with better word prompts

aaronchongth:
## What's new

* Separated `Day`, `Week`, `Month` view change buttons away from `Today` as it is confusing what those buttons do together
* Change text of `Today` button based on view

## Self-checks


aaronchongth:
Merging as this only affects the frontend.

PR #843, opened 2023-11-27
Feature/save request (#842)

aaronchongth:
* New route for saving and getting successfully dispatched requests
* Announce failed to dispatch instead of create task
* Generate api-client
* Use api task request instead of custom ttm pydantic model
* Revert alert text change

---------

(cherry picked from commit 52a7d142252508c472c701581b76a11462c60cae)

## Self-checks


Angatupyry (approved):
Just one thing I could not test, in the endpoint task/request, how do I send the ids, between commas?

Otherwise, it is perfect. 
LGTM!

aaronchongth:
yup! you can just query separating the task ids with commas, I'll be merging it first, we can review it with the subsequent PR where I am using it

PR #845, opened 2023-11-28
Hammer/permissions

aaronchongth:
## What's new

* `stub` to be admin
* only admins allowed to cancel tasks
* refactored task cancellation button into a component that handles enforcing permissions, events and toasts
* edit schedule series only available for admins
* removed all inspect buttons

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/tasks/task-alert.tsx
+++ b/packages/dashboard/src/components/tasks/task-alert.tsx
@@ -287,7 +287,7 @@ export function TaskAlertDialog({ alert, removeAlert }: TaskAlertDialogProps): J
         onDismiss={removeAlert}
         acknowledgedBy={taskAlert.acknowledgedBy}
         onAcknowledge={taskAlert.acknowledgedBy ? undefined : acknowledgeAlert}
-        onInspect={() => setOpenTaskInspector(true)}
+        // onInspect={() => setOpenTaskInspector(true)}
         title={taskAlert.title}
         progress={taskAlert.progress}
         alertContents={taskAlert.content}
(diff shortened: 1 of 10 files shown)


packages/dashboard/src/components/tasks/task-alert.tsx
```diff
@@ -287,7 +287,7 @@ export function TaskAlertDialog({ alert, removeAlert }: TaskAlertDialogProps): J
         onDismiss={removeAlert}
         acknowledgedBy={taskAlert.acknowledgedBy}
         onAcknowledge={taskAlert.acknowledgedBy ? undefined : acknowledgeAlert}
-        onInspect={() => setOpenTaskInspector(true)}
+        // onInspect={() => setOpenTaskInspector(true)}
```
Angatupyry, line 290: Maybe we can add a comment about why we comment on this line in this iteration.
Or simply delete it if it will no longer be used.

packages/dashboard/src/components/tasks/task-cancellation.tsx
```diff
@@ -0,0 +1,93 @@
...
+    return () => sub.unsubscribe();
+  }, [rmf, taskId]);
+
+  const taskCancellable =
+    taskState &&
+    taskState.status &&
+    !['canceled', 'killed', 'completed', 'failed'].includes(taskState.status);
+  const userCanCancelTask = profile && Enforcer.canCancelTask(profile);
+
+  function capitalizeFirstLetter(status: string): string {
```
Angatupyry, line 47: Maybe we can move it in the utils.ts file

Angatupyry (approved):
LGTM!

packages/dashboard/src/components/tasks/task-cancellation.tsx
aaronchongth: since it's only used here, I'll keep it here for now

packages/dashboard/src/components/tasks/task-alert.tsx
aaronchongth: https://github.com/open-rmf/rmf-web/pull/845/commits/dd8190de2092ce6df7a26088558a584c22a8118f

PR #847, opened 2023-11-28
Server-side sort and filter pickup and destination

aaronchongth:
## What's new

* adding pickup and destination to ttm task state model for filter and sorting purposes
* saving task request saves pickup and destination information into ttm task state model
* added pickup and  destination to query
* re-generated query api

## Self-checks


Angatupyry (approved):
I tested it and is works as expected!

LGTM!

PR #869, opened 2024-01-03
Hammer/fix uat

aaronchongth:
## What's new

- [x] More logs for schedule times and conversions
- [x] using `datetime.fromtimestamp(self.until.timestamp()))`
- [x] test with changed pod timezone
- [x] export csv to designate semicolons as delimiters for Excel
- [x] fix schedule glitching due to logging
- [x] change logs to debug logs instead
- [x] disables unused and confusing start time field
- [x] lift op mode to mode, from lift state
- [x] additional schedule event viewer component for entire task description

## Self-checks

### DIFF
--- a/packages/api-server/api_server/models/tortoise_models/scheduled_task.py
+++ b/packages/api-server/api_server/models/tortoise_models/scheduled_task.py
@@ -104,9 +104,8 @@ class ScheduledTaskSchedule(Model):
             print(f"setting at {self.at}")
             # job = job.at(self.at)
             # TODO(ac): not hardcode this timezone
-            job = job.at(
-                self.at, timezone("Asia/Singapore")
-            )  # pyright: ignore[reportGeneralTypeIssues]
+            tz = timezone("Asia/Singapore")  # pyright: ignore[reportGeneralTypeIssues]
+            job = job.at(self.at, tz)
 
         return job
 
--- a/packages/api-server/api_server/models/tortoise_models/scheduled_task.py
+++ b/packages/api-server/api_server/models/tortoise_models/scheduled_task.py
@@ -2,8 +2,8 @@ from datetime import datetime
 from enum import Enum
 from typing import Optional
 
+import pytz
 import schedule
-from pytz import timezone
 from schedule import Job
 from tortoise import Tortoise
 from tortoise.contrib.pydantic.creator import (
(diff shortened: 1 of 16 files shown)


packages/api-server/api_server/models/tortoise_models/scheduled_task.py
```diff
@@ -71,7 +72,15 @@ def to_job(self) -> Job:
             # schedule uses `datetime.now()`, which is tz naive
             # Assuming self.until is a datetime object with timezone information
             # Convert the timestamp to datetime without changing the timezone
-            job = job.until(datetime.utcfromtimestamp(self.until.timestamp()))
+            print(f"to_job until: {self.until}")
+            print(f"to_job until timestamp: {self.until.timestamp()}")
+            print(
+                f"to_job until utcfromtimestamp: {datetime.utcfromtimestamp(self.until.timestamp())}"
+            )
```
koonpeng, line 79: Don't think we need to print this.

packages/api-server/setup.py
```diff
@@ -26,7 +26,7 @@
         "tortoise-orm~=0.18.1",
         "pyjwt[crypto]~=2.4",
         "pydantic~=1.9",
-        "schedule~=1.1.0",
+        "schedule~=1.2.1",
```
koonpeng, line 29: Is the timezone feature not available in 1.1.0?

packages/dashboard/src/components/appbar.tsx
```diff
@@ -229,15 +229,21 @@ export const AppBar = React.memo(({ extraToolbarItems }: AppBarProps): React.Rea
       }
       if (!schedule) {
         await Promise.all(
-          taskRequests.map((request) =>
-            rmf.tasksApi.postDispatchTaskTasksDispatchTaskPost({
+          taskRequests.map((request) => {
+            console.log(`submitTask: ${request}`);
```
koonpeng, line 233: These `console.log` could be replaced with `debug`.

packages/api-server/setup.py
aaronchongth: nope, unfortunately, it was in 1.2.0, but 1.2.1 comes with some bug fixes.
https://github.com/dbader/schedule/pull/576
https://github.com/dbader/schedule/pull/600

packages/dashboard/src/components/appbar.tsx
aaronchongth: I'll clean them up as debug messages after the demo then

koonpeng (approved):

PR #881, opened 2024-01-19
New database model for Task Queue entries, to reduce network traffic

aaronchongth:
## What's new

* New `TaskQueueEntry` model to be used for Task Queue, instead of `TaskState` (which contains all the phases), to reduce network traffic
* Removed warn time feature for UX reasons
* Generated new `api-client`
* Frontend changes for this

## Self-checks


koonpeng (commented):
Can you explain how `TaskQueueEntry` improve perf? It seems to be the same as `TaskState`.

aaronchongth:
Yeah certainly, this hopes to resolve 2 things,
* `TaskState` carries a lot of nested information, TaskState->phases->events->detail. Basically carrying all the information and timestamps that we don't require on the frontend. I still need to retrieve logs and review, but it appears that for problematic tasks that require human-intervention, retrieval of 10 states (pagination) can take up to 3 or 4 seconds, as opposed to 70ms when retrieving 10 tasks that executed well.
* overall we still require server-side sorting and filtering, which means those fields need to be explicitly defined, as opposed to be in a `JSONField`. For maintainability (and not to interfere with automatically generated models from `rmf_api_msgs`), I thought it best to explicitly define a model for this table, the frontend will retrieve task states (with all information) as needed

let me know what you think, or if there are better ways to achieve this

koonpeng:
> * `TaskState` carries a lot of nested information, TaskState->phases->events->detail. Basically carrying all the information and timestamps that we don't require on the frontend. I still need to retrieve logs and review, but it appears that for problematic tasks that require human-intervention, retrieval of 10 states (pagination) can take up to 3 or 4 seconds, as opposed to 70ms when retrieving 10 tasks that executed well.

`TaskState` don't seem to contain events, phases etc, I only see them in task logs. Unless RMF is including all the data in the raw message, but even so, we still need to save the data. We can fetch only the data that we need, avoiding this json. Another thing we can do is to change it from `JsonField` to `TextField` so that both postgres and tortoise does not try to process it. But moving the json to another table may also help, depending on the database internals, it may improve indexing performance and data locality.

I think what we can do are:
* Keep `TaskState` as is, just improve the way we fetch data.
* Change `data` to `TextField`, since we are changing schema already, might as well also move it to another, but don't rename `TaskState`, instead call the new table `RawTaskState` or something.
* Keep `TaskState` as in, but do not populate the `data` field anymore. Instead create a new `RawTaskState` table and populate the json there instead.

aaronchongth:
Per VC, what we know now,
* the task phases coming in are accurate, just that it increases without limit over time as robots get stuck

what we should do,
* fix dispatcher with proper unique task IDs
* drop deployment db task state tables, start cleaner (or use a script to drop just the phases of each task)
* fix in the backend how many events we should actually add

other TODOs afterwards,
* use a separate endpoint for alert acknowledgement, instead of appending to phases
* perhaps move pickup and destination, to request, use new api's to query those (but not sure how to sort or filter across multiple database models for now). Maybe the request for the task should just be embedded in `TaskState`, so there is a single source of truth

aaronchongth:
Changing to draft, as this involves more discussions

PR #882, opened 2024-01-19
except date quickfix

koonpeng:
## What's new

Quick fix (untested) for `except_date` being applied to the wrong day. I am working on a more robust fix that address some of the problem with the way `except_date` is currently handled, but that would require some extra changes to make it possible to unittest schedules so we should probably wait till the branch gets merged to `main`.

## Self-checks


aaronchongth (approved):
Tested and working! Thanks for fixing this!

PR #886, opened 2024-01-24
[api-server] fix typings, update dependencies

koonpeng:
## What's new

* Fix typing issues, reduce use of casting.
* Update `rxpy` (now called `reactivex`) to improve typings
* Update fastapi to fix some typings

tortoise-orm and pydantic are pretty outdated as well. We should look into updating them soon.

## Self-checks


aaronchongth (changes requested):
Thanks for handling this! Just did a quick review, and things look good so far since there're not really any behavioral changes

Moving forwards, should we be avoiding any use of `typing` (where possible), and try to stick to python's base types?

I did notice that the door and lift app is broken now, probably due to model changes. You can check by navigating to `localhost:3000/robots`

koonpeng:
> Moving forwards, should we be avoiding any use of typing (where possible), and try to stick to python's base types?

Yeah in general we should aim for that, but in practice it may be hard so I think it's not something that we need to enforce.

> I did notice that the door and lift app is broken now, probably due to model changes. You can check by navigating to localhost:3000/robots

Thanks for pointing out, I will look into it.

koonpeng:
> I did notice that the door and lift app is broken now, probably due to model changes. You can check by navigating to localhost:3000/robots

Should be fixed now.

aaronchongth (approved):
Sorry for the delay! Thanks for the fix, I don't see any issues running through sim

PR #887, opened 2024-01-24
Query just the information needed for Task Queue table

aaronchongth:
## What's new

* New API to return only the information required for the Task Queue table

## Self-checks


packages/api-server/api_server/models/tortoise_models/tasks.py
```diff
@@ -82,4 +82,7 @@ class TaskFavorite(Model):
     user = CharField(255, null=False, index=True)
 
 
+TaskQueueEntryPydantic = pydantic_model_creator(
+    TaskState, exclude=("data", "category", "unix_millis_warn_time")
+)
```
koonpeng, line 87: I think we should start avoiding use of `pydantic_model_creator` as it is not type safe. It's a bit troublesome to write it by hand but I think it is worth it. In the future we can consider a new generator that keep typings.

packages/api-server/api_server/repositories/tasks.py
```diff
@@ -242,6 +243,34 @@ async def get_task_state(self, task_id: str) -> Optional[TaskState]:
...
+                "requester",
+                "pickup",
+                "destination",
+            )
+            entries = []
+            for r in results:
+                status = r["status"]
+                if "Status." in status:
+                    r["status"] = r["status"].split("Status.")[1]
+                entries.append(TaskQueueEntryPydantic(**r))
```
koonpeng, line 269: I think we should avoid massaging the data for perf and consistency reasons, you can massage it on the frontend instead. Or at least refactor it so it is consistent, e.g., add the option to filter values to `query_task_states` and just have `query_task_queue_entry` call it with the filters.

packages/api-server/api_server/routes/tasks/tasks.py
```diff
@@ -138,6 +139,74 @@ async def get_task_state(
     return result
 
 
+@router.get("/queue_entry", response_model=List[TaskQueueEntryPydantic])
+async def query_task_queue_entry(
```
koonpeng, line 143: Add docs explaining what it does, how is it different from querying task state.

koonpeng (changes requested):

packages/api-server/api_server/repositories/tasks.py
aaronchongth: Gotcha, I've moved the massaging to the frontend now

packages/api-server/api_server/models/tortoise_models/tasks.py
aaronchongth: Good call, I've created a separate model for returning

packages/api-server/api_server/routes/tasks/tasks.py
aaronchongth: https://github.com/open-rmf/rmf-web/pull/887/commits/14261193e7566e734d847a18d78e9b11f7680d49

aaronchongth:
Converting back to draft since we have identified we don't have a bandwidth issue for now

PR #888, opened 2024-01-25
Seconds and milliseconds of start time to be zero

aaronchongth:
## What's new

Zero out seconds and milliseconds of schedule start times, to avoid skipping scheduled tasks

## Self-checks


koonpeng (approved):
I can see how it makes sense as the UI only goes to minutes, but why would it cause schedules to be skipped?

aaronchongth:
The scheduled jobs are not technically skipped. The `schedule` library has no concept of `start_from` date or times, so it is not able to schedule something for example every single day, but starting tomorrow.

In our current implementation, the job gets triggered. Just that within the callback, we check if the current date and time is before `start_from` date and time, we just return if it is before

so in this case, the schedule gets triggered within 10ms of the minute, however our `start_from` time is > 10ms, so the callback just returns.

koonpeng:
Thanks for the explanation.

PR #896, opened 2024-02-10
Schedule deletion: fix date differences due to different timezones, more tests

aaronchongth:
## What's new

* Added more tests on the route level
* Convert incoming except dates (in UTC) to any local timezones defined for the server, to determine the dates for `except_dates` 
* Fix the dashboard renders schedules.
  * deleting a task on day 1 before 8am results in day 2's task not being rendered, because day 1 is added into `except_date`, but `toISOFormat` checks in UTC, which makes day 1 -> day 0, and day 2 -> day 1, therefore failing to render day 2's

Before:
* we extract the date (YYYY-MM-DD) from the provided UTC dates, however this doesn't take into account of the case where UTC and server TZ has different dates (before Singapore 8am, for example). Currently this affects deleting or editing schedules before 8am only
* Example scenario, deleting a schedule at `05:00 GMT+8 2024-02-09`, will result in `21:00 GMT+0 2024-02-08`, where we will only check if the date `2024-02-08` if it should run. When  `05:00 GMT+8 2024-02-09` comes, the scheduler will see that only `2024-02-08` is exempted, while `2024-02-09` is not, and still proceeds to perform the schedule
* rendering issue

https://github.com/open-rmf/rmf-web/assets/5383623/74ae271a-2ef5-4632-9734-8c18aeea9441

Now:
* Convert received UTC dates into server TZ, before formatting the same way and extracting the date
* In the example scenario, after changing  `21:00 GMT+0 2024-02-08` back to server TZ `05:00 GMT+8 2024-02-09`, it will save `2024-02-09` as exempted, and when the scheduled time arrives, it will skip the schedule as intended.
* rendering fixed

https://github.com/open-rmf/rmf-web/assets/5383623/0fb3e9ea-b94c-466f-8acd-fd8861859702

Next:
* schedule editing, this bug also exists in editing, however those routes will also benefit from some refactoring, hence will be excluded from this PR for now

## Self-checks

### DIFF
--- a/packages/api-server/api_server/routes/tasks/scheduled_tasks.py
+++ b/packages/api-server/api_server/routes/tasks/scheduled_tasks.py
@@ -163,14 +163,17 @@ async def del_scheduled_tasks_event(
     if task is None:
         raise HTTPException(404)
 
-    utctz = ZoneInfo("UTC")
-    localtz = ZoneInfo(app_config.timezone) if app_config.timezone else None
+    tz = ZoneInfo(app_config.timezone) if app_config.timezone else None
+    event_date_str = event_date.replace(tzinfo=tz).isoformat()
 
-    event_date_str = event_date.replace(tzinfo=utctz).isoformat()
-    if localtz is not None:
-        event_date_utc = event_date.replace(tzinfo=utctz)
-        event_date_local = event_date_utc.astimezone(localtz)
-        event_date_str = event_date_local.isoformat()
+    # utctz = ZoneInfo("UTC")
+    # localtz = ZoneInfo(app_config.timezone) if app_config.timezone else None
+
+    # event_date_str = event_date.replace(tzinfo=utctz).isoformat()
+    # if localtz is not None:
+    #     event_date_utc = event_date.replace(tzinfo=utctz)
+    #     event_date_local = event_date_utc.astimezone(localtz)
+    #     event_date_str = event_date_local.isoformat()
 
     task.except_dates.append(event_date_str[:10])
     await task.save()
--- a/packages/api-server/api_server/routes/tasks/scheduled_tasks.py
+++ b/packages/api-server/api_server/routes/tasks/scheduled_tasks.py
@@ -163,8 +163,15 @@ async def del_scheduled_tasks_event(
     if task is None:
         raise HTTPException(404)
 
-    tz = ZoneInfo(app_config.timezone) if app_config.timezone else None
-    event_date_str = event_date.replace(tzinfo=tz).isoformat()
+    utctz = ZoneInfo("UTC")
+    localtz = ZoneInfo(app_config.timezone) if app_config.timezone else None
+
+    event_date_str = event_date.replace(tzinfo=utctz).isoformat()
+    if localtz is not None:
+        event_date_utc = event_date.replace(tzinfo=utctz)
+        event_date_local = event_date_utc.astimezone(localtz)
+        event_date_str = event_date_local.isoformat()
+
     task.except_dates.append(event_date_str[:10])
     await task.save()
 
@@ -175,37 +182,6 @@ async def del_scheduled_tasks_event(
         await schedule_task(task, task_repo)
     except schedule.ScheduleError as e:
         raise HTTPException(422, str(e)) from e
-
-
-# @router.put("/{task_id}/clear")
-# async def del_scheduled_tasks_event(
-#     task_id: int,
-#     event_date: datetime,
-#     task_repo: TaskRepository = Depends(task_repo_dep),
-# ):
-#     task = await get_scheduled_task(task_id)
-#     if task is None:
-#         raise HTTPException(404)
-
-#     utctz = ZoneInfo('UTC')
-#     localtz = ZoneInfo(app_config.timezone) if app_config.timezone else None
-
-#     event_date_str = event_date.replace(tzinfo=utctz).isoformat()
-#     if localtz is not None:
-#         event_date_utc = event_date.replace(tzinfo=utctz)
-#         event_date_local = event_date_utc.astimezone(localtz)
-#         event_date_str = event_date_local.isoformat()
-
-#     task.except_dates.append(event_date_str[:10])
-#     await task.save()
-
-#     for sche in task.schedules:
-#         schedule.clear(sche.get_id())
-
-#     try:
-#         await schedule_task(task, task_repo)
-#     except schedule.ScheduleError as e:
-#         raise HTTPException(422, str(e)) from e
 
 
 @router.post(
--- a/packages/api-server/api_server/routes/tasks/scheduled_tasks.py
+++ b/packages/api-server/api_server/routes/tasks/scheduled_tasks.py
@@ -163,14 +163,28 @@ async def del_scheduled_tasks_event(
     if task is None:
         raise HTTPException(404)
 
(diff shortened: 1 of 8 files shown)


packages/api-server/api_server/routes/tasks/scheduled_tasks.py
```diff
@@ -163,8 +163,15 @@ async def del_scheduled_tasks_event(
     if task is None:
         raise HTTPException(404)
 
-    tz = ZoneInfo(app_config.timezone) if app_config.timezone else None
-    event_date_str = event_date.replace(tzinfo=tz).isoformat()
+    utctz = ZoneInfo("UTC")
+    localtz = ZoneInfo(app_config.timezone) if app_config.timezone else None
+
+    event_date_str = event_date.replace(tzinfo=utctz).isoformat()
```
koonpeng, line 169: These conversion are assuming the input is in utc, which may not be the case. The better way to fix this is to either:
* include the tz in the `except_dates`
* check the tz and convert only if it is different from local

koonpeng (changes requested):

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
aaronchongth: Good point, this would be a good step towards removing any timezone assumptions

I have just made the changes accordingly over 3 commits, https://github.com/open-rmf/rmf-web/pull/896/commits/8088ed4e9d7aa7a19513a84bec1b31786f226d61, https://github.com/open-rmf/rmf-web/pull/896/commits/1a13df309178ae058edaadb0771483a1238685bf, https://github.com/open-rmf/rmf-web/pull/896/commits/6c7af0c6c49fb0fb6a8b69b2c8e452d9ac38a34f

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
```diff
@@ -159,18 +159,54 @@ async def del_scheduled_tasks_event(
...
+    # If the event date has time zone information, we check if it is in the same
+    # time zone as the server
+    event_date_utc_offset = event_date.utcoffset()
+    server_time_utc_offset = datetime.now(tz=server_tz_info).utcoffset()
+    if event_date_utc_offset is not None and server_time_utc_offset is not None:
+        event_tz_utc_offset_seconds = round(event_date_utc_offset.total_seconds())
+        server_tz_utc_offset_seconds = round(server_time_utc_offset.total_seconds())
+        logger.info(
+            f"Event tz utc offset: {event_tz_utc_offset_seconds}, server tz utc offset: {server_tz_utc_offset_seconds}"
+        )
```
koonpeng, line 184: Don't think we need to log at info level here, debug level should be good.

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
```diff
@@ -159,18 +159,54 @@ async def del_scheduled_tasks_event(
...
+
+        # if the time zones are the same, we can extract the date directly
+        if event_tz_utc_offset_seconds == server_tz_utc_offset_seconds:
+            logger.info("Event and server are in the same timezone")
```
koonpeng, line 188: Don't think we need to log at info level here, debug level should be good.

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
```diff
@@ -159,18 +159,54 @@ async def del_scheduled_tasks_event(
...
+            event_date_str = event_date.isoformat()
+        # otherwise, we convert the date before extracting the date
+        else:
+            logger.info("Event and server are not in the same timezone")
```
koonpeng, line 192: Don't think we need to log at info level here, debug level should be good.

koonpeng (approved):

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
```diff
@@ -159,18 +159,54 @@ async def del_scheduled_tasks_event(
...
+        logger.error(f"Task with scehdule id {task_id} not found")
         raise HTTPException(404)
 
-    tz = ZoneInfo(app_config.timezone) if app_config.timezone else None
-    event_date_str = event_date.replace(tzinfo=tz).isoformat()
+    # Server time zone, if not defined to use UTC
+    server_tz_info = (
+        ZoneInfo(app_config.timezone) if app_config.timezone else ZoneInfo("UTC")
+    )
+    logger.info(f"Server tz: {server_tz_info}")
```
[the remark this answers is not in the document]
aaronchongth: I'm avoiding switching logging level to debug for now, as that will flood the logs with all other information, for this particular fix I'll keep it to info first, but will revert it soon

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
aaronchongth: I'm avoiding switching logging level to debug for now, as that will flood the logs with all other information, for this particular fix I'll keep it to info first, but will revert it soon

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
aaronchongth: I'm avoiding switching logging level to debug for now, as that will flood the logs with all other information, for this particular fix I'll keep it to info first, but will revert it soon

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
aaronchongth: I'm avoiding switching logging level to debug for now, as that will flood the logs with all other information, for this particular fix I'll keep it to info first, but will revert it soon

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
```diff
@@ -159,18 +159,54 @@ async def del_scheduled_tasks_event(
...
+
+    # If the event date has time zone information, we check if it is in the same
+    # time zone as the server
```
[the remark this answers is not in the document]
aaronchongth: Hmm true. I'd say that it is a rather safe restriction though. Users must ensure that the schedules created (into the database), and the server must remain in the same tz after starting the deployment.

If any tz changes are required, they should start from a fresh db and server

added a FIXME still, https://github.com/open-rmf/rmf-web/pull/896/commits/c3b939d4a36f271ec47b8285bed88aed34f07d04

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
```diff
@@ -159,18 +159,54 @@ async def del_scheduled_tasks_event(
...
     if task is None:
+        logger.error(f"Task with scehdule id {task_id} not found")
         raise HTTPException(404)
 
-    tz = ZoneInfo(app_config.timezone) if app_config.timezone else None
-    event_date_str = event_date.replace(tzinfo=tz).isoformat()
+    # Server time zone, if not defined to use UTC
+    server_tz_info = (
+        ZoneInfo(app_config.timezone) if app_config.timezone else ZoneInfo("UTC")
+    )
```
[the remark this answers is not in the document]
aaronchongth: https://github.com/open-rmf/rmf-web/pull/896/commits/86852786e3367b59332e6d75b0cbd1b67b51134d

PR #902, opened 2024-02-17
Update schedule with timezone information

aaronchongth:
## What's new

Similar to https://github.com/open-rmf/rmf-web/pull/896, but for schedule editing

* Convert schedule update exempted date to server timezone before appending to except
* make dashboard edit except date have time zone information
* tests

## Self-checks

### DIFF
--- a/packages/api-server/api_server/routes/tasks/scheduled_tasks.py
+++ b/packages/api-server/api_server/routes/tasks/scheduled_tasks.py
@@ -228,6 +228,7 @@ async def update_schedule_task(
     task_repo: TaskRepository = Depends(task_repo_dep),
 ):
     try:
+        logger.info(f"Updating scheduled task [{task_id}]")
         task = await get_scheduled_task(task_id)
         if task is None:
             raise HTTPException(404)
@@ -245,8 +246,13 @@ async def update_schedule_task(
                 await task.save()
 
                 for sche in task.schedules:
+                    logger.info(f"Clearing schedule: {sche.get_id()}")
                     schedule.clear(sche.get_id())
 
+                pydantic_schedule_task = (
+                    await ttm.ScheduledTaskPydantic.from_tortoise_orm(task)
+                )
+                logger.info(f"Re-scheduling task: {pydantic_schedule_task}")
                 await schedule_task(task, task_repo)
 
                 scheduled_task = await ttm.ScheduledTask.create(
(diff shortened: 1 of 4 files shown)


packages/api-server/api_server/routes/tasks/test_scheduled_tasks.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
```
koonpeng, line 1: Don't disable the linter without a valid reason. I think the line count is probably due to duplicated code, see if it is possible to refactor them. If that is still an issue, then disable the rule globally (might want to do it regardless, imo this rule is not useful at all).

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
```diff
@@ -234,7 +239,8 @@ async def update_schedule_task(
 
         async with tortoise.transactions.in_transaction():
             if except_date:
-                event_date_str = except_date.isoformat()
+                event_date_str = convert_date_server_timezone_iso_str(except_date)
+                logger.info(f"Updating event with iso format date: {event_date_str}")
```
koonpeng, line 243: Include the subject (task id).

koonpeng (changes requested):

packages/api-server/api_server/routes/tasks/test_scheduled_tasks.py
aaronchongth: Hmm, but this is only for testing though. We could refactor it and re-used code between tests, but I was trying to prevent any stack tracing, even if the checks are repeated over tests, it doesn't make much sense to have the errors on the same lines across tests.

koonpeng: I think it is really up for debate whether code duplication is justified to make test more clear. I don't really have any strong opinions, if you think that it is ok, then just disable the lint globally. imo that is not a useful rule anyway, regardless of the reasons.

packages/api-server/api_server/routes/tasks/scheduled_tasks.py
aaronchongth: Thanks for catching that, that'll be useful. I've also added more logs in the route function
https://github.com/open-rmf/rmf-web/pull/902/commits/66db89323ac657088d4118dbbeba626fe35a1ae6

We can move logs to debug during the next phase of the project when these features stabilize

aaronchongth:
Overall the update route is better off refactored into 2 routes, one for updating an entire schedule, the other for updating just an event. I've noted it down as a next step in our project, but for now this PR is a required fix without breaking API

koonpeng (approved):

PR #904, opened 2024-02-19
Hammer/UI fixes

aaronchongth:
## What's new

More UI related fixes
* Stop appending to task logs when acknowledging alerts, since there is no longer a task phase inspector. This fixes the server stalling issue when more alerts are acknowledged
* Wrap map image handling in `ErrorBoundary`, so it doesn't crash the whole dashboard when the building map server is down. (linter was not happy with try-catch wrapping `useLoader`, so I followed the guide from https://github.com/pmndrs/react-three-fiber/discussions/1881)

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/map-app.tsx
+++ b/packages/dashboard/src/components/map-app.tsx
@@ -14,6 +14,7 @@ import {
   TextThreeRendering,
   RobotData,
 } from 'react-components';
+import { ErrorBoundary } from 'react-error-boundary';
 import { EMPTY, merge, scan, Subscription, switchMap } from 'rxjs';
 import appConfig from '../app-config';
 import { ResourcesContext } from './app-contexts';
@@ -637,7 +638,9 @@ export const MapApp = styled(
               ))
             : null}
           {currentLevel.images.length > 0 && imageUrl && (
-            <ReactThreeFiberImageMaker level={currentLevel} imageUrl={imageUrl} />
+            <ErrorBoundary fallback={<></>}>
+              <ReactThreeFiberImageMaker level={currentLevel} imageUrl={imageUrl} />
+            </ErrorBoundary>
           )}
           {!disabledLayers['Robots'] &&
             robots.map((robot) => {
--- a/packages/dashboard/src/components/map-app.tsx
+++ b/packages/dashboard/src/components/map-app.tsx
@@ -1,4 +1,4 @@
-import { Box, styled, Typography, useMediaQuery, useTheme } from '@mui/material';
+import { Box, styled, Typography, useMediaQuery } from '@mui/material';
 import { BuildingMap, ApiServerModelsRmfApiFleetStateFleetState, Level, Lift } from 'api-client';
 import { Door as DoorModel } from 'rmf-models';
 import Debug from 'debug';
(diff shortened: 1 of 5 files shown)


koonpeng (changes requested):

packages/dashboard/src/components/map-app.tsx
```diff
@@ -637,7 +638,9 @@ export const MapApp = styled(
               ))
             : null}
           {currentLevel.images.length > 0 && imageUrl && (
-            <ReactThreeFiberImageMaker level={currentLevel} imageUrl={imageUrl} />
+            <ErrorBoundary fallback={<></>}>
```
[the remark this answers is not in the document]
aaronchongth: https://github.com/open-rmf/rmf-web/pull/904/commits/c777ac3e0d557f4bc4b227df4ce0bcd6c522da3b

packages/dashboard/src/components/map-app.tsx
```diff
@@ -637,7 +639,20 @@ export const MapApp = styled(
               ))
             : null}
           {currentLevel.images.length > 0 && imageUrl && (
-            <ReactThreeFiberImageMaker level={currentLevel} imageUrl={imageUrl} />
+            <ErrorBoundary
+              fallback={<></>}
```
koonpeng, line 643: fallback should not be empty

koonpeng (changes requested):

packages/dashboard/src/components/map-app.tsx
aaronchongth: Fair enough :boom: 

image

koonpeng: 1. Styling could be better
2. No need to display the map if there is an error
3. No need for toast anymore, just add to visual clutter

aaronchongth: > 1. Styling could be better

Any other styling other than the bare minimum could cause a retrieval from the internet, which will break in an airgapped environment. I just used the theme, and the same font we have added statically to prevent that.

Style for this is also very low priority.

> 2. No need to display the map if there is an error

The issue is only from failing to load an image (it could be missing from cache), so there should be no reason, the rest of the map is not rendered (lifts, doors, etc). This will at least still allow the dashboard to be used.

> 3.

Toast removed

> 4.

Made it short and simple. Anything else regarding the error can be read from console errors.

image

koonpeng: I think a missing image is a broken map, but it seems you want to ignore the error and continue? I think the image is an important enough part to just fail the whole map but I'm not totally against continuing either, if that is the case, then I guess we can revert back to the empty map + toast.

koonpeng (changes requested):
Let's revert back to the empty map + toast if the idea is to continue despite the error.

packages/dashboard/src/components/map-app.tsx
aaronchongth: yeah sounds good. I'm mainly coming from a UX perspective. With a misplaced/missing image based on the URL, we will just be missing the floor plan, whereas users can still monitor where the robots are relative to waypoints.

In this case, we won't "break" production, and the fix would just be to ensure the building map server can be brought up again, without disruption to users.

koonpeng (approved):

PR #907, opened 2024-02-28
api-server: fix docs page

koonpeng:
## What's new

Fix the broken docs after we switched to serving swagger ui locally. Also fix the root path when serving the docs, now we don't have to hack index.html anymore.

https://open-rmf.github.io/rmf-web/

## Self-checks


aaronchongth (approved):
Looks good, thanks!

PR #909, opened 2024-02-28
Hammer/max downloads

aaronchongth:
## What's new

* New API for getting count of tasks
* Exporting CSV now queries in chunks instead of all in one go
* New global backdrop loading screen and events to trigger it, lets users know that the exporting is underway (which could take sometime)
* Made export and refresh buttons more explicit

https://github.com/open-rmf/rmf-web/assets/5383623/c3b329fd-bf99-4371-b59a-128a3dd42cae

Normal resolution
image

Low resolution
image

## Self-checks


packages/api-server/api_server/dependencies.py
```diff
@@ -10,7 +10,7 @@
 
 
 def pagination_query(
-    limit: int | None = Query(None, gt=0, le=1000, description="defaults to 100"),
+    limit: int | None = Query(None, gt=0, le=5000, description="defaults to 100"),
```
koonpeng, line 13: tbh 1000 is already pushing it, what is the reason for such a high limit?

packages/api-client/schema/index.ts
```diff
@@ -798,7 +798,7 @@ export default {
             required: false,
             schema: {
               title: 'Limit',
-              maximum: 1000.0,
+              maximum: 5000.0,
```
koonpeng, line 801: iirc this is generated, you should update the source instead.

koonpeng (changes requested):

packages/api-server/api_server/dependencies.py
aaronchongth: there may be more than 30 scheduled tasks a day for the ongoing deployment, and we are required to be able to export a month's worth of tasks, so 1000 won't be enough. I set an arbitrary 5000 to just cover more ground.
What about 2000?

packages/api-client/schema/index.ts
aaronchongth: yeah I regenerated the `api-client` and this was automatically populated, I didn't modify it myself

packages/api-server/api_server/dependencies.py
koonpeng: > there may be more than 30 scheduled tasks a day for the ongoing deployment, and we are required to be able to export a month's worth of tasks, so 1000 won't be enough. I set an arbitrary 5000 to just cover more ground. What about 2000?

The limit shouldn't be an issue, in fact we should reduce it to say 100, and default to around 20. The frontend can/should do pagination to retrieve as many tasks as needed.

aaronchongth: Hmm, so in the case of trying to dump all the task states from DB, we should have the frontend query for 1-100, 101-200, 201-300 and so on, with separate calls?

aaronchongth: I've added endpoints for counting and made the frontend query by chunks

Also added a global loading backdrop to let users know that something is happening

koonpeng: revert the limit to 1000.

packages/api-server/api_server/routes/tasks/tasks.py
```diff
@@ -121,8 +119,102 @@ async def query_task_states(
             if status_string not in valid_values:
                 continue
             filters["status__in"].append(mdl.Status(status_string))
+    return DbTaskState.filter(**filters)
 
-    return await task_repo.query_task_states(DbTaskState.filter(**filters), pagination)
+
+@router.get("/count", response_model=int)
+async def task_states_count(
```
koonpeng, line 126: I don't think we should have this endpoint, counting the number of rows is an expensive db operation.

packages/dashboard/src/components/tasks/tasks-app.tsx
```diff
@@ -245,23 +246,42 @@ export const TasksApp = React.memo(
...
+          await rmf.tasksApi.taskStatesCountTasksCountGet(
+            undefined,
+            undefined,
+            undefined,
+            undefined,
+            undefined,
+            undefined,
+            undefined,
+            undefined,
+            `0,${timestamp.getTime()}`,
```
koonpeng, line 259: Do we need to get every tasks? Can we limit it to the last X days?

packages/dashboard/src/components/tasks/tasks-app.tsx
```diff
@@ -270,23 +290,29 @@ export const TasksApp = React.memo(
...
-          .data;
-
         const taskRequestMap: Record<string, TaskRequest> = {};
-        let requestIndex = 0;
-        for (const id of taskIds) {
-          if (requestIndex < taskRequests.length && taskRequests[requestIndex]) {
-            taskRequestMap[id] = taskRequests[requestIndex];
+        const allTaskIds: string[] = tasks.map((task) => task.booking.id);
+        const queriesRequired = Math.ceil(allTaskIds.length / QueryLimit);
+        for (let i = 0; i < queriesRequired; i++) {
```
koonpeng, line 296: We don't need `queriesRequired` (and the task count endpoint), the frontend can just queries until there is no more result.

A potential use of the task count is to provide a progress gauge but I don't think we are doing it, but if so, we can try using the task date to give an estimated progress.

koonpeng (changes requested):

packages/api-server/api_server/routes/tasks/tasks.py
aaronchongth: Hmmm, I didn't know this was the case, thanks for the heads up. I'll do pagination directly

packages/dashboard/src/components/tasks/tasks-app.tsx
aaronchongth: yeah it makes sense, I'll start off with 31 days, which should satisfy the requirements.

packages/api-server/api_server/routes/tasks/tasks.py
aaronchongth: https://github.com/open-rmf/rmf-web/pull/909/commits/40b91f0941bff7e7123fa9fb8cf3a284f0c57e86

packages/dashboard/src/components/tasks/tasks-app.tsx
aaronchongth: https://github.com/open-rmf/rmf-web/pull/909/commits/1de251a91f11ab970c2f114b16c69fe514ff7b63

packages/dashboard/src/components/tasks/tasks-app.tsx
aaronchongth: https://github.com/open-rmf/rmf-web/pull/909/commits/1de251a91f11ab970c2f114b16c69fe514ff7b63

packages/dashboard/src/components/tasks/tasks-app.tsx
koonpeng: `exportTasksToCsv` still fetches all tasks, it is also doing unneeded double query, first it query for the task states then it query or the task requests, it could just query the task requests directly.

koonpeng (changes requested):

packages/dashboard/src/components/tasks/tasks-app.tsx
aaronchongth: This is a workaround until we get pickup and destination information into labels, basically using https://github.com/open-rmf/rmf_ros2/pull/328 (we had this conversation over VC to use labels for warning time and other pieces of information)

We are currently using task requests to parse for pickup and destination information as it is the only source available, since phases change over time. Task states also do not contain it's request.

Not all task requests are getting queried in `getPastMonthTaskRequests` too, just the requests of the task states that we got from `getPastMonthTasks`, and also queried in chunks of 100.

Added a fixme to make the intention more clear, https://github.com/open-rmf/rmf-web/pull/909/commits/d1267670d119315eecfd86fed18ecb5b81b449c4

packages/dashboard/src/components/tasks/tasks-app.tsx
```diff
@@ -309,15 +309,19 @@ export const TasksApp = React.memo(
...
+        // FIXME: Task requests are currently required for parsing pickup and
+        // destination information. Once we start using TaskState.Booking.Labels
+        // to encode these information, we can skip querying for task requests.
+        const pastMonthTaskRequests = await getPastMonthTaskRequests(pastMonthTasks);
+        if (!pastMonthTasks || !pastMonthTasks.length) {
           return;
         }
         if (minimal) {
-          downloadCsvMinimal(now, allTasks, allTaskRequests);
+          downloadCsvMinimal(now, pastMonthTasks, pastMonthTaskRequests);
```
koonpeng, line 322: 1. "download*" is misleading since it doesn't actually download, suggest to rename it to "save*" or add docs.
2. It's weird that "minimal" contains info that is not in "full" (from what I see, pickup and dropoff), I will say to include the task request in `full` so it has the same data.
3. Parsing the task request for pick and dropoff is also a red flag, it is out of scope of this PR so just add a `FIXME`. In future PR we can look at using task tags to store this information and just dump it, it also solves the issue of requiring double query to get all the required data.

koonpeng (changes requested):

packages/dashboard/src/components/tasks/tasks-app.tsx
aaronchongth: 1. Changing it to `export`, https://github.com/open-rmf/rmf-web/pull/909/commits/651562d7f1db787161725b79a9198f5cec0cd4e5

2. Technically, `full` contains all the information of `minimal` and much more, since any number of  pickups and dropoffs can be parsed from the task phases if needed. And since we will be adding pick and dropoff information in labels soon, I will not make any changes to this. 

3. FIXME was added previously https://github.com/open-rmf/rmf-web/pull/909/commits/d1267670d119315eecfd86fed18ecb5b81b449c4

koonpeng: > Technically, full contains all the information of minimal and much more, since any number of pickups and dropoffs can be parsed from the task phases if needed. And since we will be adding pick and dropoff information in labels soon, I will not make any changes to this.

Then we don't need to fetch task requests for "minimal"? Or do you mean that it will be included after we need the tags? If so then add a `TODO`.

koonpeng (approved):
> Technically, full contains all the information of minimal and much more, since any number of pickups and dropoffs can be parsed from the task phases if needed. And since we will be adding pick and dropoff information in labels soon, I will not make any changes to this.

Then we don't need to fetch task requests for "minimal"? Or do you mean that it will be included after we add the tags? If so then add a TODO.

aaronchongth:
> > Technically, full contains all the information of minimal and much more, since any number of pickups and dropoffs can be parsed from the task phases if needed. And since we will be adding pick and dropoff information in labels soon, I will not make any changes to this.
> 
> Then we don't need to fetch task requests for "minimal"? Or do you mean that it will be included after we add the tags? If so then add a TODO.

The ticket detailing this is already in our deployment project board, will not be adding additional TODOs here.

PR #910, opened 2024-03-05
Hammer/update scheduler

aaronchongth:
## What's new

* Updated version for `react-scheduler` and its dependencies to fix overlapping scheduled events and how they are organized
* Cleaned up some of the dark/light theme stuff that is no longer used, but conflicts with the updated `mui`
* Fix schedule calendar view jumping to wrong day/week/month after a schedule reload is triggered
* update `caniuse-lite` too to dismiss browserslist warning

After:
image

Before:
image

## Self-checks


packages/dashboard/src/components/app-base.tsx
```diff
@@ -1,16 +1,7 @@
...
-  createTheme,
-  CssBaseline,
-  GlobalStyles,
-  Grid,
-  Snackbar,
-} from '@mui/material';
+import { Alert, AlertProps, createTheme, CssBaseline, Grid, Snackbar } from '@mui/material';
 import { ThemeProvider } from '@mui/material/styles';
 import React from 'react';
-import { rmfDark, rmfDarkLeaflet, rmfLight } from 'react-components';
```
koonpeng, line 12: Remove the themes in `react-components` as well.

I know the themes have many issues both design and implementation wise, but dark mode was one of the requested feature as well.

koonpeng (changes requested):

packages/dashboard/src/components/app-base.tsx
aaronchongth: Gotcha, it was probably introduced before I was aware of it. I'll see if I can just retain it and work around the new Theme props and API

packages/dashboard/src/components/app-base.tsx
```diff
@@ -73,7 +65,6 @@ export function AppBase({ children }: React.PropsWithChildren<{}>): JSX.Element
   return (
     <ThemeProvider theme={theme}>
       <CssBaseline />
-      {settings.themeMode === ThemeMode.RmfDark && <GlobalStyles styles={rmfDarkLeaflet} />}
```
koonpeng, line 76: This PR still keeps the themes now right? This should probably be reverted even if we are not exposing it in the ui.

koonpeng (approved):

packages/dashboard/src/components/app-base.tsx
aaronchongth: Good point, it'll be cleaned up when we clean up leaflet stuff next time, https://github.com/open-rmf/rmf-web/pull/910/commits/3b9f7279f9cbe0b08da4d93e354ed3eccb2f69a4

PR #912, opened 2024-03-12
Use task request labels to encode information required by dashboard

aaronchongth:
## What's new

Requires https://github.com/open-rmf/rmf_ros2/pull/328 and https://github.com/open-rmf/rmf-web/pull/903

This PR adds `TaskBookingLabel` to be used by frontend dashboards to encode additional information required for rendering.

* Encode json into task booking label for
  * task name (not to be confused with task request category, this is an internal encoding of what task "type" it is). The plan is to encode the default name of the task, or the remapped name (supported in https://github.com/open-rmf/rmf-web/pull/925), and the dashboard will render it accordingly, otherwise it will be considered an unknown task and revert to custom
  * warning time (to be populated and used later on, `None` for now).
  * pickup
  * destination
  * cart ID
* Clean up workaround of saving pickup and destination when task request is saved
* Clean up workaround where task requests were queried when tasks were exported
* Adding `task_definition_id` field to `TaskFavorite`,
* Script for migrating past database to encode labels (`TaskState`, `ScheduledTask`)
* Refactored task state update/create to only lock mutex when needed

## Migration (postgres)

Backup database, enter database
```
pg_dump -U USERNAME -W DBNAME > /tmp/DUMPNAME.sql
```

Handling additional field in `TaskFavorite`,
```
# enter db
psql -U USERNAME -d DATABASENAME

# add column
ALTER TABLE "taskfavorite" ADD COLUMN task_definition_id VARCHAR(255)
```

Migrate the rest of the models
```
cd packages/api-server
# modify psql_local_config.py to point to the database

source ../../.venv/bin/activate
source RMF_WORKSPACE/install/setup.bash
export RMF_API_SERVER_CONFIG=/PATH_TO_RMF_WEB/packages/api-server/psql_local_config.py

python3 migrate_db_912.py
```

## Self-checks


packages/react-components/lib/tasks/utils.ts
```diff
@@ -20,6 +20,32 @@ export function taskTypeToStr(taskType: number): string {
...
+  destination?: string;
+  cart_id?: string;
+}
+
+export function parseTaskRequestLabel(taskState: TaskState): TaskRequestLabel | null {
+  let requestLabel: TaskRequestLabel | null = null;
+  if (taskState.booking.labels) {
+    for (const label of taskState.booking.labels) {
+      try {
+        const parsedLabel: TaskRequestLabel = JSON.parse(label);
```
koonpeng, line 36: This does not validate the json, technically the correct return value is `unknown`. We should at least check that

1. it is an object
2. all of the fields have the correct type

koonpeng (changes requested):

koonpeng (changes requested):
The api-server should eventually treat the labels as black boxes, i.e. there should be no models for them. For that to happen, we need to support querying and sorting the labels in a generic way.

I can work on it after this is merged, but in order to avoid more migration hell, we should prepare the labels in a new table in this PR.

packages/api-server/api_server/repositories/tasks.py
```diff
@@ -181,6 +86,14 @@ async def save_task_state(self, task_state: TaskState) -> None:
...
             if task_state.booking.requester
             else None,
+            "pickup": booking_label.description.pickup
+            if booking_label is not None
+            and booking_label.description.pickup is not None
+            else None,
+            "destination": booking_label.description.destination
+            if booking_label is not None
+            and booking_label.description.destination is not None
+            else None,
```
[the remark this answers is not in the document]
aaronchongth: As per discussion from DMs, https://github.com/open-rmf/rmf-web/pull/912/commits/82422f4dc51a7a443965ec3acd7914d556b434a1 creates the generic label model, while we use the reverse relations for sorting and filtering purposes.

packages/api-server/api_server/models/tortoise_models/tasks.py
```diff
@@ -29,6 +30,14 @@ class TaskState(Model):
...
     pickup = CharField(255, null=True, index=True)
     destination = CharField(255, null=True, index=True)
+    labels = ReverseRelation["TaskLabel"]
+
+
+class TaskLabel(Model):
+    state = ForeignKeyField("models.TaskState", null=True, related_name="labels")
+    label_name = CharField(255, null=False, index=True)
+    label_value_str = CharField(255, null=True, index=True)
+    label_value_num = BigIntField(null=True, index=True)
```
koonpeng, line 40: Since the value come from json, technically we need both double and int. Or just double could be good enough if we don't work with values > max safe double (~48 bits iirc).

packages/api-server/api_server/repositories/tasks.py
```diff
@@ -119,6 +97,43 @@ async def save_task_state(self, task_state: TaskState) -> None:
             self.logger.error(
                 f"Failed to save task state of id [{task_state.booking.id}] [{e}]"
             )
+            return
+
+        if not created:
+            return
```
koonpeng, line 103: iiuc, task labels is created when we create a task and never changes in task state updates, could we move label processing to when we dispatch tasks instead?

koonpeng (changes requested):

packages/api-server/api_server/repositories/tasks.py
aaronchongth: no, when we dispatch a task we will not know the task ID to save these labels to, only when the dispatcher replies with the first task state for saving, will we be able to save the labels

packages/api-server/api_server/models/tortoise_models/tasks.py
```diff
@@ -30,6 +30,14 @@ class TaskState(Model):
     unix_millis_warn_time = DatetimeField(null=True, index=True)
     pickup = CharField(255, null=True, index=True)
     destination = CharField(255, null=True, index=True)
+    labels = ReverseRelation["TaskLabel"]
+
+
+class TaskLabel(Model):
+    state = ForeignKeyField("models.TaskState", null=True, related_name="labels")
```
koonpeng, line 37: Because we do not have a canonical "Task" object, this foreign key is causing issues with task requests and task states. I would say better to remove the foreign key constraint and just use task booking id.

koonpeng (changes requested):

packages/api-server/api_server/models/tortoise_models/tasks.py
aaronchongth: Resolved as discussed

packages/api-server/api_server/models/tortoise_models/tasks.py
aaronchongth: https://github.com/open-rmf/rmf-web/pull/912/commits/7c2d8efa27ad3edacc438dec24e756bda3c2baa9

packages/api-server/api_server/repositories/tasks.py
aaronchongth: Added more comments https://github.com/open-rmf/rmf-web/pull/912/commits/7c2d8efa27ad3edacc438dec24e756bda3c2baa9

koonpeng (approved):

PR #913, opened 2024-03-13
api-server: update dependencies

koonpeng:
## What's new

uvicorn changed the behavior of handling `root_path`, after we updated fastapi, it assumes the new uvicorn behavior which causes things to fail when behind a proxy.

What started out as a update to uvicorn turns into updating most dependencies in api_server as new uvicorn broke tortoise-orm, and new tortoise-orm broke pydantic etc.

## Self-checks


koonpeng:
@aaronchongth pnpm has just dropped support for node16, we should get this merged asap.

aaronchongth (changes requested):
Can you modify the README to set up with node 20? I'll test the changes so we can move forward for `main` while we work on the upcoming new `main` branch

aaronchongth (approved):

PR #914, opened 2024-03-13
Hammer/log user actions

aaronchongth:
## What's new

* Add username in task cancellation request label
* Log task cancellation request
* Log delivery alert responses and the user who responded

## Self-checks

### DIFF
--- a/packages/api-server/api_server/routes/delivery_alerts.py
+++ b/packages/api-server/api_server/routes/delivery_alerts.py
@@ -1,7 +1,11 @@
+from fastapi import Depends
 from rx import operators as rxops
 
+from api_server.authenticator import user_dep
 from api_server.fast_io import FastIORouter, SubscriptionRequest
 from api_server.gateway import rmf_gateway
+from api_server.logger import logger
+from api_server.models import User
 from api_server.models.delivery_alerts import (
     DeliveryAlert,
     action_to_msg,
@@ -26,6 +30,7 @@ async def respond_to_delivery_alert(
     task_id: str,
     action: DeliveryAlert.Action,
     message: str,
+    user: User = Depends(user_dep),
 ):
     delivery_alert = DeliveryAlert(
         id=delivery_alert_id,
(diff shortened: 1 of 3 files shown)


packages/api-server/api_server/routes/delivery_alerts.py
```diff
@@ -35,6 +40,8 @@ async def respond_to_delivery_alert(
         task_id=task_id,
         message=message,
     )
+    logger.info(f"Delivery alert responded by {user.username}")
+    logger.info(delivery_alert)
```
koonpeng, line 44: In our case it doesn't matter since we only run a single thread (for now), but in general it is better either
1. Put both into a single log so it can be easily identified.
2. Include the request id so it can be traced, that is if we/fastapi have a system to uniquely tag each request.

packages/api-server/api_server/routes/tasks/tasks.py
```diff
@@ -186,6 +186,7 @@ async def post_activity_discovery(
 async def post_cancel_task(
     request: mdl.CancelTaskRequest = Body(...),
 ):
+    logger.info(request)
```
koonpeng, line 189: This log can be very messy/noisy as we can't identify the source, although in this case I don't think we are using `CancelTaskRequest` any where else so it is technically traceble. For now should be ok but we should have a better logging framework in the future.

koonpeng (approved):
Not needed for this PR but I realize we should add a better logging system. We should add a logger `Depends` which provides a logger formatted with

1. request id
2. client info, check if there is a way to sync the request id with the server, aka uvicorn. If so and if uvicorn already logs the client info then we don't need this.
3. python code file and line number

packages/api-server/api_server/routes/tasks/tasks.py
aaronchongth: Yup, for sure, we will migrate to use the logging server or anything equivalent later on. This is just a project requirement for now, that we should be able to tell who cancelled tasks.

packages/api-server/api_server/routes/delivery_alerts.py
aaronchongth: The task ID is included in the delivery alert model

PR #925, opened 2024-03-28
Hammer/demo tasks

aaronchongth:
## What's new

* Introduced `patrol`, `delivery` and `clean` back
* Refactored each task to it's own file to handle its own description and form
* Configurable list of supported tasks, order of task is respected in the menu
* Task name remapping
* Using task definition ID to identify task type
* Allow configuring of request category via task definition

Here is how the `rmf-web/packages/dashboard/src/assets/resources/main.json` can look like,

```
{
    ...,
    "supportedTasks": [
        {
            "taskDefinitionId": "delivery_pickup",
            "taskDisplayName": "Delivery - 1:1",
            "requestCategory": "compose"
        },
        {
            "taskDefinitionId": "delivery_sequential_lot_pickup",
            "taskDisplayName": "Delivery - Sequential lot pick up",
            "requestCategory": "compose"
        },
        {
            "taskDefinitionId": "delivery_area_pickup",
            "taskDisplayName": "Delivery - Area pick up",
            "requestCategory": "compose"
        },
        {
            "taskDefinitionId": "patrol",
            "taskDisplayName": "Patrol",
            "requestCategory": "patrol"
        },
        {
            "taskDefinitionId": "custom_compose",
            "taskDisplayName": "Custom Compose Task",
            "requestCategory": "compose"
        }
    ]
    ...
}
```

image

image

## Self-checks


packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -976,159 +164,47 @@ function FavoriteTask({
...
+    case 'clean':
+      return makeDefaultCleanTaskDescription();
+    default:
+      return undefined;
+  }
 }
 
-function defaultTaskDescription(taskCategory: string): TaskDescription | undefined {
-  switch (taskCategory) {
+function taskRequestCategory(taskName: string): string | undefined {
```
koonpeng, line 185: If we use the minimal "task definition", we would no longer need to do this hardcoded mapping.

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -1581,6 +649,69 @@ export function CreateTaskForm({
...
+      case 'delivery_area_pickup':
+        return (
+          <MenuItem value={taskName} disabled={Object.keys(dropoffPoints).length === 0}>
+            {taskNameRemap[taskName] ?? 'Delivery - Area pick up'}
+          </MenuItem>
+        );
+      default:
+        return undefined;
+    }
+  };
```
koonpeng, line 713: Do we really need different behavior for different category? Can we just use the same rendering for all task types?

koonpeng (changes requested):

packages/react-components/lib/tasks/types/clean.tsx
```diff
@@ -0,0 +1,123 @@
...
+    cleanPhase: {
+      activity: {
+        category: string;
+        description: {
+          activities: [goToPlaceActivity: GoToPlaceActivity, cleanActivity: CleanActivity];
+        };
+      };
+    },
+  ];
+}
```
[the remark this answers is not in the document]
aaronchongth: That is outdated now, because we are now using `perform_action` to do a cleaning action, https://github.com/open-rmf/rmf_demos/blob/main/rmf_demos_tasks/rmf_demos_tasks/dispatch_clean.py

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -1353,28 +433,51 @@ export function CreateTaskForm({
...
+          <CleanTaskForm
+            taskDesc={taskRequest.description as CleanTaskDescription}
+            cleaningZones={cleaningZones}
+            onChange={(desc: CleanTaskDescription) => {
+              desc.category = taskRequest.description.category;
+              handleTaskDescriptionChange('compose', desc);
+            }}
+            allowSubmit={allowSubmit}
+          />
+        );
```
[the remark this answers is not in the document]
aaronchongth: That is outdated now, because we are now using perform_action to do a cleaning action, https://github.com/open-rmf/rmf_demos/blob/main/rmf_demos_tasks/rmf_demos_tasks/dispatch_clean.py

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -1217,6 +293,8 @@ export interface CreateTaskFormProps
    * Shows extra UI elements suitable for submittng batched tasks. Default to 'false'.
    */
   user: string;
+  supportedTasks?: string[];
+  taskNameRemap?: Record<string, string>;
```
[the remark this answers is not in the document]
aaronchongth: Using a basic definition with `name` and `remapName`, https://github.com/open-rmf/rmf-web/pull/925/commits/eb9aedae37e39d2e764ccd1a840440aa1aa8c8b1

added a comment saying its temporary until we start using proper schemas

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -1285,7 +365,7 @@ export function CreateTaskForm({
 
   const [taskType, setTaskType] = React.useState<string | undefined>(undefined);
   const [taskRequest, setTaskRequest] = React.useState<TaskRequest>(
-    () => requestTask ?? defaultTask(),
+    () => requestTask ?? defaultTaskRequest(supportedTasks[0]),
```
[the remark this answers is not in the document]
aaronchongth: https://github.com/open-rmf/rmf-web/pull/925/commits/eb9aedae37e39d2e764ccd1a840440aa1aa8c8b1

packages/react-components/lib/tasks/types/utils.ts
```diff
@@ -0,0 +1,102 @@
...
+  return value.length > 0;
+}
+
+export function isPositiveNumber(value: number): boolean {
+  return value > 0;
+}
+
+export function getShortDescription(
+  taskRequest: TaskRequest,
+  taskNameRemap?: Record<string, string>,
```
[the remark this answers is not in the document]
aaronchongth: https://github.com/open-rmf/rmf-web/pull/925/commits/eb9aedae37e39d2e764ccd1a840440aa1aa8c8b1

packages/react-components/lib/tasks/create-task.tsx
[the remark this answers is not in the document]
aaronchongth: I would actually leave out top-level-category and compose-description-category until we start working on the json schema
as those are intricately tied to the task descriptions which are defined quite flexibly, so the only source of truth should be the implementation for now

packages/react-components/lib/tasks/create-task.tsx
aaronchongth: this mapping should be left in, as it is currently the best source of truth side-by-side the implementation, otherwise we will have to perform a lot more additional error checking and validation on the build-time configuration to ensure we are allowing valid tasks.

this mapping will be removed once we shift to the json schema based task definition

packages/react-components/lib/tasks/create-task.tsx
aaronchongth: Yes we do, there were instances where users were confused why certain tasks did not allow any options (no cleaning zones in office world, for example), by disabling menu items whenever these tasks are invalid (according to the implementation), we can direct user's actions and prevent mistakes

packages/react-components/lib/tasks/types/clean.tsx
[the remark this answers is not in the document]
koonpeng: You mean the schemas are outdated? Can you open a PR to fix it in rmf?

packages/react-components/lib/tasks/create-task.tsx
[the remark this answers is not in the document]
koonpeng: > I would actually leave out top-level-category and compose-description-category until we start working on the json schema
as those are intricately tied to the task descriptions which are defined quite flexibly, so the only source of truth should be the implementation for now

Right now the "name" is being used as the top level category which makes things more confusing. The compose description category is tied to the task description which is why we also need to define it in the task definition.

packages/react-components/lib/tasks/create-task.tsx
koonpeng: We could just remove the invalid tasks from the `supportedTasks` if we know a deployment does not a certain task. Disabling the menu items would cause a different set of ux issues as well, if we expect a deployment to support cleaning task, and if there is no valid cleaning zone for some reason, users will be confused why cleaning tasks are disabled even though it is supported. Not to mention the way it is now, with different item having different styling (font size) is also bad for ui.

packages/react-components/lib/tasks/types/clean.tsx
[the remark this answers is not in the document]
aaronchongth: No the schemas are not outdated, just it is no longer used in rmf_demos. I believe there were past deployments that were using the schemas, so we should not be changing anything on that end

packages/react-components/lib/tasks/create-task.tsx
aaronchongth: I beg to differ, it will still be the responsibility of the developer to keep in sync their building map yaml to contain the proper waypoints and properties. In the event that their deployment is meant to support a `Clean` task for example, after the building map has been changed, it should show up but not allow them to run it (because the waypoints may have been changed). This is a sign for the developer to fix what is needed

Users won't care or be touching the dashboard config or the building map yaml at all

aaronchongth: We can add tooltips to disabled menu items to let them know why certain menu items are disabled.

> Not to mention the way it is now, with different item having different styling (font size) is also bad for ui.

This is only temporary until we move to json form. I would rather have bad looking UI for now, than things not working

packages/dashboard/src/managers/resource-manager.ts
```diff
@@ -74,6 +77,11 @@ export default class ResourceManager {
     this.cartIds = resources.cartIds || [];
     this.loggedInDisplayLevel = resources.loggedInDisplayLevel;
     this.emergencyLots = resources.emergencyLots || [];
+    this.supportedTasks = resources.supportedTasks || [
+      { name: 'patrol', nameRemap: undefined },
+      { name: 'delivery', nameRemap: undefined },
+      { name: 'clean', nameRemap: undefined },
```
koonpeng, line 83: How about instead of `name` and `nameRemap`, we use `name` and `type`. Essentially we rename "name" -> "type" and "nameRemap" -> "name". Technically, this is only a semantic change but this change better describe how "name" is treated in the code. In terms of design, it is better if task names (display name) do not have to be unique.

packages/react-components/lib/tasks/types/clean.tsx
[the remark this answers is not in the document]
koonpeng: iiuc, we have a native "clean" task and a custom "clean" task that is actually a compose task that includes the native "clean"? This presents a problem with the way we define `supportedTasks`. The name `clean` now refers to the compose-clean, and since the "name" is also used to determine the category for non-compose task, we end up with no way to define both a native clean and compose-clean task.

This would be solved if we use "name" and "type" to define `supportedTasks`, the compose-clean task could be defined as `{ "name": "clean", "type": "composeClean" }`.

koonpeng (changes requested):

packages/react-components/lib/tasks/create-task.tsx
koonpeng: > I beg to differ, it will still be the responsibility of the developer to keep in sync their building map yaml to contain the proper waypoints and properties. In the event that their deployment is meant to support a Clean task for example, after the building map has been changed, it should show up but not allow them to run it (because the waypoints may have been changed). This is a sign for the developer to fix what is needed

I don't understand the example, if the clean task shows up but cannot be ran, it means there is no valid waypoints. Even if we leave the menu item enabled, there would be no valid waypoints so the user still cannot submit the task.

> This is only temporary until we move to json form. I would rather have bad looking UI for now, than things not working

I would say this is more important because we are moving towards using ui schemas. We would not be able to define custom logic to enable/disable a category with the schemas, this particular menu is also selecting the "task type" or the schema to use and is outside the scope of the schemas. Because we are going to use ui schemas, this needs to be a list where we treat every menu item the same.

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -1851,7 +914,9 @@ export function CreateTaskForm({
                       setOpenFavoriteDialog(true);
                     }}
                     style={{ marginTop: theme.spacing(2), marginBottom: theme.spacing(2) }}
-                    disabled={taskDefinitionId === 'custom_compose'}
+                    disabled={
+                      taskDefinitionId === DefaultCustomComposeTaskDefinition.taskDefinitionId
```
koonpeng, line 918: Document why custom compose is not allowed as a favorite task.

packages/dashboard/src/managers/resource-manager.ts
```diff
@@ -71,6 +80,12 @@ export default class ResourceManager {
     this.attributionPrefix = resources.attributionPrefix || 'OSRC-SG';
     this.cartIds = resources.cartIds || [];
     this.loggedInDisplayLevel = resources.loggedInDisplayLevel;
+    this.supportedTasks = resources.supportedTasks || [
+      DefaultPatrolTaskDefinition,
+      DefaultDeliveryTaskDefinition,
+      DefaultComposeCleanTaskDefinition,
+      DefaultCustomComposeTaskDefinition,
+    ];
```
koonpeng, line 88: iirc, we cannot support a fully new task definitoin yet, we only support task definition ids which is hardcoded in create task form. Either validate that the `supportedTasks` have a valid task definition ids or make the TaskDefinition typing more strict, e.g. `taskDefinitiodId: 'foo' | 'bar'`.

packages/react-components/lib/tasks/types/compose-clean.tsx
```diff
@@ -0,0 +1,155 @@
+import { Autocomplete, TextField } from '@mui/material';
+import React from 'react';
+import type { TaskBookingLabel } from 'api-client';
+import { TaskDefinition } from '../create-task';
+
+export const DefaultComposeCleanTaskDefinition: TaskDefinition = {
+  taskDefinitionId: 'compose-clean',
+  taskDisplayName: 'Clean',
+  requestCategory: 'compose',
+};
```
koonpeng, line 10: 1. I think these default definitions should be moved closer to resource manager.
2. The resource manager allows configuring the entire `TaskDefinitoin`, including `requestCategeory`, but the default request category is always used in `CreateTaskForm`. It should respect the configured category, even if it is invalid or not allow configuring it.

packages/dashboard/src/components/tasks/task-schedule-utils.ts
```diff
@@ -166,8 +166,20 @@ export const apiScheduleToSchedule = (scheduleTask: ApiSchedule[]): Schedule =>
...
+export const getScheduledTaskTitle = (
+  task: ScheduledTask,
+  supportedTasks?: TaskDefinition[],
+): string => {
+  let remappedTaskName: string | undefined = undefined;
+  if (supportedTasks) {
+    for (const s of supportedTasks) {
+      if (s.taskDefinitionId === task.task_request.category) {
+        remappedTaskName = s.taskDisplayName;
+      }
```
koonpeng, line 178: I thought `task.task_request.category` is the top level rmf category? Shouldn't we check the task definition id in the booking label?

koonpeng (changes requested):

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -1058,159 +194,13 @@ function FavoriteTask({
...
-      return undefined;
-  }
-}
-
-function defaultTask(): TaskRequest {
-  return {
-    category: 'compose',
-    description: defaultDeliveryTaskDescription(),
+    category: category ?? 'compose',
+    description: description ?? makeDefaultDeliveryTaskDescription(),
```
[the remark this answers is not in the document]
aaronchongth: https://github.com/open-rmf/rmf-web/pull/925/commits/bce846ff06b82b85bbc727e4fbdf912a9135103d

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -1500,34 +522,63 @@ export function CreateTaskForm({
...
+      case DefaultCustomComposeTaskDefinition.taskDefinitionId:
       default:
-        return null;
+        return (
+          <CustomComposeTaskForm
+            taskDesc={taskRequest.description as CustomComposeTaskDescription}
+            onChange={(desc) => {
+              handleCustomComposeTaskDescriptionChange(desc);
+            }}
+            allowSubmit={allowSubmit}
```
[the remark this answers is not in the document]
aaronchongth: https://github.com/open-rmf/rmf-web/pull/925/commits/bce846ff06b82b85bbc727e4fbdf912a9135103d

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -1364,11 +361,20 @@ export function CreateTaskForm({
...
+        supportedTasks.length > 0 ? supportedTasks[0].taskDefinitionId : DefaultTaskDefinitionId,
+      ),
+  );
+  const initialBookingLabel = requestTask ? getTaskBookingLabelFromTaskRequest(requestTask) : null;
+  const [taskDefinitionId, setTaskDefinitionId] = React.useState<string>(
+    initialBookingLabel && initialBookingLabel.description.task_definition_id
+      ? (initialBookingLabel.description.task_definition_id as string)
+      : supportedTasks.length > 0
+      ? supportedTasks[0].taskDefinitionId
+      : DefaultTaskDefinitionId,
```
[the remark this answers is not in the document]
aaronchongth: https://github.com/open-rmf/rmf-web/pull/925/commits/bce846ff06b82b85bbc727e4fbdf912a9135103d

packages/dashboard/src/components/tasks/task-schedule-utils.ts
aaronchongth: thanks for catching that, https://github.com/open-rmf/rmf-web/pull/925/commits/d452affaa133634466a30fb774f3af44a2ca417a

packages/react-components/lib/tasks/create-task.tsx
aaronchongth: There is already a FIXME in the code

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -1364,11 +361,20 @@ export function CreateTaskForm({
...
+    () =>
+      requestTask ??
+      defaultTaskRequest(
+        supportedTasks.length > 0 ? supportedTasks[0].taskDefinitionId : DefaultTaskDefinitionId,
+      ),
+  );
+  const initialBookingLabel = requestTask ? getTaskBookingLabelFromTaskRequest(requestTask) : null;
+  const [taskDefinitionId, setTaskDefinitionId] = React.useState<string>(
+    initialBookingLabel && initialBookingLabel.description.task_definition_id
+      ? (initialBookingLabel.description.task_definition_id as string)
```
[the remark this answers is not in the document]
aaronchongth: https://github.com/open-rmf/rmf-web/pull/925/commits/d452affaa133634466a30fb774f3af44a2ca417a

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -1677,7 +734,11 @@ export function CreateTaskForm({
       onSuccessFavoriteTask &&
         onSuccessFavoriteTask('Deleted favorite task successfully', favoriteTaskBuffer);
 
-      setTaskRequest(defaultTask());
+      setTaskRequest(
+        supportedTasks && supportedTasks.length > 0
+          ? defaultTaskRequest(supportedTasks[0].taskDefinitionId)
+          : defaultTaskRequest('patrol'),
```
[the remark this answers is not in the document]
aaronchongth: Per discussion, will refactor to allow the dashboard to fail without supported tasks as it is a misconfiguration

packages/react-components/lib/tasks/types/compose-clean.tsx
aaronchongth: Split out "configuration" from "definition", where the configuration stays nearer to the resource manager, https://github.com/open-rmf/rmf-web/pull/925/commits/7f7eb6203121b1f1631da5978761d2b86a2eb4ec

this allows us to modify only the modifiable fields (display name for now)

packages/dashboard/src/managers/resource-manager.ts
aaronchongth: the resource manager side will do the checking and only allow supported tasks,
https://github.com/open-rmf/rmf-web/pull/925/commits/7f7eb6203121b1f1631da5978761d2b86a2eb4ec

packages/dashboard/src/managers/resource-manager-tasks.ts
```diff
@@ -1,14 +1,29 @@
...
+      const definition = getTaskDefinition(taskResource.task_definition_id);
+      if (!definition) {
+        console.error(
+          `Invalid tasks configured for dashboard: [${taskResource.task_definition_id}]`,
+        );
+        continue;
       }
+
+      if (taskResource.display_name !== undefined) {
+        definition.taskDisplayName = taskResource.display_name;
```
koonpeng, line 23: Modifying global const is usually is a code smell. Have a function to get the display name instead.

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -355,26 +351,68 @@ export function CreateTaskForm({
...
   const isScreenHeightLessThan800 = useMediaQuery('(max-height:800px)');
 
-  const [favoriteTaskBuffer, setFavoriteTaskBuffer] = React.useState<TaskFavorite>(
-    defaultFavoriteTask(),
-  );
+  // Note that we are not checking if the number of supported tasks is larger
+  // than 0, this will cause the dashboard to fail when the create task form is
+  // opened. This is intentional as it is a misconfiguration and will require
+  // the build-time configuration to be fixed.
+  let allTasksToDisplayAreValid = true;
```
koonpeng, line 358: unused

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -355,26 +351,68 @@ export function CreateTaskForm({
...
+  const validTasks: TaskDefinition[] = [];
+  let defaultTaskDescription: string | TaskDescription | null = null;
+  let defaultTaskRequest: TaskRequest | null = null;
+  tasksToDisplay.forEach((supportedTask: TaskDefinition, index: number) => {
```
koonpeng, line 362: Should memo this

packages/react-components/lib/tasks/create-task.tsx
```diff
@@ -355,26 +351,68 @@ export function CreateTaskForm({
...
+      }
+    }
+  });
+
+  if (!defaultTaskDescription || !defaultTaskRequest) {
+    // We should never reach this state unless a misconfiguration happened.
+    const err = Error('Default task could not be generated, this might be a configuration error');
+    onFail && onFail(err, []);
+    console.error(err.message);
+    throw new TypeError(err.message);
```
koonpeng, line 390: This is technically not a type error, just throw the `err` created earlier.

koonpeng (approved):

packages/react-components/lib/tasks/create-task.tsx
aaronchongth: https://github.com/open-rmf/rmf-web/pull/925/commits/f2c0b08afc76c006185537d6fc4e44f12c441cb7

packages/react-components/lib/tasks/create-task.tsx
aaronchongth: https://github.com/open-rmf/rmf-web/pull/925/commits/ba6be3062714389593027cdac565d0e593027c1b

packages/react-components/lib/tasks/create-task.tsx
aaronchongth: https://github.com/open-rmf/rmf-web/pull/925/commits/ba6be3062714389593027cdac565d0e593027c1b

packages/dashboard/src/managers/resource-manager-tasks.ts
aaronchongth: refactored it in https://github.com/open-rmf/rmf-web/pull/925/commits/ba6be3062714389593027cdac565d0e593027c1b, still not using a function as there will be 2 sources of truth, I would rather the resource manager handles all the configuration and massaging of definitions.

PR #928, opened 2024-04-03
Hammer/save state race condition

aaronchongth:
## What's new

* Prevent spamming of alerts when fleet adapter stalls and repeatedly sends completed or failed tasks state updates
* Adding a mutex for saving task state, to prevent writing race-condition in DB. (Note: I don't have logs for this, and I'm unsure if this only happens with postgres, but the issue comes from saving the task state from `ApiResponse` and websocket response from the task dispatcher at nearly the same time)

## Self-checks


packages/api-server/api_server/repositories/tasks.py
```diff
@@ -1,5 +1,6 @@
 import sys
 from datetime import datetime
+from multiprocessing import Lock
```
koonpeng, line 3: Using normal lock will likely cause a deadlock, need to use asyncio lock instead.

packages/api-server/api_server/repositories/tasks.py
```diff
@@ -162,32 +164,39 @@ async def query_task_requests(self, task_ids: List[str]) -> List[DbTaskRequest]:
...
-            "unix_millis_request_time": task_state.booking.unix_millis_request_time
-            and datetime.fromtimestamp(
-                task_state.booking.unix_millis_request_time / 1000
-            ),
-            "requester": task_state.booking.requester
-            if task_state.booking.requester
-            else None,
-        }
+        # FIXME: this may be fixed upstream in DB or ORM, this mutex can be
+        # removed once these libraries have been updated and tested to be fixed
```
koonpeng, line 168: Can you add/document a manual test (due to it's nature, we probably don't want to run the test in ci) that can reproduce this?

packages/api-server/api_server/routes/internal.py
```diff
@@ -93,13 +98,28 @@ async def process_msg(msg: Dict[str, Any], fleet_repo: FleetRepository) -> None:
...
+                failed_tasks.append(task_state.booking.id)
+
+                # TODO: check if the failure alert content is the same
+                alert = await alert_repo.create_alert(task_state.booking.id, "task")
+                if alert is not None:
+                    alert_events.alerts.on_next(alert)
+            else:
+                logger.warning(
+                    "Received a repeated failed task state update, please check if the fleet adapter has stalled"
+                )
```
koonpeng, line 122: Please document the behavior of the fleet adapter when it stalls.

This workaround can cause problems, there is a chance of a false negative where a "real" alert get filtered because there are not enough alerts that pushes the booking id out of the queue. We need to use additional info to detect the duplicated alert. Time, alert id etc, alert id is ideal if available.

koonpeng (approved):
Pre-approve as I will be on leave.

packages/api-server/api_server/routes/internal.py
aaronchongth: I've reverted it so that we can handle it properly in a subsequent fix, most likely we'll be deprecating this push method, and just have the frontend handle the alerts.

packages/api-server/api_server/repositories/tasks.py
aaronchongth: https://github.com/open-rmf/rmf-web/pull/928/commits/cf37addbef64dcd00d6fd7bf0de7994682e4c651

packages/api-server/api_server/repositories/tasks.py
aaronchongth: https://github.com/open-rmf/rmf-web/pull/928/commits/cf37addbef64dcd00d6fd7bf0de7994682e4c651

PR #930, opened 2024-04-05
Hammer/adapter lift request qos

luca-della-vedova:
## What's new

Ported from `deploy/hammer`, make the lift request qos transient local, having this publisher volatile and the subscription on `rmf_ros2` transient local means communication does not take place and the lift request widget does not actually work. 

## Self-checks

## Discussion

I tested this locally and without the fix the widget doesn't work, with it requests are respected.


aaronchongth (approved):
LGTM

PR #940, opened 2024-04-24
resubscribe on reconnect

koonpeng:
## What's new

Fixes things not being updated after a reconnect.

Can be tested by using chrome's debug console, in the network tab, set throttle to "offline", then wait for socketio to timeout, then set it back to no throttle.

## Self-checks


packages/dashboard/src/components/map-app.tsx
```diff
@@ -310,7 +309,15 @@ export const MapApp = styled(
       const sub = rmf.fleetsObs
         .pipe(
           switchMap((fleets) =>
-            merge(...fleets.map((f) => (f.name ? rmf.getFleetStateObs(f.name) : EMPTY))),
+            merge(
+              ...fleets.map((f) =>
+                f.name
+                  ? rmf
+                      .getFleetStateObs(f.name)
+                      .pipe(throttleTime(5000, undefined, { leading: true, trailing: true }))
```
aaronchongth, line 317: Can we have the throttle time to be 1000?
5000 makes the robots rather glitchy and jump around every 5 seconds. The motion of the robot is quite important for remote monitoring so I think 1000 milliseconds is good, while every other subscription we can keep at 5000.

aaronchongth (changes requested):
Thank you! Looks great, I only have 1 comment about the fleet subscription on map app

packages/dashboard/src/components/map-app.tsx
koonpeng: reduced to 500 and remove unused code

koonpeng:
Reduced throttle for other topics to 3s for more responsive ui.

aaronchongth (approved):
Awesome, thanks!

PR #949, opened 2024-05-28
api-server: Support generic labels

koonpeng:
## What's new

Supports labels without requiring the schema to be strongly defined. This is still a bit hacky, it still assumes the label is a json with a `description` field while labels in RMF are free-form strings. RMF needs to support key value pair labels for us to be able to use them "natively".

Another issue is that tortoise-orm is lacking many features which prevent us to implement proper sorting on generic labels so we must still use hacks for the `pickup` and `destination` fields. We could use raw sql for this or drop support for sorting by labels.

## Self-checks

### DIFF
--- a/packages/api-server/api_server/rmf_io/test_rmf_service.py
+++ b/packages/api-server/api_server/rmf_io/test_rmf_service.py
@@ -108,17 +108,17 @@ class TestRmfService(unittest.TestCase):
         cls._server_thread.join()
         cls.loop.close()
 
-    def test_call(self):
-        async def run():
-            result = await self.rmf_service.call("hello")
-            self.assertEqual("hello", result)
+    # def test_call(self):
+    #     async def run():
+    #         result = await self.rmf_service.call("hello")
+    #         self.assertEqual("hello", result)
 
-        self.loop.run_until_complete(run())
+    #     self.loop.run_until_complete(run())
 
-    def test_multiple_calls(self):
-        async def run():
-            tasks = [self.rmf_service.call("hello"), self.rmf_service.call("world")]
-            results = await asyncio.gather(*tasks)
-            self.assertListEqual(["hello", "world"], list(results))
+    # def test_multiple_calls(self):
+    #     async def run():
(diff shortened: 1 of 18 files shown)


packages/api-server/api_server/routes/tasks/tasks.py
```diff
@@ -99,10 +99,10 @@ async def query_task_states(
         None, description="comma separated list of requester names"
     ),
     pickup: Optional[str] = Query(
-        None, description="comma separated list of pickup names"
+        None, description="comma separated list of pickup names", deprecated=True
```
aaronchongth, line 102: why is this marked as deprecated? Same for the destination query below

aaronchongth (changes requested):
Looks good, just 1 question about the query route

packages/api-server/api_server/routes/tasks/tasks.py
koonpeng: They are being replaced with `label=name=value`, in `main`. Let me see if I can port generic label filtering over (already tested sorting doesn't work due to tortoise bug).

koonpeng: managed to port label sorting over, this also fixes an issue with filtering where if you try to filter by pickup and destination you get no result even though there is a task that meets the criteria.

aaronchongth (changes requested):
Looks like with the extra fields in the API, anywhere using this API needs to be updated too

packages/api-server/api_server/rmf_io/test_rmf_service.py
```diff
@@ -108,17 +108,17 @@ def tearDownClass(cls) -> None:
         cls._server_thread.join()
         cls.loop.close()
 
-    def test_call(self):
-        async def run():
-            result = await self.rmf_service.call("hello")
-            self.assertEqual("hello", result)
+    # def test_call(self):
```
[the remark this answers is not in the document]
koonpeng: commented them out to make testing faster :sweat_smile:, reverted now.

aaronchongth (approved):
Thanks for setting this up! LGTM

PR #950, opened 2024-06-03
api-server: Support task labels in main

koonpeng:
## What's new

Add support for task labels in main. Unlike in `deploy/hammer`, we only support single-value and key-value pair labels, we no longer support embedding json inside labels.

Bug fixes in tortoise-orm also allows us to support generic label sorting and filtering, now you can filter any labels with `label=name=value,name2=value2` and sort by any labels with `order_by=label=name,-label=name2`.

## Self-checks


packages/api-client/lib/openapi/api.ts
```diff
@@ -9372,6 +9372,7 @@ export const TasksApiAxiosParamCreator = function (configuration?: Configuration
      * @param {string | null} [category] comma separated list of task categories
      * @param {string | null} [assignedTo] comma separated list of assigned robot names
      * @param {string | null} [status] comma separated list of statuses
+     * @param {string | null} [label] comma separated list of labels, each item must be in the form &lt;key&gt;&#x3D;&lt;value&gt;, multiple items will filter tasks with all the labels
```
aaronchongth, line 9375: Looks like this needs to be updated on the dashboard package's usage

aaronchongth (approved):
Thanks! Just the dashboard side needs to be updated, otherwise LGTM

packages/api-client/lib/openapi/api.ts
koonpeng: thanks for the note!

PR #954, opened 2024-06-07
Shows outdated tasks as stale

aaronchongth:
## What's new

When a fleet adapter restart is required during a task execution, the api-server will stop receiving task state updates for the ongoing task and the status will continue to remain `underway` or `queued`. This change hopes to make it clearer that something has interrupted the task and will never be completed/cancelled properly. 

* Shows outdated tasks (fleet adapter restarts, etc) as `stale`
* Tooltip for explanation
* Turning this feature on or off using build-time configuration, `resources.dismissStaleTasks`

image

## Testing in sim

* as sim time is always behind real time, turning this feature will make ongoing tasks always be `stale`
* Here is how the `rmf-web/packages/dashboard/src/assets/resources/main.json` can look like
```
{
    ...,
    "dismissStaleTasks": true,
    ...
}

```

## Self-checks


packages/react-components/lib/tasks/task-table-datagrid.tsx
```diff
@@ -251,8 +276,30 @@ export function TaskDataGridTable({
...
+              title={
+                <React.Fragment>
+                  <Typography>
+                    Finish time is in the past, but task is still queued or underway.
+                  </Typography>
+                  <Typography>
+                    The task may have been interrupted or stalled during the execution.
+                  </Typography>
+                  <Typography>Last status update: {statusString}</Typography>
+                </React.Fragment>
```
koonpeng, line 293: The "stale" state is not an official rmf state? Instead of hiding the last known state in a tooltip, suggest to format it together. e.g. `Underway (stale)`. This prevent a out of place state from showing when sorting/filtering.

koonpeng (changes requested):
The features seems very helpful, is there any reason why we don't just always enable it? Doesn't seem that it will hurt anything.

packages/react-components/lib/tasks/task-table-datagrid.tsx
aaronchongth: Fair enough, those are good points. Per discussion, I will be making this feature enabled by default, and also removing the configuration part.

koonpeng (approved):

PR #955, opened 2024-06-12
support jazzy

koonpeng:
## What's new

Update to noble+jazzy.

* Build a minimal rmf image because the rmf demos image is not yet updated to jazzy, also it's huge.
* Update all workflows to use that minimal image.
* Update references of humble to jazzy.
* Cleanup workflows.
* bootstrap venv + pipenv because pip no longer allows installing to system packages in noble

## Self-checks


aaronchongth (approved):

aaronchongth (changes requested):
While running simulation tests, I noticed that
* the simulation is already running, then only we start the api-server, it will segfault
* after starting the api-server, then starting the simulation, the segfault occurs when a task is dispatched

koonpeng:
By segfault, you mean that rmf crashes? Let me see if I can reproduce it.

koonpeng:
Not able to reproduce it, though I get a different error

```
[fleet_adapter-14] Other error for tinyRobot2 in get_data: 'robot_name'
```

aaronchongth:
Sorry that was cryptic of me, I mean't api-server crashes. Let me attach some screenshots

image

As for the different error, it looks like you may have an outdated workspace, I remember seeing this error before I rebuilt off `main` for everything

koonpeng:
I'm not able to reproduce it, I can run with sim and managed to send a task.

aaronchongth (approved):
Thanks for supporting this! LGTM

Looks like our migration effort of deployment branches have more to keep up with!

PR #957, opened 2024-06-20
api-server: fix postgres label sorting

koonpeng:
## What's new

Tested working on both postgres and sqlite

## Self-checks


aaronchongth (approved):
Awesome! This fixes the issue of using postgres on `main`, while sorting is still working for both sqlite and postgres. However since we are not using any label fields on the frontend for querying/sorting/filtering, we are likely to still have issues on `deploy/hammer` where we are using these fields.

We can merge this as a temporary measure, and apply the changes here to test on `deploy/hammer`

koonpeng:
Tried to port this to `deploy/hammer` but atm it's not possible because the version of fastapi is too old. More specifically, the version of starlette it depends on is too old and breaks with newer version of anyio, it also incorrectly specify the version range of anyio it supports so anything that causes pipenv to update the lock file will break it.

PR #958, opened 2024-06-24
api-server: port postgres fix to deploy/hammer

koonpeng:
## What's new

A partial port of the postgres fix to deploy/hammer. The version of fastapi is too old and has conflict with newer versions of anyio, it also incorrectly specify the version range supported so any change to the dependency list will break it. A hacky port has to be done where the required test dependencies are not updated and is manually installed, this does mean that it is not possible to run the postgres tests in CI.

## Self-checks


aaronchongth (approved):
Thanks for handling the port! I have the changes needed to make the UI work https://github.com/open-rmf/rmf-web/pull/959, once we merge that we are good to go

PR #960, opened 2024-06-24
api-server: first impl of rio

koonpeng:
## What's new

A set of endpoints for RIO (Robot Interaction Objects), a simple system to represent generic "objects" which can be tracked by RMF. For now this is just endpoints to store and query RIO. There is no usage of RIO in RMF yet, this is just an initial data structure to facilitate future integrations.

## Self-checks


aaronchongth (approved):
Thanks for setting this up so quickly! We will need this for the SCAS integration, could you have this ported to `deploy/hammer-humble`(edited) as well? I guess either target it from this PR or port it over

PR #969, opened 2024-07-16
Adding build-time configuration to bring back custom and admin tabs

aaronchongth:
## What's new

* Bring back admin tab and custom tabs
* Make them build-time configurable

## Self-checks

### DIFF
--- a/packages/dashboard/app-config.json
+++ b/packages/dashboard/app-config.json
@@ -33,5 +33,7 @@
       }
     }
   },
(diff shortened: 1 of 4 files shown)


koonpeng (changes requested):
Might need to rework this as #977 removes resource manager. The complex way that resource manager imports stuff does not work well with vite, now it uses a simple json config.

koonpeng (approved):

PR #976, opened 2024-07-25
Ac/remove permission requirements

aaronchongth:
## What's new

* Allow all users to access admin actions until authz is implemented properly
* Allow all users to edit or delete schedule series

### DIFF
--- a/packages/dashboard/src/components/appbar.tsx
+++ b/packages/dashboard/src/components/appbar.tsx
@@ -520,8 +520,7 @@ export const AppBar = React.memo(({ extraToolbarItems }: AppBarProps): React.Rea
                 </MenuItem>
                 <Divider />
                 <MenuItem
-                  // FIXME: reinstate after authz is implemented properly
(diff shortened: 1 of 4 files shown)


packages/dashboard/src/components/appbar.tsx
```diff
@@ -520,7 +520,8 @@ export const AppBar = React.memo(({ extraToolbarItems }: AppBarProps): React.Rea
                 </MenuItem>
                 <Divider />
                 <MenuItem
-                  disabled={!profile.user.is_admin}
+                  // FIXME: reinstate after authz is implemented properly
+                  // disabled={!profile.user.is_admin}
```
koonpeng, line 524: I think we should keep this check, there should be no reason for non admin to see this tab, even if the concept of "admin" has no meaning now.

koonpeng (changes requested):

koonpeng (approved):

PR #979, opened 2024-07-29
Hammer humble/update kc

aaronchongth:
## What's new

Updating `keycloak-js` due to the extremely outdated version we are still using, and that it doesn't work with keycloak version 17 and above due to https://www.keycloak.org/migration/migrating-to-quarkus

Update to
* node 18
* `pnpm` to 9
* `keycloak-js` to 22.0.5
* `canvas` to 2.11.2

requires environment variable `NODE_OPTIONS=--openssl-legacy-provider`

## Self-checks


koonpeng (approved):

packages/rmf-auth/package.json
```diff
@@ -28,7 +28,7 @@
     "@mui/lab": "5.0.0-alpha.86",
     "api-client": "workspace:*",
     "eventemitter3": "^4.0.7",
-    "keycloak-js": "^11.0.2"
+    "keycloak-js": "^22.0.5"
```
koonpeng, line 31: any reason why it is updated to 22 and not the latest?

aaronchongth: if I update it to the latest or anything > 22.0.5, a significant jump in number of subdependencies will need to be updated and things start to break

I will be continuing with more tests and trying again today before merging

aaronchongth: yup anything above 22.0.5 requires ES6 and above, it's probably not the best idea to migrate even this branch to the latest, I'll try to test with the cluster

PR #980, opened 2024-07-31
Bump versions to 0.1.0

aaronchongth:
## What's new

* Bump versions
* Added descriptions where necessary

## Self-checks

### DIFF
--- a/packages/api-client/package.json
+++ b/packages/api-client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "api-client",
-  "version": "1.0.0",
+  "version": "0.1.0",
   "description": "API client library to interface with the API server",
   "main": "dist/lib/index.js",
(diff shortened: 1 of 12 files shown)


koonpeng (commented):
I'm not sure about v1.0, there are still many issues which require large refactors, there is a high chance that many things will change.

aaronchongth:
hmmm, with some of the packages set out as version `0.0.1` in the past, we are currently making backward incompatible changes, so technically it should be a major version increase. Unless we ignore the versions mentioned in the past entirely, and start from a unified `0.0.1`. What do you think? @koonpeng

koonpeng:
Bumping version is fine but I don't think we should go straight to v1, there are too many issues atm to consider it stable. I think we should look at v0.1 atm.

koonpeng (approved):

PR #986, opened 2024-08-02
dashboard: fix and improve build

koonpeng:
Some fixes and improvement to the build script.

1. Fix deps not being built before the dashboard build.
2. Uses a simple and a bit crude way to allow some variables to be configured after the bundle is built, inspired by [import-meta-env](https://import-meta-env.org/). It uses the same method but a bit more crude and specialized, advantage is that we don't need node at runtime.
3. Improvement on how we load some configs so that they can be analyzed and removed by terser (used by vite internally). Particularly keycloak/stub authenticator, all the admin and custom tabs components can be removed from the bundle if they are disabled.
4. Fix some issues when building in humble.


packages/dashboard/vite.config.ts
```diff
@@ -1,11 +1,58 @@
...
+function booleanToString(b: boolean | null | undefined) {
+  return b ? 'true' : 'false';
+}
 
 // https://vitejs.dev/config/
 export default defineConfig({
-  plugins: [react()],
+  define: {
+    APP_CONFIG_AUTH_PROVIDER: `'${appConfig.auth.provider}'`,
+    APP_CONFIG_ENABLE_CUSTOM_TABS: `${booleanToString(appConfig.adminTab)}`,
```
aaronchongth, line 52:
```suggestion
    APP_CONFIG_ENABLE_CUSTOM_TABS: `${booleanToString(appConfig.customTabs)}`,
```

packages/dashboard/vite.config.ts
```diff
@@ -1,11 +1,58 @@
...
+};
+
+function booleanToString(b: boolean | null | undefined) {
+  return b ? 'true' : 'false';
+}
 
 // https://vitejs.dev/config/
 export default defineConfig({
-  plugins: [react()],
+  define: {
```
aaronchongth, line 50: May I know what is different about `APP_CONFIG_AUTH_PROVIDER`, `APP_CONFIG_ENABLE_CUSTOM_TABS`, and `APP_CONFIG_ENABLE_ADMIN_TAB` from the other basic configurations in `app-config.json` like `helpLink`, `reportIssue`, such that they need to be defined here? It doesn't seem like we are using them for search-and-replace after the bundle is built too

aaronchongth (changes requested):
This is awesome! Thanks considering the code differences between staging and production. Just have a question down below

packages/dashboard/vite.config.ts
koonpeng: They need to be preprocessed before terser so that they can be tree shaken. vite will transform them to literals before passing to terser, if we reference them via a variable then terser can't analyze them.

aaronchongth: gotcha thanks for explaining

packages/dashboard/vite.config.ts
koonpeng: good catch! https://github.com/open-rmf/rmf-web/pull/986/commits/363bf503a65793ca0c184aa914738b495a7ba0b8

koonpeng:
New changes put all the build time config in `buildConfig`, fix some issues when building in humble and add a couple more build options.

aaronchongth (approved):
Thanks! LGTM

PR #987, opened 2024-08-06
Allow pydantic 2.8.2 to be installed after python 3.8 following pypi

aaronchongth:
## What's new

This fixes the issue of `pipenv-install` setting up `pydantic==1.10.7` when developing locally on Ubuntu Jammy + ROS 2 Humble.

It is a little odd that setting up the web stack in docker doesn't run into this issue.

Updated the markers to follow https://pypi.org/project/pydantic/. The markers are listed based on the environment where `pipenv lock` was run, resulting in `python >= 3.12`

## Self-checks

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -705,7 +705,7 @@
                 "sha256:6f62c13d067b0755ad1c21a34bdd06c0c12625a22b0fc09c6b149816604f7c2a",
                 "sha256:73ee9fddd406dc318b885c7a2eab8a6472b68b8fb5ba8150949fc3db939f23c8"
             ],
-            "markers": "python_version >= '3.12' and python_version < '4.0'",
+            "markers": "python_version >= '3.8' and python_version < '4.0'",
             "version": "==2.8.2"
         },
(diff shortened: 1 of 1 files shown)


koonpeng (commented):
Did you run `pipenv lock` (on python3.12) and see if it overrides the marker?

koonpeng (changes requested):
There seems to be some bug in pydantic as it should error out if it cannot find a compatible package, but instead it silently installs pydantic 1, ignoring the requirements that we need pydantic 2.

What I found was that it works when you run `pipenv update -d`, but the problem is that it overwrites all the markers to target python3.10 so we cannot commit that. There doesn't seem to be a way to support multiple python versions so I guess using `humble` on main would be a case of "if you hack it enough it will work but it's not officially supported".

aaronchongth:
yeah sorry forgot to post my reply earlier, doing a `pipenv lock` on noble with this commit will revert it back to `python >= 3.12` so I agree this is a little weird, we'll keep a mental note on this for deployments. But this PR is not valid, thanks for checking through

PR #989, opened 2024-08-06
Minor dashboard fixes

aaronchongth:
## What's new

* use relative path for icon
* make trajectory server connection optional as core library nodes are sometimes shut down

## Self-checks

### DIFF
--- a/packages/dashboard/index.html
+++ b/packages/dashboard/index.html
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <link rel="icon" href="/favicon.ico" />
+    <link rel="icon" href="favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <meta
       name="description"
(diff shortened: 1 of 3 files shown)


packages/dashboard/index.html
```diff
@@ -2,7 +2,7 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <link rel="icon" href="/favicon.ico" />
+    <link rel="icon" href="favicon.ico" />
```
koonpeng, line 5:
```suggestion
    <link rel="icon" href="%BASE_URL%favicon.ico" />
```

As shown in vite docs, seems that it expects base url to always end with a trailing slash https://vitejs.dev/config/shared-options#base. So we may need to update the app-config, we can also probably simplify `url.ts` with this assumption.

koonpeng (changes requested):

packages/dashboard/index.html
aaronchongth: thanks for the reference, changed in https://github.com/open-rmf/rmf-web/pull/989/commits/8e6d672b0ea95b93849d97a85723d0653f5505e0

koonpeng (approved):

PR #993, opened 2024-08-08
Setting camera target to most negative number to handle arbitrary large maps

aaronchongth:
## What's new

* Tweak camera target to most negative number to better accommodate almost-orthogonal view for extremely large maps
* refactored camera control to prevent creation of camera every time zoom changes

## Self-checks

### DIFF
--- a/packages/dashboard/src/components/map-app.tsx
+++ b/packages/dashboard/src/components/map-app.tsx
@@ -547,7 +547,7 @@ export const MapApp: StyledComponent<MicroAppProps & MUIStyledCommonProps<Theme>
           }}
           orthographic={true}
         >
-          <CameraControl zoom={zoom} sceneBoundingBox={sceneBoundingBox} />
+          <CameraControl zoom={zoom} />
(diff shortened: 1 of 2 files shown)


koonpeng (commented):
Not sure what this affects, but if we know the size of the map, then could we calculate this dynamically? If not then is there value to make this an app-config?

aaronchongth:
This changes the 3D position that the camera points towards, raising it to a large negative number implies that the camera will almost always be pointing down vertically, no matter where the camera is moved to in 3D space. This helps mimic an orthogonal view.

Although this value falls under the category of bigger-is-better, there's no harm in using the scene size to calculate it dynamically I guess.

I've also refactored the camera controls a little to reduce repeated construction of cameras and its controller

koonpeng (changes requested):
Sounds like there is no harm in having a smaller value? If so then let's revert the dynamic calculation change.

aaronchongth:
As explained the purpose of these changes in DM

koonpeng (approved):

PR #1002, opened 2024-08-21
dashboard: fix alert audio

koonpeng:
1. fix react hooks eslint config
2. remove docs saying that it must be a .wav file, it should support any format supported by the browser, .wav is highly discouraged because of the filesize.
3. fix new alert element being created with every render.
4. fix other hooks violations


aaronchongth (approved):
Thanks for polishing it up! LGTM!

PR #899, opened 2024-02-13
Shuts down server if internal websocket connections exceed max allowed defined in config

aaronchongth:
## What's new

Implements an optional static limit to the number of internal websocket connections allowed. Supercedes https://github.com/open-rmf/rmf-web/pull/895

Temporary fix for https://github.com/open-rmf/rmf-web/issues/897

* Removes `WebsocketHealthManager`, just uses `ConnectionManager`
* New configurable parameter `max_internal_websocket_connections`. This value should be number of fleet adapters using `BroadcastClient` + number of task dispatchers

To test
* add `"max_internal_websocket_connections": 1,` to `packages/api-server/sqlite_local_config.py`, we only allow one fleet adapter connection
* launch office demo, `ros2 launch rmf_demos_gz office.launch.xml server_uri:="ws://localhost:8000/_internal" headless:=1`, this launches `rmf_task_dispatcher` without server connection at the moment
* in a separate terminal start a new task dispatcher and provide it with the server URI, `ros2 run rmf_task_ros2 rmf_task_dispatcher --ros-args -p use_sim_time:=true -p server_uri:="ws://localhost:8000/_internal"`
* The server will announce that the number of internal websocket connections was exceeded and shut down

## Self-checks

- [ ] I have prototyped this new feature (if necessary) on Figma 
- [ ] I'm familiar with and follow this [Typescript guideline](https://basarat.gitbook.io/typescript/styleguide)
- [ ] I added unit-tests for new components
- [ ] I tried testing edge cases
- [ ] I tested the behavior of the components that interact with the backend, with an e2e test

PR #1003, opened 2024-08-22
Allow task and fleet updates over ROS 2

aaronchongth:
## What's new

Sets up ROS 2 subscriptions for fleet and task related updates, that were solely on websockets `_internal` route in the past.

The motivation is related to https://github.com/open-rmf/rmf-web/pull/899, where non-reproducible issues with `websockets` occur unpredictably in production environments, causing various systems to fail.

This allows the option to mitigate issues with the upstream websocketpp library, and just rely on ROS 2 to publish updates instead. 

Works with https://github.com/open-rmf/rmf_ros2/pull/383

## Self-checks

### DIFF
--- a/packages/api-server/api_server/gateway.py
+++ b/packages/api-server/api_server/gateway.py
@@ -3,9 +3,11 @@
 import asyncio
 import base64
 import hashlib
+import json
 import logging
 from datetime import datetime
 from typing import Any, cast
+from uuid import uuid4
 
 import rclpy
 import rclpy.client
@@ -35,19 +37,27 @@ from rmf_task_msgs.srv import CancelTask as RmfCancelTask
 from rmf_task_msgs.srv import SubmitTask as RmfSubmitTask
 from rosidl_runtime_py.convert import message_to_ordereddict
 from std_msgs.msg import Bool as BoolMsg
+from std_msgs.msg import String as StringMsg
 from tortoise.exceptions import IntegrityError
 
 from api_server.exceptions import AlreadyExistsError, InvalidInputError, NotFoundError
 from api_server.fast_io.singleton_dep import singleton_dep
+from api_server.logging import default_logger
 from api_server.models.user import User
 from api_server.repositories.alerts import AlertRepository
 from api_server.repositories.cached_files import get_cached_file_repo
+from api_server.repositories.fleets import FleetRepository
 from api_server.repositories.rmf import RmfRepository
+from api_server.repositories.tasks import TaskRepository
 from api_server.rmf_io.events import (
     AlertEvents,
+    FleetEvents,
     RmfEvents,
+    TaskEvents,
     get_alert_events,
+    get_fleet_events,
     get_rmf_events,
+    get_task_events,
 )
 from api_server.ros import get_ros_node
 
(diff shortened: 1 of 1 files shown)


koonpeng (changes requested):
It's still not clear where the connection issues come from, there is a chance that upgrading dependencies would have fixed it. There is also a chance that this doesn't fix it as we have seen ros2 messages being significantly delayed before as well.

That said, I'm not opposed to using ros2, but I think we should only receive updates from a single source, there should be an option to choose which source to use.

aaronchongth:
I fully agree that focusing on a single transport mode would be the cleanest way to handle it.

For updating dependencies, we could be limited to half of the equation only, we could update rmf-web as we go, but as the cpp stack uses https://github.com/zaphoyd/websocketpp, which is generally stable but won't be having any updates any time soon.

> There is also a chance that this doesn't fix it as we have seen ros2 messages being significantly delayed before as well.

It does not resolve the root cause of the issue, but that could actually be infrastructure for all we know and therefore there is nothing we can do to properly diagnose and provide a solution for. However allowing the option to use ROS 2 will at least circumvent the symptoms of stalling all the components that use websockets (API server, fleet adapter, task dispatcher), and require manual restarting of these components in a production environment. 

> That said, I'm not opposed to using ros2, but I think we should only receive updates from a single source, there should be an option to choose which source to use.

I agree with this too, however we might have situations where there are fleet adapters running on the same system as the API server (also encountering websocket connection issues) and could benefit from rmw shared memory transport by using ROS 2, whereas there are fleet adapters running in the cloud somewhere that would like to use websockets instead.

The fleet adapters will only have 1 mode of transport to choose from either websockets or ROS 2, but not both, and we can plaster as many warnings or deprecation notices as we want on the ROS 2 endpoints

koonpeng:
> The fleet adapters will only have 1 mode of transport to choose from either websockets or ROS 2, but not both, and we can plaster as many warnings or deprecation notices as we want on the ROS 2 endpoints

I see that there is multiple way this can play out.

1. If the fleet adapters offer both transport simultaneously, it will cause issues if we also listen on both transport. But if we only listen on 1 transport, then there would be no issue.
2. If the fleet adapters only offer one transport at once, then listening on both ws and ros2 gives the advantage of supporting a mix of ws and ros2 adapters.

I think there is value in supporting both transport simultaneously so I prefer going with 1.

packages/api-server/api_server/gateway.py
```diff
@@ -474,6 +498,116 @@ def handle_fire_alarm_trigger(msg):
...
+                await self._fleet_repo.save_fleet_log(fleet_log)
+                self._fleet_events.fleet_logs.on_next(fleet_log)
+
+            self._loop.create_task(save(FleetLog.model_validate(json_msg["data"])))
+
+        fleet_log_update_sub = self._ros_node.create_subscription(
+            StringMsg, "fleet_log_update", handle_fleet_log_update, 10
+        )
+        self._subscriptions.append(fleet_log_update_sub)
+
```
koonpeng, line 610: Please refactor so both the internal route and this share the same code.

koonpeng (changes requested):

aaronchongth:
With the discussion to allow publishing over ROS 2 regardless of the availability of websockets, https://github.com/open-rmf/rmf_ros2/pull/383, we will inevitably run into issues where the API server will be receiving duplicated information if a locally running fleet adapter is provided websockets access. 

Apologies for the back-and-forth, we can retain the current behavior of using websockets only on rmf-web for now, until there is a more tangible requirement for supporting both websockets and ROS 2.

For now, with the fleet adapters and task dispatcher publishing over ROS 2 by default, since the original motivation is to provide a workaround for some obscure issue, users/collaborators can reference this PR to implement the changes they require for their deployments.

Closing this PR

PR #1008, opened 2024-09-06
api-server: remove stub authenticator

koonpeng:
Remove the stub authenticator in api server. Now instead of a stub, we support using a secret to verify the jwt, for testing, the frontend hardcodes a token signed with the secret. This brings the test environment closer to prod and allows us to test with other claims in the future (e.g. address, emails, custom claims).

### DIFF
--- a/packages/dashboard/src/services/stub-authenticator.ts
+++ b/packages/dashboard/src/services/stub-authenticator.ts
@@ -13,7 +13,21 @@ export class StubAuthenticator
   constructor() {
     super();
     this.user = 'stub';
-    // hardcoded token that expires in 2035-01-01
+    // hardcoded token using the secret 'rmfisawesome', expires in 2035-01-01
+    // header:
+    // {
+    //   "alg": "HS256",
+    //   "typ": "JWT"
+    // }
+    // payload:
+    // {
+    //   "sub": "stub",
+    //   "preferred_username": "stub",
+    //   "iat": 1516239022,
+    //   "aud": "rmf_api_server",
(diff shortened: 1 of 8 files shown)


packages/dashboard/src/services/stub-authenticator.ts
```diff
@@ -10,10 +10,26 @@ export class StubAuthenticator
...
+    // payload:
+    // {
+    //   "sub": "stub",
+    //   "preferred_username": "stub",
+    //   "iat": 1516239022,
+    //   "aud": "rmf_api_server",
+    //   "iss": "stub",
+    //   "exp": 2051222400
+    // }
+    this.token =
```
aaronchongth, line 31: this is awesome! Can you regenerate another token such that it is an admin?
You can verify it by seeing if Admin Actions are enabled for this user

aaronchongth (changes requested):
Thanks for cleaning up the `superuser` fiasco too. Other than a new token that's required, LGTM!

packages/dashboard/src/services/stub-authenticator.ts
koonpeng: Added another token for user, also need to change the stub username to "admin". Without using other claims to determine if an user is admin, the only admin we can have is the default admin named "admin". We actually can use jwt claims to determine admin, but just not the `realm_access` claim as that is exclusive to keycloak, we need to define custom roles and document, then configure keycloak to sign those claims instead.

aaronchongth (approved):

PR #1019, opened 2024-10-10
Fix ghpages workflow

aaronchongth:


packages/api-server/scripts/extract_docs.py
```diff
@@ -34,7 +34,7 @@ def cleanup():
 outdir = f"{args.output}"
 os.makedirs(outdir, exist_ok=True)
 
-base_url = "http://localhost:8000/rmf-web"
+base_url = "http://localhost:8000"
```
koonpeng, line 37: The docs is hosted at https://open-rmf.github.io/rmf-web/ and the config public url is `/rmf-web`, in theory this should point to `rmf-web` for it to work.

koonpeng (changes requested):

packages/api-server/scripts/extract_docs.py
aaronchongth: Since we are not running the API server behind a proxy while collecting the docs, the `/rmf-web` public URL will not be applied, and everything needs to be scraped from `http://localhost:8000/docs` for example. However we will still need `/rmf-web` listed in the eventual html to work on ghpages

aaronchongth: For hosted environment, do you mean the docs hosting environment, https://open-rmf.github.io/rmf-web/?
We are populating the `/rmf-web` path using the docs retrieval API server config that still maintains the public URL `/rmf-web`

It is just the downloading of source files while it is hosted on localhost that needs to have no `/rmf-web` path

koonpeng (approved):
If it works it works

aaronchongth:
For future reference and more context, the server's public URL had to be behind a proxy to take effect was already documented 3 years ago, https://github.com/open-rmf/rmf-web/blob/main/packages/api-server/api_server/default_config.py#L9-L12, with some more docs from fastapi, https://fastapi.tiangolo.com/advanced/behind-a-proxy/#proxy-with-a-stripped-path-prefix

I'm just more surprised it worked in the past, perhaps it was the major version updates that broke the behavior.

PR #1036, opened 2024-11-25
Increase testing

aaronchongth:
## What's new

* Correct `codecov.yml`
* Increase testing

## Self-checks


aaronchongth:
To keep the commits small, I'll continue with more testing on a subsequent PR

codecov.yml
```diff
@@ -2,6 +2,11 @@ coverage:
   status:
     project: false
     patch: false
+ignore:
+  - packages/rmf-dashboard-framework/examples
+  - "**/*.js"
+  - "**/*.stories.tsx"
+  - "**/index.ts"
```
koonpeng, line 9: `index.ts` shouldn't be ignored.

packages/rmf-dashboard-framework/src/components/map/map.tsx
```diff
@@ -532,6 +533,7 @@ export const Map = styled((props: MapProps) => {
           camera.updateProjectionMatrix();
         }}
         orthographic={true}
+        resize={{ polyfill: ResizeObserver }}
```
koonpeng, line 536: Looking at the browser support, I don't think there is a need for us to polyfill.

koonpeng (changes requested):

packages/rmf-dashboard-framework/src/components/map/map.tsx
aaronchongth: Yup, that was added only for testing, I've found a solution for that on https://github.com/ZeeCoder/use-resize-observer/issues/40#issuecomment-644536259 to use a mock resize observer

https://github.com/open-rmf/rmf-web/pull/1036/commits/3affe31c348d916d8f6858051c45479455a9caf6

codecov.yml
aaronchongth: Thanks for catching that, I accidentally left that in after trying some stuff

https://github.com/open-rmf/rmf-web/pull/1036/commits/3affe31c348d916d8f6858051c45479455a9caf6

koonpeng (approved):

Issue #1034, opened 2024-11-25
[Other issue]: Waypoints labels are not properly align with waypoint.


harshp1509:
### Before proceeding, is there an existing issue or discussion for this?

- [X] I have done a search for similar [issues](.?q=type%3Aissue) and [discussions](https://github.com/open-rmf/rmf/discussions).

### Description

I am trying to visualize the world environment than waypoint label are not align with the waypoints. so what parameters i have to change to align it?
![Screenshot 2024-11-25 101601](https://github.com/user-attachments/assets/4ad3e374-402d-474b-a8e2-1b1c9d68c429)

harshp1509:
I am using humble branch of rmf_web.
I have tried with two maps one is gazebo_world and another is my own custom created map. for both map, i am facing same issue.

This both map's building.yaml and image :-

gazebo world map:-

```
coordinate_system: reference_image
crowd_sim:
  agent_groups:
    - {agents_name: [], agents_number: 0, group_id: 0, profile_selector: external_agent, state_selector: external_static, x: 0, y: 0}
  agent_profiles:
    - {ORCA_tau: 1, ORCA_tauObst: 0.40000000000000002, class: 1, max_accel: 0, max_angle_vel: 0, max_neighbors: 10, max_speed: 0, name: external_agent, neighbor_dist: 5, obstacle_set: 1, pref_speed: 0, r: 0.25}
  enable: 0
  goal_sets: []
  model_types: []
  obstacle_set: {class: 1, file_name: turtlebot_world_navmesh.nav, type: nav_mesh}
  states:
    - {final: 1, goal_set: -1, name: external_static, navmesh_file_name: ""}
  transitions: []
  update_time_step: 0.10000000000000001
graphs:
  {}
levels:
  turtlebot_world:
    drawing:
      filename: world.png
    elevation: 0
    lanes:
      - [0, 1, {bidirectional: [4, true], demo_mock_floor_name: [1, ""], demo_mock_lift_name: [1, ""], graph_idx: [2, 0], orientation: [1, ""], speed_limit: [3, 0]}]
      - [1, 2, {bidirectional: [4, true], demo_mock_floor_name: [1, ""], demo_mock_lift_name: [1, ""], graph_idx: [2, 0], orientation: [1, ""], speed_limit: [3, 0]}]
[...]

PR #1037, opened 2024-11-25
Remove offset based on text length

aaronchongth:
## What's new

Resolves https://github.com/open-rmf/rmf-web/issues/1034
* removes an offset for labels based on text length, it will be more intuitive that the text is left-justified at the center of the rendering

## Before
Screenshot_20241125_171228
Screenshot_20241125_171250

## After
Screenshot_20241125_171149
Screenshot_20241125_171320

## Self-checks


koonpeng (approved):
I recall there used to be an arrow pointing to the target, which is why there is an X offset. It appears there is a regression which removes the arrow? Or was it never ported over to the threejs map?

aaronchongth:
It was never ported to threejs, and we added the layer filters and hover feature to enlarge the labels too, so the arrow was kind of redundant

PR #1047, opened 2025-02-07
Deprecate delivery alert model fields and route

aaronchongth:
## What's new

Now that we have the generic alerts system, `delivery_alerts` should be deprecated before removed, since those were application specific.

## Self-checks


koonpeng (approved):

packages/api-server/api_server/routes/delivery_alerts.py
```diff
@@ -17,7 +17,9 @@ async def sub_delivery_alerts(_req: SubscriptionRequest):
     return get_rmf_events().delivery_alerts.pipe(rxops.filter(lambda x: x is not None))
 
 
-@router.post("/{delivery_alert_id}/response", response_model=DeliveryAlert)
+@router.post(
+    "/{delivery_alert_id}/response", response_model=DeliveryAlert, deprecated=True
```
koonpeng, line 21: Add docs for alternatives for those who are using this api.

aaronchongth: Good point, added that and also deprecated the router,
image

PR #1049, opened 2025-02-12
Revert to pnpm official mirror

aaronchongth:
## What's new

* Revert to using official `pnpm` mirror
* Target nightly minimal rmf docker image
* Fix CI

TODO: fix api-server rolling CI, will open another ticket for that

## Self-checks


packages/rmf-dashboard-framework/vitest.setup.ts
```diff
@@ -0,0 +1,9 @@
+import { vi } from 'vitest';
+
+/* @ts-ignore */
```
koonpeng, line 3: We should use `ts-expect-error` instead, with a comment explaining why that is the case.

koonpeng (approved):

PR #1060, opened 2025-06-24
Prep for kilted

aaronchongth:
## What's new

* image building for jazzy has additional step to build and install orphaned alert messages
* api server conditionally imports the alert related messages from the orphaned `rmf_alert_msgs` if using jazzy, otherwise the usual `rmf_task_msgs`
* more reusable image build for kilted and rolling
* retarget latest as rolling
* updated docs

## Self-checks

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,9 +1,8 @@
-name: Tmp push
+name: Nightly
 on:
-  pull_request:
-  # schedule:
-  #   # 2am SGT
-  #   - cron: '0 18 * * *'
+  schedule:
+    # 2am SGT
+    - cron: '0 18 * * *'
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true


.github/minimal-rmf/Dockerfile
```diff
@@ -31,3 +16,8 @@ ENV PATH "$PNPM_HOME:$PATH"
 RUN pnpm env use --global lts
 
 RUN apt update && apt install -y python3-venv
+
+# cleanup
+RUN rm -rf \
+    /var/lib/apt/lists \
+    /dist
```
koonpeng, line 23: Is `/dist` used by any commands?

.github/minimal-rmf-jazzy/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+ENV PATH "$PNPM_HOME:$PATH"
+
+RUN pnpm env use --global lts
+
+RUN apt update && apt install -y python3-venv
+
+# cleanup
+RUN rm -rf \
+    /var/lib/apt/lists \
+    /dist
```
koonpeng, line 40: Is `/dist` used by any commands?

.github/minimal-rmf-jazzy/Dockerfile
```diff
@@ -0,0 +1,40 @@
+ARG ROS_DISTRO=jazzy
+ARG BASE_IMAGE=docker.io/ros:$ROS_DISTRO-ros-core
+FROM $BASE_IMAGE
+
+# These branches are only required for Jazzy, as some required messages are not
+# backported to be released.
+ARG INTERNAL_MSGS_BRANCH=dd2951debfe53d0391896ecb9ddf07f027af05e9
+ARG BUILDING_MAP_MSGS_BRANCH=e26cf73ec7b1f61bbd450a4f85450e0db20d6c72
```
koonpeng, line 8: I'm not sure if I understand this, `minimal-rmf/Dockerfile` is changed to install rmf from binary, it should work for jazzy as well. The comments suggest that we are using messages incompatible with jazzy, even if there is no practical differences, it still invalidates this as a jazzy build.

koonpeng (changes requested):

.github/minimal-rmf-jazzy/Dockerfile
aaronchongth: https://github.com/open-rmf/rmf-web/pull/1060/commits/e794621373f4e907a9c2ecaf37eacd87da329b0f

.github/minimal-rmf/Dockerfile
aaronchongth: https://github.com/open-rmf/rmf-web/pull/1060/commits/e794621373f4e907a9c2ecaf37eacd87da329b0f

.github/api-server/Dockerfile
```diff
@@ -29,9 +29,9 @@ RUN apt update && apt install -y ros-$ROS_DISTRO-rmw-cyclonedds-cpp
 
 # according to rosdep, the messages have no deps so we can just copy the files directly in!
 # this reduces the image size by ~50%.
-COPY --from=base /opt/rmf/include /opt/ros/$ROS_DISTRO/include
-COPY --from=base /opt/rmf/lib /opt/ros/$ROS_DISTRO/lib
-COPY --from=base /opt/rmf/share /opt/ros/$ROS_DISTRO/share
+COPY --from=base /opt/ros/$ROS_DISTRO/include /opt/ros/$ROS_DISTRO/include
+COPY --from=base /opt/ros/$ROS_DISTRO/lib /opt/ros/$ROS_DISTRO/lib
+COPY --from=base /opt/ros/$ROS_DISTRO/share /opt/ros/$ROS_DISTRO/share
```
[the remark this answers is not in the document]
aaronchongth: > reducing the image size

image

I've switched to using ros-core images and only installing the bare minimum `ros-dev-tools` required to build the messages. Along the way it seems like there is still a lot of caching or cruft being installed. Performing this copy onto a fresh ros-core image still gets us down on the image size significantly

> This whole 2nd step seems to be redundant, instead of copying the whole workspace, the previous step should build the api-server package, then the 2nd step will install it. This is to strip away rmf-web build dependencies, if we are copying the entire workspace in, that defeats the purpose of a 2-step build.

Note that this is a quick start docker image for users, and not a deployment. There will not be a DB available, so we are relying on the default launch settings of creating a local DB for each run within the container for quick testing. This also helps with debugging as the source code is readily available in the image.

.github/minimal-rmf-jazzy/Dockerfile
aaronchongth: as mentioned in the conversation below

README.md
```diff
@@ -143,6 +142,8 @@ ros2 launch rmf_demos_gz office.launch.xml server_uri:="http://localhost:8000/_i
 
 # Troubleshooting
 
+- Alert related messages were introduced after Open-RMF Jazzy releases, hence it is not available with Open-RMF Jazzy binaries, and have to be built from source to be used by the API server. See https://github.com/open-rmf/rmf-web/issues/1041 for more information, as well as how the [Jazzy API server base image is built](.github/minimal-rmf-jazzy/Dockerfile). ROS 2 distributions after Jazzy can just use the released binaries.
```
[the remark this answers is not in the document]
aaronchongth: Since the migration to Jazzy it has already not been compatible in the sense of debian packages, but the API server has never been compatible, we are already using pip to install dependencies on a virtual environment. Since our past decision about releasing the API server (or the lack of ability) onto the buildfarm or pypi, we will be relying on docker images instead, hence `rmf_deployment_template`

The alerts are used for implementing crucial features of the API server and dashboard, I'd rather keep the feature alive with this caveat instead of rolling a feature back. If a user intends to deploy the API server, only an underlay image is required. At the end of the day, the requirement of using released binaries on any given ROS 2 distro, is to enable release on the buildfarm, which we will not be achieving any time soon without a total re-write.

packages/api-server/api_server/gateway.py
```diff
@@ -63,6 +62,20 @@
...
 from .repositories import CachedFilesRepository
 
+# This conditional import is necessary for Alert related messages as they were
+# not released in to ROS 2 Jazzy. Once we stop supporting Jazzy, we can remove
+# this conditional import and use the Alert messages from rmf_task_msgs directly.
+# fmt: off
+# isort: off
+# pylint: disable=import-error,no-name-in-module
+if os.getenv("ROS_DISTRO") == "jazzy":
+    from rmf_alert_msgs.msg import Alert as RmfAlert, AlertResponse as RmfAlertResponse  # pyright: ignore[reportMissingImports,reportAttributeAccessIssue]
```
koonpeng, line 72: For the jazzy branch, we cannot import messages not part of jazzy release. What I would recommend is to refactor out all the alert subscriptions to another file and create a jazzy version of that file that does not do anything. So we would have

`alert_gateway.py`
`alert_gateway_stub.py`

`alert_gateway_stub.py` would just be a stub that does not do anything. Then on the main gateway, do a conditional import for either one of them depending on the env. For pyright to work, we may also need to add a `if TYPE_CHECKING:` conditional import to tell it to use the stub. pyright may still complain errors in the jazzy CI so we may need to exclude the file as well (only in jazzy CI).

aaronchongth:
Closing with https://github.com/open-rmf/rmf-web/pull/1066

PR #1084, opened 2026-05-11
feat(api-server): accept namespaced preferred_username claim fallback

ren-diao-quikbot:
## Problem

The current authenticator requires a non-namespaced `preferred_username` claim in access tokens:

```python
# packages/api-server/api_server/authenticator.py
if not "preferred_username" in claims:
    raise AuthenticationError(
        "expected 'preferred_username' username claim to be present"
    )
```

For access tokens issued via the OAuth 2.0 `client_credentials` (M2M) flow, **Auth0 cannot put `preferred_username` (or any other OIDC-standard claim name) on access tokens with a custom API audience**. Two of the general restrictions Auth0 documents apply together:

> "OPENID standard claims and claims used internally by Auth0 cannot be customized or modified"
>
> "Access tokens with an Auth0 API audience, excluding the `/userinfo` endpoint, cannot have private, non-namespaced custom claims"
>
> — Auth0 docs: [Create Namespaced Custom Claims](https://auth0.com/docs/secure/tokens/json-web-tokens/create-namespaced-custom-claims) (under "General restrictions")

The same page closes the loop on what happens when a non-namespaced custom claim is set anyway:

> "In case of collisions, the transaction won't fail, but your custom claim won't be added to your tokens."

So an Auth0 Action that calls `api.accessToken.setCustomClaim("preferred_username", ...)` during a `credentials-exchange` trigger silently drops the claim — the Action completes with no error, but the bare claim never lands on the token. The same call with a namespaced name (`api.accessToken.setCustomClaim("https://example.com/preferred_username", ...)`) does land.

This two-rule policy is one provider's defensible implementation of the standards' guidance on attribute names in JWT access tokens:

- **RFC 9068 §2.2.2 (Identity Claims)** — "Authorization servers MAY return arbitrary attributes not defined in any existing specification, as long as the corresponding claim names are **collision resistant** or the access tokens are meant to be used only within a private subsystem. Please refer to Sections 4.2 and 4.3 of [RFC 7519] for details." [link](https://www.rfc-editor.org/rfc/rfc9068.html#section-2.2.2)
- **RFC 7519 §4.2 (Public Claim Names)** — "Claim Names can be defined at will by those using JWTs. However, in order to prevent collisions, any new Claim Name should either be registered in the IANA 'JSON Web Token Claims' registry … or be a Public Name: a value that contains a Collision-Resistant Name." [link](https://www.rfc-editor.org/rfc/rfc7519.html#section-4.2)

Together: the standards require collision-resistant names for attributes that aren't IANA-registered; Auth0 enforces that requirement strictly for access tokens with a custom API audience, with the consequence that the M2M flow cannot satisfy the bare `preferred_username` requirement on an Auth0-issued access token without either (a) using an ID token instead of the access token (not what the spec or the API server expects), or (b) patching `authenticator.py`.

I'm not asserting that Okta and AWS Cognito apply the same filtering policy — I do not have first-party documentation to back that claim, so I have withdrawn it from the patch and the docstrings. Operators on other providers with comparable collision-avoidance policies may benefit; the value-add of this PR is specifically the Auth0 path, with the optional fallback shape leaving room for similar cases without naming them.

## Reproduce on Auth0 (no code required)

In any Auth0 tenant with a custom API audience:

1. Create an Auth0 Action on the `credentials-exchange` trigger:

   ```js
   exports.onExecuteCredentialsExchange = async (event, api) => {
     api.accessToken.setCustomClaim("https://your-namespace.example/preferred_username", "test-literal");
     api.accessToken.setCustomClaim("preferred_username", "test-literal"); // non-namespaced
   };
   ```

2. Mint an access token via `client_credentials` against the custom API.
3. Decode the JWT payload (e.g. `jwt.io`). You will see:
   - `"https://your-namespace.example/preferred_username": "test-literal"` ✓
   - **No** bare `"preferred_username"` claim — silently dropped by Auth0 with no error to the caller.

## Fix

Add an optional `preferred_username_claim_namespace` config field. When set, the authenticator falls back to looking up `f"{namespace}preferred_username"` if the bare `preferred_username` claim is absent.

- The bare `preferred_username` claim remains the **first** preference — existing deployments are byte-identical in behavior.
- The namespaced lookup runs **only** when the bare claim is missing **and** the namespace config is set.
- Default value is `None` (no fallback) — so existing deployments need no configuration changes.

Example config snippet for an Auth0 deployment:

```python
config = {
    ...
    "preferred_username_claim_namespace": "https://your-namespace.example/",
    ...
}
```

The auth flow then resolves `preferred_username` from `f"https://your-namespace.example/preferred_username"` in the access-token claims.

## Tests

`packages/api-server/api_server/test_authenticator.py` covers five scenarios through the public `verify_token(token)` surface (HS256 token built with PyJWT, verified end-to-end through `jwt.decode` + claim lookup):

1. Bare `preferred_username` claim is accepted (no namespace configured).
2. Missing claim raises `AuthenticationError`.
3. Namespaced fallback is used when `preferred_username_claim_namespace` is configured and the bare claim is absent.
4. Namespaced claim is ignored when no namespace is configured (current behavior preserved).
5. Bare claim takes precedence over the namespaced variant when both are present.

## Notes / open questions for the reviewer

1. **Config-field naming.** `preferred_username_claim_namespace` is verbose but precise. If a more generic name like `jwt_claim_namespace` (paired with the existing `jwt_public_key` / `jwt_secret` family) reads better and leaves room for future namespaced-claim lookups, happy to rename — just say the word.
2. **Scope.** This only changes the `preferred_username` lookup. If `authenticator.py` starts consuming other OIDC-reserved claims later, the same fallback can be applied trivially.

## Changes since first push

- (2026-05-12) Tightened RFC citations to §2.2.2 of RFC 9068 + §4.2 of RFC 7519 after reviewer feedback. Quoted text verified character-by-character against the canonical RFC editor plain-text source.
- (2026-05-12) Reframed the Auth0 case using the two general-restrictions Auth0 documents — earlier drafts conflated Auth0's *restricted-claims* list (which contains `sub`/`aud`/`iss`/etc.) with the *general restrictions* on access tokens (which is what actually filters `preferred_username`). `preferred_username` is in Auth0's "non-restricted claims" group but is still affected by the non-namespaced-custom-claim general restriction.
- (2026-05-12) Withdrew the Okta + AWS Cognito assertions from the README and code docstrings; the directly-verified case is Auth0 only.
- (2026-05-12) Rewrote `test_authenticator.py` to exercise the public `verify_token(token)` API instead of the protected `_get_user(claims)` helper, per reviewer feedback. All five scenarios remain covered.
- (2026-05-12) Added `Generated-by: Anthropic Claude Code (Claude Opus 4.7)` trailer to all commits in compliance with [OSRF's policy on the use of generative tools in contributions](https://github.com/openrobotics/osrf-policies-and-procedures/blob/main/OSRF%20Policy%20on%20the%20Use%20of%20Generative%20Tools%20(%E2%80%9CGenerative%20AI%E2%80%9D)%20in%20Contributions.md).

---

**AI-tool disclosure (per OSRF policy):** This contribution was drafted with assistance from Anthropic Claude Code (Claude Opus 4.7). The contributor (Ren Diao) reviewed and is responsible for the final content of every commit.

### DIFF
--- a/packages/api-server/api_server/test_authenticator.py
+++ b/packages/api-server/api_server/test_authenticator.py
@@ -1,72 +1,89 @@
 import unittest
 from unittest.mock import AsyncMock, patch
 
+import jwt
+
 from api_server.authenticator import AuthenticationError, JwtAuthenticator
 from api_server.models import User
+
+
+SECRET = "test-secret"
+AUD = "rmf_api_server"
+ISS = "test"
 
 
 def _make_authenticator(
     *, preferred_username_claim_namespace: str | None = None
 ) -> JwtAuthenticator:
     return JwtAuthenticator(
-        "test-secret",
-        aud="rmf_api_server",
-        iss="test",
+        SECRET,
+        aud=AUD,
+        iss=ISS,
         preferred_username_claim_namespace=preferred_username_claim_namespace,
     )
 
 
-class TestGetUser(unittest.IsolatedAsyncioTestCase):
+def _make_token(extra_claims: dict) -> str:
+    """Encode a HS256 JWT with the test secret + standard aud/iss + extra claims."""
+    claims = {"aud": AUD, "iss": ISS, **extra_claims}
+    return jwt.encode(claims, SECRET, algorithm="HS256")
+
+
+class TestVerifyToken(unittest.IsolatedAsyncioTestCase):
     async def test_bare_preferred_username_claim_is_accepted(self):
         authenticator = _make_authenticator()
+        token = _make_token({"preferred_username": "alice"})
         with patch.object(
             User,
             "load_or_create_from_db",
             new=AsyncMock(return_value=User(username="alice")),
         ) as mock_load:
-            user = await authenticator._get_user({"preferred_username": "alice"})
+            user = await authenticator.verify_token(token)
         self.assertEqual(user.username, "alice")
         mock_load.assert_awaited_once_with("alice")
 
     async def test_missing_claim_raises_authentication_error(self):
         authenticator = _make_authenticator()
+        token = _make_token({})
         with self.assertRaises(AuthenticationError):
-            await authenticator._get_user({})
+            await authenticator.verify_token(token)
 
     async def test_namespaced_claim_used_as_fallback_when_namespace_configured(self):
         authenticator = _make_authenticator(
             preferred_username_claim_namespace="https://example.com/"
         )
-        claims = {"https://example.com/preferred_username": "bob"}
+        token = _make_token({"https://example.com/preferred_username": "bob"})
         with patch.object(
             User,
             "load_or_create_from_db",
             new=AsyncMock(return_value=User(username="bob")),
         ) as mock_load:
-            user = await authenticator._get_user(claims)
+            user = await authenticator.verify_token(token)
         self.assertEqual(user.username, "bob")
         mock_load.assert_awaited_once_with("bob")
 
     async def test_namespaced_claim_ignored_when_namespace_not_configured(self):
         authenticator = _make_authenticator()
-        claims = {"https://example.com/preferred_username": "bob"}
+        token = _make_token({"https://example.com/preferred_username": "bob"})
         with self.assertRaises(AuthenticationError):
-            await authenticator._get_user(claims)
+            await authenticator.verify_token(token)
 
     async def test_bare_claim_preferred_over_namespaced_claim(self):
         authenticator = _make_authenticator(
             preferred_username_claim_namespace="https://example.com/"
         )
-        claims = {
-            "preferred_username": "alice",
-            "https://example.com/preferred_username": "bob",
-        }
+        token = _make_token(
+            {
+                "preferred_username": "alice",
+                "https://example.com/preferred_username": "bob",
+            }
+        )
         with patch.object(
             User,
             "load_or_create_from_db",
             new=AsyncMock(return_value=User(username="alice")),
         ) as mock_load:
-            user = await authenticator._get_user(claims)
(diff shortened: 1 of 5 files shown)


koonpeng:
Thanks for the PR. I am trying to understand why a namespace is needed. The link to the RFC you provided does not contain the quote you give. Can you also help point out where in the auth0 and okta docs does it mention that it will filter out non namespaced claims?

packages/api-server/api_server/test_authenticator.py
```diff
@@ -0,0 +1,75 @@
...
+
+class TestGetUser(unittest.IsolatedAsyncioTestCase):
+    async def test_bare_preferred_username_claim_is_accepted(self):
+        authenticator = _make_authenticator()
+        with patch.object(
+            User,
+            "load_or_create_from_db",
+            new=AsyncMock(return_value=User(username="alice")),
+        ) as mock_load:
+            user = await authenticator._get_user({"preferred_username": "alice"})
```
koonpeng: `_get_user` is a protected method, use `verify_token` instead.

koonpeng (changes requested):
See https://github.com/openrobotics/osrf-policies-and-procedures/blob/main/OSRF%20Policy%20on%20the%20Use%20of%20Generative%20Tools%20(%E2%80%9CGenerative%20AI%E2%80%9D)%20in%20Contributions.md on the OSRF policy on ai generated code.

tldr: If AI is used, please add a `Generated-by` or `Co-authored-by` footer to the commit message.

koonpeng (approved):
Thanks for the PR.

PR #1092, opened 2026-06-15
move pnpm overrides

aaronchongth:
### Fixed bug

Move pnpm overrides (added when fixing security issues) to `pnpm-workspace.yaml`, due to deprecation of the override parameters in package.json

### GenAI Use
We follow [OSRA's policy on GenAI tools](https://github.com/openrobotics/osrf-policies-and-procedures/blob/main/OSRF%20Policy%20on%20the%20Use%20of%20Generative%20Tools%20(%E2%80%9CGenerative%20AI%E2%80%9D)%20in%20Contributions.md)

- [ ] I used a GenAI tool in this PR.
- [x] I did not use GenAI

Generated-by:

### DIFF
--- a/.github/actions/bootstrap/action.yml
+++ b/.github/actions/bootstrap/action.yml
@@ -12,7 +12,7 @@ runs:
   using: composite
   steps:
     - name: Install dependencies
-      run: (pnpm install --filter ${{ inputs.package }}... --no-frozen-lockfile || (pnpm approve-builds --all && pnpm install --filter ${{ inputs.package }}... --no-frozen-lockfile))
+      run: (pnpm install --filter ${{ inputs.package }}... || (pnpm approve-builds --all && pnpm install --filter ${{ inputs.package }}...))
       shell: bash
     - name: Build
       if: inputs.skip-build == 'false'
--- a/.github/actions/bootstrap/action.yml
+++ b/.github/actions/bootstrap/action.yml
@@ -12,7 +12,7 @@ runs:
   using: composite
   steps:
     - name: Install dependencies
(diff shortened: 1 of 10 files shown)


.github/actions/bootstrap/action.yml
```diff
@@ -12,7 +12,7 @@ runs:
   using: composite
   steps:
     - name: Install dependencies
-      run: (pnpm install --filter ${{ inputs.package }}... || (pnpm approve-builds --all && pnpm install --filter ${{ inputs.package }}...))
+      run: (pnpm install --filter ${{ inputs.package }}... --no-frozen-lockfile || (pnpm approve-builds --all && pnpm install --filter ${{ inputs.package }}... --no-frozen-lockfile))
```
koonpeng: Don't think we should use `--no-frozen-lockfile`. If it fails with `--frozen-lockfile`, it usually means the lockfile is out of sync and we should commit an updated lockfile instead.

pnpm-workspace.yaml
```diff
@@ -1,3 +1,31 @@
...
   - 'packages/*'
   - 'pipenv-install'
+
+allowBuilds:
+  '@nestjs/core': true
+  '@openapitools/openapi-generator-cli': true
+  '@swc/core': true
+  protobufjs: true
+
+overrides:
```
koonpeng, line 18: Good to document the use of these overrides. Usually they indicate some form of tech debt and will bite us back in the future.

koonpeng (changes requested):

koonpeng (approved):

